PR #20, opened 2016-05-19
Make failing tests pass

### LINKED ISSUE #15: Tests are failing on fresh install

sebbacon:
**Make failing tests pass**

Fixes #15


annapowellsmith:
Ah, so for me `open` in Python is indifferent to the case of the filename, I guess that's why it was passing for me and not you. Still don't understand why the first test was failing, but it's a mystery I'm content to leave unsolved :)

PR #21, opened 2016-05-19
Tweak GA tracking event sent on chart load

### LINKED ISSUE #19: Add analytics event for bookmarked URLs containing search queries
On analyse page, send URL back as an event in analytics on load (so search URLs are logged on page load, not just on search, so we can track bookmarked URLs)

sebbacon:
**Tweak GA tracking event sent on chart load**
* Always send it, even for searches that return no data;
* Add a parameter indicating if the source was a search form click, or a
  page load

Closes #19


openprescribing/media/js/src/chart.js
```diff
@@ -101,21 +101,29 @@ var analyseChart = {
...
+        if ($(this.el.submitButton).data('clicked'))
+            button_clicked = true;
         this.el.submitButton.button('reset');
         this.globalOptions.activeOption = 'items';
         this.setUpData();
         this.globalOptions.allMonths = utils.getAllMonthsInData(this.globalOptions.data.combinedData);
         this.globalOptions.activeMonth = this.globalOptions.allMonths[this.globalOptions.allMonths.length-1];
         this.globalOptions.friendly = formatters.getFriendlyNamesForChart(this.globalOptions);
+        this.hash = hashHelper.setHashParams(this.globalOptions);
+        if (button_clicked)
```
annapowellsmith: I think adding the source to the hash is a good move. Maybe we could just have `this.hash += ($(this.el.submitButton).data('clicked')) ? '&source=button' : 'source=pageload'` and get rid of the lines above and below?

annapowellsmith: ooh also just spotted while writing that that we need an ampersand in the second option too.

annapowellsmith:
As discussed we'll have to wait for deploy to know if this is actually working, but looks good!

Issue #10, opened 2016-05-16
Update URL for barchart / timegraph / map


sebbacon:
When you run a search, the URL updates so you can go back / bookmark
It would be nice to be able to link directly to the map (for example)

PR #25, opened 2016-05-20
Support selecting analysis panel via URL

sebbacon:
**Support selecting analysis panel via URL**
This allows people to bookmark maps and time series.


sebbacon:
Closes #10

annapowellsmith:
Needs rebase? Also needs the minified JS file I'm afraid.

annapowellsmith:
Looks good! Tested cross-browser. The only nitpicky change I might make is instead of `data-target` we could just use the `href` attribute, since we have to have the `href` for Bootstrap tabs anyway. (For some reason GitHub won't let me leave a line note.)

sebbacon:
See 6126f28b0e5f5086dfc8d6f40c6e731541fda2bf for a (slightly unsatisfactory) explanation of why we are using data-target

annapowellsmith:
Ah OK! Merging 😸

PR #32, opened 2016-05-25
Add a "share" button to the active tab

sebbacon:
It should be more obvious to end users that it is possible to pass
around URLs for particular analyses.

This takes the form of a small button with a modal dropdown that shows
the URL that can be copied (plus the ability to copy by clicking a
button).

Another option would have been to use "social sharing" buttons, but it's
not clear yet if this would be the main form of sharing, as opposed to
sending things around in emails.

The button is positioned in the active tab because it was difficult to
find a place we could consistently place it within the visualisation,
because the visualisation layouts vary quite a lot (e.g. between a chart
and a map).


sebbacon:
Note that this branch is based off `sticky-search-view-urls-seb`, so should be merged to master from the command line (as apparently it's not possible to change the base for an existing PR).

As of the time of this comment, it should merge cleanly with master.

annapowellsmith:
Er, slightly confused, sorry... do you want me to merge to master, and then push, before reviewing?

sebbacon:
No, it is reviewable as it is, but when you're happy, merge to master from the command line, rather than via github (which will merge it to a now-closed branch).

annapowellsmith:
Think we are missing JS build artefacts. NB I would include these in the same commit as the base file changes, rather than making separate commits for them.

annapowellsmith:
Re merge - ok, so when happy, I should merge to master at command line, run tests, check everything looks OK in browser, force-push to master, and then close PR without merging? that the plan?

sebbacon:
Re artefacts, I think they're in ef94d2141d9854e3fda377fd791f02839605f4fc ?

Re merge - yes, that's right, except you won't need to force push, just a normal push, and assuming the merge is clean, testing is a bit redundant (we don't test after every merge to master normally do we?)

annapowellsmith:
Is there not some compiled JS too?

Re merge - ah ok, I was thinking rebase so the commits from this branch would be below other commits from master (hence need to test etc) - but I guess if I merge rather than rebase that won't be the case.

sebbacon:
Ah, sorry, I've now squashed the missing JS into that commit.

annapowellsmith:
Don't forget unminified files too...

Looks gorgeous in Chrome, but not in Safari / IE11 I'm afraid:

screenshot 2016-05-25 15 51 55

Also copy button seems not to work in Safari. 

Degrades below 600px but I'm not sure if that's a problem (Analytics might tell us this).

annapowellsmith:
Browserstack is [free for open source](https://www.browserstack.com/pricing) ❤️ so we can do that if necessary

sebbacon:
Getting a running ie10 on my computer is the best next step. Progressing slowly due to enormously slow downloads.

sebbacon:
WFM in IE11 and Safari 8 (using BrowserStack -- painful experience but managed in the end! See screenshot)
bs_macyos_safari_8 0

annapowellsmith:
Can you pop it on staging and I'll try it out on Safari/iOS?

On 5 July 2016 at 08:36, Seb Bacon <notifications@github.com> wrote:

> WFM in IE11 and Safari 8 (using BrowserStack -- painful experience but
> managed in the end! See screenshot)
> [image: bs_macyos_safari_8 0]
> <https://cloud.githubusercontent.com/assets/211271/16577183/87c6240a-428b-11e6-9c5f-a85f0b3fe352.jpg>
>
> —
> You are receiving this because you were assigned.
> Reply to this email directly, view it on GitHub
> <https://github.com/ebmdatalab/openprescribing/pull/32#issuecomment-230407202>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAExTImWQtrGPfpXRAkFxIHvDDeBxbHtks5qSglwgaJpZM4Imda_>
> .
>



-- 
Anna Powell-Smith MA MPhil
07812 357037
anna.powell-smith@phc.ox.ac.uk
Honorary Research Fellow
Centre for Evidence-Based Medicine
Department of Primary Care Health Sciences
University of Oxford
Woodstock Road
Oxford OX2 6GG
www.ebmDataLab.net <http://www.ebmdatalab.net/>

sebbacon:
Deployed to staging

annapowellsmith:
Nice one getting Browserstack working! 

The good news: this now looks perfect on Safari / OSX. 

The bad news is that I now see a JS error on staging on IE8 (production is ok):

screen shot 2016-07-11 at 11 50 37

These are typically just syntax and easy to fix, once you've found the offending line, which seems to be somewhere around here (bit difficult to tell with minified JS and IE doesn't do sourcemaps - would it be easy to set a flag on staging not to use minified JS?):

screen shot 2016-07-11 at 11 53 55

If you have an IE10 VM set up then you could try setting compatibility mode and debugging that way. 

Finally, sorry to be a pain, but could we get rid of the superfluous commits on this PR? The easiest solution may be to create a new branch and cherry-pick them in.

sebbacon:
Compatibility mode doesn't reproduce the error. Downloading new virtualbox now.

PR #29, opened 2016-05-24
Automated deployments

sebbacon:
Support automated deploys from the command line.

This PR includes resetting our migrations so we can confidently rely on migrations that have been checked into the source code from now on (and can confidently automate running migrations as part of a deploy).

Therefore, before merging this branch, it should be deployed to staging with:

    fab deploy:staging,branch=deployment

And then we should run the following to ensure the production database thinks its migrations are fully up to date:
  
    cd /webapps/openprescribing_staging
    echo "DELETE FROM django_migrations" | python manage.py dbshell --settings=openprescribing.settings.production
    python manage.py migrate --fake

Finally, it should be safe to merge this PR, then deploy with 

    fab deploy:production

PR #76, opened 2016-06-06
Reduce number of 

sebbacon:
**Remove redundant files**
This files are now generated by the deploy process (since #29 was
completed)

**Support selecting keys to return for CCGs from API**
Also provide relevant feedback to users if the requested keys are not
supported.

**Use new key-selection API when rendering charts**

**Update docs to match new key-selection API**


openprescribing/api/views_org_details.py
```diff
@@ -31,7 +39,19 @@ def org_details(request, format=None):
...
     elif org_type == 'ccg':
-        query = 'SELECT pct_id AS row_id, name as row_name, *'
+        if keys:
+            keys = keys.split(",")
+            cols = []
+            aliases = {'row_id': 'pct_id',
+                       'row_name': 'name'}
+            for k in keys:
+                if k in aliases:
+                    cols.append("%s AS %s" % (aliases[k], k))
```
annapowellsmith: We need to pass `keys` as a parameter to `execute_query` rather than inserting directly, to prevent SQL injection attacks.

sebbacon: Oops! (Hooray for code reviews...)

annapowellsmith:
A great improvement. 

I started fiddling around to see if we could reduce the size of the whitelist, and I believe we can. I've also updated the key code to work for both practices and CCGs:

http://pastebin.com/raw/e1Mj5J52
http://pastebin.com/raw/tesu9F2C

I also made `row_id`, `row_name` and `date` return in every request, on the basis that the data isn't much use without them. 

Review appreciated, and if you approve then feel free to update the commits. 

This needs a minor change to the JS to stop asking for `row_id` etc.

sebbacon:
Looks great, thanks!

annapowellsmith:
Just realised that this will probably need the URLs to be updated in the dashboards JS file too.

PR #77, opened 2016-06-09
Add a dummy search pageview for search Analytics

### LINKED ISSUE #38: Implement analyse queries as Site Search queries in analytics
This will give us more powerful query facets in GA. 

We should continue to send them as events too for back-compatibility.

sebbacon:
Closes #38


sebbacon:
I've also set the parameters up as "search" parameters in Analytics

annapowellsmith:
Excellent! 

Do the search parameters need to be set up manually in GA, i.e. if we add new parameters in the hash, do we need to remember to configure them in GA too? Or will it pick them up automatically?

sebbacon:
It's manual, I'm afraid.

PR #78, opened 2016-06-09
Use minified js in production

sebbacon:
In production, use minified files; when in debug mode, use non-minified.

Note this PR is based off remove-redundant-files, so should be closed and merged before that one (and then we can let the tests pass there before merging to master)


annapowellsmith:
Looks great

PR #86, opened 2016-06-21
Add switch to compute measures based on file path

sebbacon:
This is required to support automated monthly updates. It allows us to
pass the path of the last-imported prescribing data as use that as the
key for knowing which date measures need updating for

### DIFF
--- a/openprescribing/frontend/management/commands/import_measures.py
+++ b/openprescribing/frontend/management/commands/import_measures.py
@@ -107,8 +107,8 @@ class Command(BaseCommand):
         if not options['month'] and not options['end_date'] \
            and not options['month_from_prescribing_filename']:
             err = 'You must supply either --month or --end_date '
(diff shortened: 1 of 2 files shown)


openprescribing/frontend/management/commands/import_measures.py
```diff
@@ -99,15 +104,23 @@ def parse_options(self, options):
             options['measure_ids'] = [k for k in options['measures']]
 
         # Get months to cover from options.
-        if not options['month'] and not options['end_date']:
+        if not options['month'] and not options['end_date'] \
+           and not options['month_from_prescribing_filename']:
             err = 'You must supply either --month or --end_date '
-            err += 'in the format YYYY-MM-DD. You can also '
+            err += 'in the format YYYY-MM-DD, supply a path to a file which '
```
annapowellsmith: Sorry to be niggly but is this comment missing an "or"? I fear that if we don't fix it it will cause confusion in the future.

PR #88, opened 2016-06-22
Import shorter BNF codes

### LINKED ISSUE #81: Don't reject presentations with BNF codes shorter than 15 chars
In our `import_bnf_codes` script we reject presentations with BNF codes that are shorter than 15 chars. 

However, some devices seem to have presentation codes of less than 15 chars. For example:

21010220113    Flutter Oscillating Positive Expiratory Pressure Dev   
21010220114   Acapella Oscillating Positive Expiratory Pressure Dev   

These are legitimate so we should not exclude them.

sebbacon:
Closes #81 
Examples include dressings, ileostomy bags, etc

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -8,8 +8,9 @@ engines:
     enabled: true
     config:
       languages:
-      - python
-      - javascript
+        python:
+          mass_threshold: 70
+        javascript:
   eslint:
     enabled: true
   fixme:
--- a/openprescribing/frontend/management/commands/import_bnf_codes.py
+++ b/openprescribing/frontend/management/commands/import_bnf_codes.py
@@ -57,7 +57,7 @@ class Command(BaseCommand):
 
             pres_name = row['BNF Presentation'].strip()
             pres_code = row['BNF Presentation Code'].strip()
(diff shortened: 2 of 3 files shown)


annapowellsmith:
Have you tried running this with the BNF dataset? We may make assumptions elsewhere in the codebase about the length of BNF codes.

sebbacon:
Yes, as in, a full dataset imports directly, but that only exercises the script and the models; is there anything else I should be running?

annapowellsmith:
It's just odd that I've explicitly written a test to ensure that product codes are always 15 characters. I can't see any assumptions about it elsewhere in the codebase, but it would be good just to run a few searches, check out the dashboards etc, and be sure it doesn't break anything.

sebbacon:
Yes, it is odd isn't it - that's why I've described  the shape of the data in #81 in case it gives any insights. Which it didn't ;)

It's hard for me to do anything that feels like an exhaustive test on my development box as I don't have so much data ingested.

Now I understand the data better, perhaps I should invest some time in generating a representative dataset for development...

annapowellsmith:
Can you run searches for these products on the analyse page on your dev box, and see the correct data in the front-end? If so, I'll merge.

sebbacon:
Tested and happy with it.

PR #89, opened 2016-06-22
Return 404 for unknown BNF codes

### LINKED ISSUE #69: Invalid BNF codes in API spending calls should not 500
This should not 500 https://openprescribing.net/api/1.0/spending/?code=0

It should either 404, or return 200 with empty response.

sebbacon:
Closes #69


annapowellsmith:
:thumbsup:

PR #95, opened 2016-06-30
Make importing list sizes idempotent

### LINKED ISSUE #93: Correctly integrate patient list size importing to workflow
See #83 for background

sebbacon:
Running the `import_list_sizes` management command with a path to a
filename will cause the data in that file to be applied to all practices
up to the date of the latest prescribing data.

There is also a check to ensure that the latest prescribing data is not
newer than the list sizes data. Note that this continues to rely on the
user correctly naming a file path: rather than the filename mattering,
it's now the correct use of the `%Y_%m/`-containing-folder-convention in
the source data that is used to determine the last date to which the
list size data is valid.

Closes #93


annapowellsmith:
We seems to have lost the ability to import all files at once (by not specifying a filename). Not necessarily a problem, but does the new setup process have another way to do that?

annapowellsmith:
Just read through this but I think we'll need to have a chat to make sure I understand it properly... Sorry 🐱

sebbacon:
Need to add comments to tests.

annapowellsmith:
As discussed, I just need a few more comments on the tests to understand what's happening, and a bit of documentation in the openprescribing-data repo on how to import all list sizes when you're setting up. Hope that's OK.

sebbacon:
For the record, I did a pretty major refactoring of the single commit in this PR, so decided to amend the commit and force-push it (so the comments above relate to a commit that no longer exists)

annapowellsmith:
I'll review this when the tests pass.

sebbacon:
*sigh*, the tests were failing due to conflicting migrations (which didn't conflict in terms of code).

So I fixed that with a merge, and then noticed that this PR is not based off master.

I think I'll have to unpick this on Monday as I have to go now.

annapowellsmith:
ok no worries. I've unassigned myself, please just reassign me when this is ready to review.

sebbacon:
Superceded by #119

PR #98, opened 2016-07-06
Sort measures with worse performers at the top

### LINKED ISSUE #97: Sort measures correctly
On the new CCG and practice dashboards, we sort the measures by "outsiderishness", highest first. However, it turns out that higher is not always worse, when it comes to the KTTs.

Establish for which of the KTTs lower is worse - Richard has sent us a partial list, we need to chase him to complete this. 

Then work out a way in the JS to rank correctly. This will probably involve adding a new field to the Measure table, returning it in the API, and using this in the JS to order the ranking. 

And write comprehensible copy explaining how the ranking works, not too long or confusing.

sebbacon:
* Add a `low_is_good` flag to measures
* Use this to decide sort order on relevant pages
* Refer to the sort order in the measure's descriptive text

Fixes #97


annapowellsmith:
I'm not sure this is working correctly on per-measure pages - if I go to e.g. http://localhost:8000/ccg/02Q/statins/ locally (a measure for which high=bad) it now seems to be sorting the practices with the lowest centiles at the top. We want the worst practices to be at the top. 

We also need to amend the sorting of measures on per-organisation pages, i.e. on a dashboard for a CCG, KTT12 (for which low=bad) should be near the top if it's low, near the bottom if it's high. That doesn't seem to be the case for me locally.

openprescribing/media/js/src/measure_utils.js
```diff
@@ -67,11 +67,20 @@ var utils = {
         panelData = _this._rollUpByOrg(panelData[0], options.orgType);
       }
       panelData = _this._getSavingAndPercentilePerItem(panelData,
-        numMonths);
-      data = _.sortBy(panelData, function(d) {
-        if (d.meanPercentile === null) return -1;
-        return d.meanPercentile;
-      }).reverse();
+                                                       numMonths);
+      if (options.rollUpBy == 'measure_id' || options.lowIsGood) {
```
annapowellsmith: I think this is the issue. `panelData` can be structured such that each chart panel is a measure (on pages like `/ccg/XXX/measures`) or such that each chart panel is an organisation (on pages like `/ccg/XXX/statins`). We'll need to amend the sorting code for each of these scenarios.

sebbacon:
OK, sorry it's taken me so long, but I've got to the bottom of it now.  I had only dealt with sorting orgs on a per-measure page. Due to a misreading of the original issue #97 I thought that we didn't need to change anything on the pages where there are many measures for one organisation. I now realise that was an obvious misreading, sorry.

> I'm not sure this is working correctly on per-measure pages - if I go to e.g. 
> http://localhost:8000/ccg/02Q/statins/ locally (a measure for which high=bad) it now seems to be sorting 
> the practices with the lowest centiles at the top. We want the worst practices to be at the top.

Did you run `import_measures` first when testing? This case definitely works for me:

A measure where `low_is_good == true`:
low_is_good

A measure where `low_is_good == false`:
low_is_not_good

sebbacon:
Deployed to staging.

Some example URLs:

* A [measure where low is good](http://staging.openprescribing.net/measure/arb/)
* A [measure where high is good](http://staging.openprescribing.net/measure/ktt13_nsaids_ibuprofen/)
* A [CCG whose worst-performing measure is one where high is good](http://staging.openprescribing.net/ccg/09W/measures/)
* A [CCG whose worst-performing measure is one where low is good](http://staging.openprescribing.net/ccg/01G/measures/)

annapowellsmith:
Those links are really helpful, thanks!

All looks great.

PR #106, opened 2016-07-14
Use an environment variable to select API host

### LINKED ISSUE #105: Parameterize API Host
It is useful in development to be able to test the charts against the production API. It should be possible to set a runtime parameter to control this.

sebbacon:
When set, client-side code will prepend the host to API calls. When
unset, code will issue requests without a host (i.e. using whatever
domain the client is using).

Closes #105


annapowellsmith:
Looks good except I think you need to update BNF and chemical pages too - that will probably need a template change as the URL is set there.

annapowellsmith:
👍

PR #107, opened 2016-07-14
Improvements to measure definitions

annapowellsmith:
Make ARB, ACE, statin and PPI descriptions consistent.


annapowellsmith:
Spotted this while reviewing

sebbacon:
Looks good.

Question: what are the long versions for? The short versions are barely shorter and I don't see the long versions in the graphs.

annapowellsmith:
Good spot, I never used these for anything in the end. Have removed them and updated models.

PR #111, opened 2016-07-15
Fix some ie8 issues

sebbacon:
The homepage buttons have lost their protuding-button styling that they inherited from the browser (see below). I thought that probably didn't matter enough to reintroduce?

home


annapowellsmith:
Thanks - no the styling is not a problem.

annapowellsmith:
Thanks for spotting this!

PR #114, opened 2016-07-15
Add levetiracetam to the list of custom measures

### LINKED ISSUE #113: Add levetiracetam as a cost saving measure
My analysis identified Keppra as the single drug for which the cost saving from switching to generic (levetiracetam from branded (Keppra) would be largest, around £12 million per year. 

The MHRA says that [generic levetiracetam is generally fine](https://www.epilepsy.org.uk/news/news/mhra-new-guidance-switching-meds-63630).

So we should add it as a cost saving measure (numerator branded Keppra, denominator all levetiracetam).

annapowellsmith:
Fixes #113.


sebbacon:
:+1:

PR #118, opened 2016-07-18
In spine chart, always show ticks for active orgs

### LINKED ISSUE #117: It should always be possible to see selected practice(s) on spine chart
We've had a report that when using the Highcharts "print" function, the histogram labelling does not always show the currently selected practice(s).

sebbacon:
This PR adds ticks on the x-Axis for any selected orgs.

This ensures that a practice someone is interested in always has a label, so can be identified on the graph.

However, it does make the layout look a bit jarring sometimes, because the default tick algorithm ensures ticks are evenly-spaced. Adding extra ticks between evenly-spaced ticks therefore crowds the labels sometimes.

It may be worth investing a bit more time into refining the tick-insertion algorithm to remove non-active orgs which are adjacent to active ones; on the other hand, if someone has selected two adjacent orgs, then the crowding will be unavoidable, so the current solution may be acceptable. See screen grabs for illustration.

Closes #117

### DIFF
--- a/openprescribing/media/js/src/bar-chart.js
+++ b/openprescribing/media/js/src/bar-chart.js
@@ -34,11 +34,27 @@ var barChart = {
     };
     var activeOrgs = _.pluck(globalOptions.orgIds, 'id');
     this.barData = this._indexDataByMonthAndRatio(globalOptions.data.combinedData,
-                                                      activeOrgs);
+                                                  activeOrgs);
+    // Ensure we always show ticks for any active (selected) orgs:
+    barOptions.xAxis.tickPositioner = function() {
+      var calculated = this.tickPositions;
+      var activeOrgsIndex = [];
+      _.each(this.series[0].options.data, function(d, i) {
+        if (d.active) {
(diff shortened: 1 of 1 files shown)


annapowellsmith:
What's a good test case to have a look at?

openprescribing/media/js/src/bar-chart.js
```diff
@@ -34,11 +34,27 @@ var barChart = {
...
+    // Ensure we always show ticks for any active (selected) orgs:
+    barOptions.xAxis.tickPositioner = function() {
+      var calculated = this.tickPositions;
+      var activeOrgsIndex = [];
+      _.each(this.series[0].options.data, function(d, i) {
+        if (d.active) {
+          activeOrgsIndex.push(i);
+        }
+      });
+      console.log(this.chart.series[0].data)
```
annapowellsmith: rogue log statement?

annapowellsmith:
Looks nice - I don't think the overlapping is a problem.

Issue #124, opened 2016-07-19
Cost-saving descriptions are missing from charts


annapowellsmith:
On the measures pages, we calculate the potential cost savings for each measure/organisation, but these descriptions are not included in the template. Fix this.

annapowellsmith:
Fixed in https://github.com/ebmdatalab/openprescribing/pull/125

Issue #103, opened 2016-07-11
Change sort options text on new measures pages


annapowellsmith:
On the new [organisation measures pages](http://staging.openprescribing.net/practice/A82077/measures/), I think the first button should be "Potential for improvement" rather than "Percentile".

This is less jargony and explains more clearly why the order matters. 
 
It'll also work better when we have more measures for which low=bad, in which case the sort order will be less obvious visually.

annapowellsmith:
Fixed in https://github.com/ebmdatalab/openprescribing/pull/125

PR #125, opened 2016-07-19
Change wording of sort options 

annapowellsmith:
Minor fix for #103. We will test this wording but I think it's clearer.

Also includes a fix for #124.


sebbacon:
Looks good. Was this something I had let fall by the wayside?

annapowellsmith:
Don't think so! Think they vanished in my refactor.

PR #127, opened 2016-07-19
Fix import measures argparse #122

sebbacon:
This PR is a new PR on the branch covered by now-closed PR #123.

That PR got a commit it shouldn't have, which was nothing to do with the issue.

Now it's been merged with master, I think it's better to have the original branch addressing the original issue, albeit with commits that don't really belong there. The alternative is to rewrite history, which is arguably a better option. Comments please!

See the individual commit messages for more info.

### DIFF
--- a/openprescribing/frontend/management/commands/import_measures.py
+++ b/openprescribing/frontend/management/commands/import_measures.py
@@ -120,7 +120,7 @@ class Command(BaseCommand):
             filename = options['month_from_prescribing_filename']
             date_part = re.findall(r'/(\d{4}_\d{2})/', filename)[0]
             month = datetime.strptime(date_part + "_01", "%Y_%m_%d")
-            options['month'] = month.strftime('%Y-%m-01')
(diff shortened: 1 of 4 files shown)


annapowellsmith:
Ah OK. I must admit I was baffled by #123 - that will teach me not to ask questions!

Will take a look now.

annapowellsmith:
Let's not rewrite history 😄 

So, there are two commits here:

1. I can see what d76dbf7 is doing 
2. 0ef6c95 creates an initial value for `current_at` in `ImportLog`, except under test conditions - is that correct? And if so why?

sebbacon:
Because migrations are run before each test, and if you add seed data before tests, you'll end up with unexpected data in your database (i.e. that didn't come from the fixtures).

The alternative is not to use a migration but some kind of one-off command. I chose migrations because it was guaranteed to be run on all deployments -- the alternative is some kind of manual step and/or a different system for hooking post-deploy steps.

PR #131, opened 2016-07-27
Fix sorting measures by potential saving 

### LINKED ISSUE #128: 'Sorting by cost saving' isn't sorting by monetary savings available
https://openprescribing.net/ccg/12F/measures/

Click on the 'Potential savings' button - this should sort by the size of the cost saving available, not the mean percentile. 

Probably my bad for not specifying exactly how this should work!

sebbacon:
Fixes #128 

Also introduceds a selenium-and-saucelabs-based functional test suite (including deleting the unused tests that were already there).


annapowellsmith:
Please could fixes be added to the previous commit as an amendment (`git add`, `git commit --amend`, `git push -f`)? That way the commit log is cleaner.

sebbacon:
Sure. Getting it all to play nicely (with saucelabs, travis environment, docker) takes lots of individual pokes. When I have the setup working I'll squash the commits.

annapowellsmith:
cool thanks!

.travis.yml
```diff
@@ -17,7 +23,12 @@ cache:
...
+  - docker-compose run --service-ports test
 notifications:
   email:
-    - tech@ebmdatalab.net
\ No newline at end of file
+    - tech@ebmdatalab.net
+addons:
+  sauce_connect:
+    username: ebmdatalab
+    access_key: 397c2f14-5ab3-43cf-a47c-019807cd3c44
```
annapowellsmith: Are we happy it's ok to have the access key here? (It may be unavoidable, but just wanted to check.)

README.md
```diff
@@ -110,6 +110,46 @@ If required, you can run individual Django tests as follows:
...
+browser being launched.
+
+If you want to run the tests headless (i.e. without launching a
+browser), you have two choices:
+
+### Run the functional tests using Xvbf
+
+To do this, you'll need to install
+[pyvirtualdisplay](http://pyvirtualdisplay.readthedocs.io/en/latest/#installation)
+and Xvbf. This is, apparently, quite hard to do on OS X.
```
annapowellsmith: If I install these libraries, will the tests automatically run headlessly? Or do I need to do something else too?

annapowellsmith:
All works for me, so pace the above comments, looks ready to merge! 🎆 

It would be nice to bring back the old functional tests - I doubt the interface has changed much since I wrote them. But probably better to do this as and when we work on the interface, rather than spending a lot of time on it now.

README.md
sebbacon: Yes.

.travis.yml
sebbacon: Well spotted, my intention was to ask myself that question properly before deploying :S

PR #140, opened 2016-08-08
In measures, flip yAxis when a low value is bad

### LINKED ISSUE #138: Make measures consistently high=bad
Feedback from Henry and Lisa that it's really confusing having a mix of high=good and high=bad, and I agree. 

Henry suggested rewriting the definitions so that high=bad consistently. 

Breaking this out from #133 as a separate issue. 

I think we can leave the `low_is_good` JSON setting implemented previously in place, but use it to change the calculation of the measure, rather than the sort order.

sebbacon:
Fixes #138


annapowellsmith:
👍

Issue #146, opened 2016-08-15
Configure DNS for mailgun


sebbacon:
We need to add the following to our DNS so we can send email via mailgun (connected to #130)

I don't know who our DNS provider is -- can I do this myself or can you do it @annaps?

```
TXT	openprescribing.net	v=spf1 include:mailgun.org ~all
TXT	pic._domainkey.openprescribing.net  k=rsa; p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQClHgXdmjdhuprKRY8sDRgGs8d9D1NAfmNDxq9mB+bqz7VmIMhuGpQf9dkzFERKnV8/byBX849zAMQhHppJCV9UlEEOW2qyRQQbvbuyTPHSah4ZTzKsyUk9JjbbCXoe7d3jWXXihdEY0c880EERHJHhtK6mDkPORvhoTuc2JmcY2wIDAQAB
CNAME	email.openprescribing.net mailgun.org
```

sebbacon:
Done in cloudflare

PR #147, opened 2016-08-15
Add feedback form

### LINKED ISSUE #130: Add JavaScript "send us feedback" overlay thing
One of the generic ones

UPDATE: To clarify the background... We would like somewhere that users can send feedback in an instant, frictionless way. I believe are various JS plugins that make it easy for users to do this, without having to leave their workflow - typically these present as an unobtrusive popup on the page. We'd like one that simply sends us an email, rather than a full CRM system.

sebbacon:
A popup feedback form on every page of the site.

Closes #130 

Note: uses mailgun to send emails, therefore connected to #146

### DIFF
--- a/.env-dev
+++ b/.env-dev
@@ -7,3 +7,4 @@ SECRET_KEY=secret
 POSTGRES_PASSWORD=pass
 POSTGRES_USER=user
 POSTGRES_DB=openprescribing_dev
+MAILGUN_API_KEY=xxxxxx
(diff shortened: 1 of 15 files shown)


annapowellsmith:
As discussed - this needs more UX / browser / validation work, and Seb's time is better spent elsewhere. Closing and replacing with #158, but cherry-picking the selenium change into the new PR.

PR #158, opened 2016-08-19
Add doorbell.io feedback overlay

### LINKED ISSUE #130: Add JavaScript "send us feedback" overlay thing

annapowellsmith:
As discussed - this gives us a generic feedback form for the site. 

Closes #130.

Replaces #147, but cherry-picks the selenium commit from that.

### DIFF
--- a/openprescribing/frontend/tests/custom_runner.py
+++ b/openprescribing/frontend/tests/custom_runner.py
@@ -1,0 +1,21 @@
+import os
+import subprocess
+
+from django.conf import settings
+from django.test.runner import DiscoverRunner
+
+
+class AssetBuildingTestRunner(DiscoverRunner):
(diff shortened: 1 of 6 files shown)


annapowellsmith:
looks like we've got some unhappy selenium tests...

sebbacon:
Must be something missing from that refactor commit. I'll have to take a look Monday, got to go shortly.

sebbacon:
Looks good to me.

PR #166, opened 2016-08-24
Add get in touch banner to selected pages

### LINKED ISSUE #164: Add "get in touch for custom analytics" banner

sebbacon:
Closes #164


sebbacon:
Example:

download 1

annapowellsmith:
Thanks. Can you see that image? For me it is transparent...

sebbacon:
No I can't, but weirdly I definitely could earlier...

annapowellsmith:
👻 spooky

annapowellsmith:
Nice idea to use Doorbell! 

However I'm afraid it doesn't work for me - nothing happens when I click the link. There are no errors in the console. This is the case on all the pages, not just the JS-heavy ones. This is Chrome/OSX. 

Please could we set the alert to `max-width: 800px` like the intro paragraphs, so we don't have orphan words?

openprescribing/media/js/bar-charts.js
```diff
@@ -127,3 +127,7 @@ var barChart = {
 };
 
 barChart.setUp();
+
+$(document).ready(function() {
```
annapowellsmith: Might be worth checking that the global `doorbell` variable exists, because I don't think the DOM being ready will guarantee that.

Also, I've just observed that the rest of the code isn't inside `document.ready`, which it maybe should be...

annapowellsmith:
Oh, I know what's going on - you need an `event.preventDefault()` in there, so `$('.doorbell-show').click(function(e) { e.preventDefault(); doorbell.show(); });`.

annapowellsmith:
And finally, we might want to add a `mailto` link to the `a` tag, for non-JS users.

sebbacon:
You're right,  of course, about `preventDefault()`, but it's odd that it worked in Firefox, Chrome and IE for me... I'll get on it.

Is there any point adding a `mailto` when the site is basically useless without Javascript?

annapowellsmith:
Yeah, interesting!

Re `mailto` - yes I think so! We can't get around the fact that it's a JavaScript application, but we can still offer those users something useful.

openprescribing/media/css/general.less
```diff
@@ -16,7 +16,7 @@ h1, h2, h3, h4, h5, h6, .navbar-brand, #navbar {
 h1, h2, h3, h4, h5, h6 {
   text-shadow: 1px 1px 1px rgba(150, 150, 150, 0.64);
 }
-.container p {
+.container p, .alert {
```
annapowellsmith: this will affect all alerts not just the new one, but I assume that looks sensible in other places too?

sebbacon: Yes it does

sebbacon:
Tests passing now @annapowellsmith

annapowellsmith:
👍

PR #179, opened 2016-09-05
Don't error when measures are missing

### LINKED ISSUE #178: Fix low_is_good javascript issue
Example:

```Unable to get property 'low_is_good' of undefined or null reference on
Line 33, column 9288, in https://openprescribing.net/static/js/measures.min.js?```


When we're updating measures, there are certain edge cases where the list of measures returned by the main API endpoint may refer to measures for which there is not yet data for a given CCG.

Specifically, this has happened where "new" measures have been added but not yet calculated yet (in fact, in this case these were old measures which had not been completely removed yet).

The problem is exacerbated where the data returned by the API has been cached by Cloudflare, so what is often a short race condition becomes a longer-term fixture.

The immediate fix is to catch the error where a measure expected to exist does not yet actually exist. Longer term when we use BigQuery for measures (#165) the race condition should be much shorter lived and therefore less likely to be cached; we should also consider only showing measures for which full data has been generated.

sebbacon:
Fixes #178

### DIFF
--- a/openprescribing/media/js/src/measure_utils.js
+++ b/openprescribing/media/js/src/measure_utils.js
@@ -277,7 +277,9 @@ var utils = {
         // If each chart is a different measure, get the
         // centiles for that measure, and if lowIsGood
         var series = _.findWhere(globalData, {id: d.id});
-        d.lowIsGood = series.low_is_good;
+        if (typeof series !== 'undefined') {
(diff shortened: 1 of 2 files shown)


annapowellsmith:
👍 

(tiny request: in future please could lint fixes be squashed into the previous commit?)

PR #180, opened 2016-09-05
Remove "bucket" measures.

### LINKED ISSUE #176: High Cost Statins measure not working
Something is strange...

https://openprescribing.net/ccg/04L/statins/

annapowellsmith:
We will replace these with simpler definitions. Fixes #176.


sebbacon:
Do we need to ticket replacing these?

PR #185, opened 2016-09-07
Documentation for training courses

sebbacon:
Currently imports Google Docs HTML in case the content gets tweaked -
means we can use Google as a kind of CMS.

Longer term this content should probably be included as part of the
source code.

Connected to https://github.com/ebmdatalab/private/issues/12


annapowellsmith:
👍 nice way of doing it

and looks Good Enough, yes! If we carry on doing it like this long-term then we might want to put borders around the images, but not crucial for now.

annapowellsmith:
Please deploy!

PR #188, opened 2016-09-09
Test the app can be loaded in production pip env

sebbacon:
The intention is catch when you accidentally forget to add stuff to
requirements, because it happens to be in the local test/dev
environments).

I expect this to fail until I add a further commit to include the missing requirement.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ services:
   - docker
 script:
   - export SAUCE_HOST=$(ip -4 addr show scope global dev docker0 | grep inet | awk '{print $2}' | cut -d / -f 1)
-  - docker-compose run --service-ports test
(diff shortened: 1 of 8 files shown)


sebbacon:
When this is merged, a new docker image should be built off master: 

    docker build -t ebmdatalab/openprescribing-base .
    docker build -t ebmdatalab/openprescribing-test -f Dockerfile-test .
    docker push ebmdatalab/openprescribing-base
    docker push ebmdatalab/openprescribing-test

sebbacon:
Note that as a side-effect, tests now output warnings about configuration -- see the end of the output at https://travis-ci.org/ebmdatalab/openprescribing/jobs/163114007

annapowellsmith:
Not exactly reviewing this with an expert eye (and I haven't tried running the Docker image), but LGTM 👍

PR #189, opened 2016-09-09
Only allow query params to select API data format

### LINKED ISSUE #168: Debug strange results from BNF name autocomplete search
Was trying to search for lipid regulating drugs in Analyse.  If I type 'lipi' or 'lipid-' it will load search results for it, but if I type 'lipid' nothing comes up. (Initially I thought it wasn't able to find lipid because of quick typing & then noticed this when I started deleting the word). 

I'm presuming this isn't a unique problem to lipid, but not sure what is causing it. I tried searching for anti-arrythmic drugs by just using 'anti' and the search results found it, so it isn't the '-'

image

image

image

image

sebbacon:
Inconsistent use of `?format=json` as opposed to the `Accepts:
application/json` HTTP header meant that sometimes data expected to be
JSON was being served from the cache as HTML.

Fixes #168

### DIFF
--- a/openprescribing/frontend/negotiation.py
+++ b/openprescribing/frontend/negotiation.py
@@ -1,0 +1,28 @@
+from rest_framework.negotiation import DefaultContentNegotiation
+
+
+class IgnoreAcceptsContentNegotiation(DefaultContentNegotiation):
+    def select_parser(self, request, parsers):
+        """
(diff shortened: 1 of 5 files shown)


annapowellsmith:
Looks good!

Will need full cache flush and JS rebuild when deployed.

Issue #211, opened 2016-10-04
Text changes on home page and analyse page


lmfrench:
Can you add a link to the how to use the analyse search somewhere on the analyse page? (https://openprescribing.net/docs/analyse/)

annapowellsmith:
Hey @lmfrench, please could you make this two separate tickets? Sorry, we are EXTREMELY STRICT about keeping tickets about single issues. It's a programmer thing :)

annapowellsmith:
Some further requests from Ben, so the final list of tasks is this:

1. Add a link to "how to use the analyse search" somewhere on the analyse page: (https://openprescribing.net/docs/analyse/)
2. Add the following to the home page: "Use our data: Got a tricky query for the data? We can provide custom extracts, we know the data well, and we’re keen to collaborate with academics and clinicians. Get in touch to find out more." and "How to cite: If you use our data or graphs, please cite as 'OpenPrescribing.net 2016, Anna Powell-Smith and Ben Goldacre', so that others can find us and use our tools."

PR #213, opened 2016-10-07
Helpful text additions

sebbacon:
search
home

Closes issue #211


annapowellsmith:
Nice! The blue is a great improvement. Let's make the alert box on analyse blue too. 

Please could we tweak the margins around the front page alert box so it is the same top and bottom? (or if you put the branch on staging then I will fiddle with it and get the exact numbers)

sebbacon:
done (force-pushed)

margin

annapowellsmith:
Excellent! I would like to find a way to make this whole page friendlier and less busy, but I think we can punt that till our

annapowellsmith:
Broken Travis, sad times :( https://twitter.com/andyhume/status/784404732483923968

annapowellsmith:
:thumbsup:

PR #220, opened 2016-10-18
Ensure dropdowns work on documentation pages

### LINKED ISSUE #219: Fix drop down for 'Trends' and 'More'  on top bar when on help document page
The drop down menus for 'trends' and 'more' don't work if you are on one of the help document pages - if you click them nothing happens, makes it difficult to navigate the site. The other links on the top bar all work fine.

example:
https://openprescribing.net/docs/analyse/#

sebbacon:
Fixes #219

### DIFF
--- a/openprescribing/frontend/tests/functional/test_general.py
+++ b/openprescribing/frontend/tests/functional/test_general.py
@@ -1,5 +1,8 @@
 # -*- coding: utf-8 -*-
 import unittest
+from mock import patch
+from mock import MagicMock
+from mock import PropertyMock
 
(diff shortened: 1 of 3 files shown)


annapowellsmith:
Looks good! Out of interest though, could we not just use openprescribing.js in the same way as other pages? The user will presumably have it cached, so although it's a bigger, may not be less efficient, and would mean we could get rid of some code.

sebbacon:
Yeah, I guess you're right. I'll change it...

sebbacon:
ping @annapowellsmith

annapowellsmith:
Hooray!

PR #238, opened 2016-11-11
Use crosshair on charts to indicate zoomability

### LINKED ISSUE #227: Make it clearer that users can zoom in on our graphs
Make it clearer that you can zoom in on our graphs - change the cursor to a crosshair

As a prescribing analyst, I should be able to know you can zoom in on the graphs without being shown, so that I can more clearly see the time period I am interested in

Priority: Low
Ease: Probably easy

sebbacon:
Closes #227

### DIFF
--- a/openprescribing/media/js/src/highcharts-options.js
+++ b/openprescribing/media/js/src/highcharts-options.js
@@ -87,7 +87,8 @@ var baseOptions = {
     style: {
       color: '#ccc',
       fontSize: '24px',
-      opacity: 0.2
+      opacity: 0.2,
+      cursor: 'crosshair'
     },
     text: "openprescribing.net"
   },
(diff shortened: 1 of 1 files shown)


sebbacon:
Comments from @lmfrench:

> Tested in Safari, Chrome, Firefox, IE10 & IE11 (IE8 crashes on browserstack)
> 
> When using openprescribing.net/analyse, the crosshairs won't allow you to hover over the bars to see items. Works fine in the 'over time' section of analyse, and also works on the dashboards

Fixed

> On the analyse page if you hover below the graph title it becomes a hand, I think it would be better either as a crosshair or normal pointer

I agree it would better; but I starting fixing this, and it turned out to be a surprising faff, and therefore (I reckon) not worth the bother.

lmfrench:
Agree not worth it! :)

lmfrench:
tested on IE11, Chrome & Firefox - works when hovering over bars. 

New problem though - mouse pointer becomes a hand again when hovering over the 'openprescribing.net' part of the graph background. This doesn't happen on the 'over time' version of the chart. (You can still drag & zoom when it is a hand)

lmfrench:
tested in IE11, Chrome & Firefox and working

PR #249, opened 2016-11-24
Remove local prescribing table

sebbacon:
Since we now use Google Bigquery to generate views on our data (see PR #248) to cache locally, there is no longer a need to maintain a local table of prescribing data.

This gains us considerable disk space and processing time on the monthly updates.


sebbacon:
Realised I'm wrong about being able to drop the table: it's required for looking up spending by individual presentation over the API.

PR #252, opened 2016-11-25
Partition prescribing table

### LINKED ISSUE #246: Monthly import is still painfully slow
Copying one month of prescribing data to postgres takes about 3 hours and 10 minutes.

This makes the debug cycle painfully slow when there is an issue with the import; and also slows the whole website down.

It would probably speed things up considerably if we were to partition the table by month, and drop and create indexes as we do when generating measurevalues.

Think about a cheap way to assess if partitioning by month would hurt performance in any significant way; estimate the amount of work and benefits of doing it to make a decision.

sebbacon:
Fixes #246 

When deployed:

* The one-off migration `migrate_hscic_prescribing` needs to be run. The commit that introduced it could then be reverted.
* MeasureValues before 5 years ago should be deleted manually

### DIFF
--- a/openprescribing/frontend/management/commands/create_views.py
+++ b/openprescribing/frontend/management/commands/create_views.py
@@ -10,13 +10,14 @@ import time
 import traceback
 
 from google.cloud import storage
-from google.cloud.storage import Blob
 
 from django.core.management.base import BaseCommand
 from django.db import connection
 
 from common import utils
 from ebmdatalab import bigquery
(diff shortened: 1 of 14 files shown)


sebbacon:
As I'm also migrating to a server with attached storage for the database, there are some extra migration steps:

* pg_dump and pg_restore of all tables except `frontend_prescription`
* set up a separate database connection for the migration, so we read from the old database and write to the new one

PR #266, opened 2016-12-12
20 new measures

sebbacon:


sebbacon:
**Deployment note**

Some of these measures have been removed / renamed, which means they have to be removed in the database:

```python
from frontend.models import Measure
for measure in ['celecoxib.json', 'ktt11_minocycline.json', 'ktt12_diabetes_blood_glucose.json', 'ktt8_antidepressant_first_choice.json', 'ktt8_dosulepin.json', 'ktt9_uti_antibiotics.json', 'pioglitazone.json']:
   Measure.objects.get(pk=measure).delete()

```

PR #355, opened 2017-03-10
Docker

jamesscottbrown:
This pull request makes three changes related to Docker:

- it causes the dev service  to specify a requirements file by providing the argument ``local`` when calling ``docker_setup.sh``. Currently, this is not done, resulting in the error ``Could not open requirements file: [Errno 2] No such file or directory: 'requirements/.txt'``

- it causes the dev service to also change directory into code/openprescribing, so django can be started directly

- it adds an explanation of how to make the application accessible from outside the Docker container to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,18 @@ travis.
     docker push ebmdatalab/openprescribing-base
     docker push ebmdatalab/openprescribing-test
 
+
+### Running the application from within Docker
+To be able to access the django instance running inside Docker from outside the container, docker must be told to publish the port on which Django will listen:
+
+    docker-compose run -p 8000:8000 dev
+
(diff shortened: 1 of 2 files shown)


sebbacon:
Thanks!

Please note [pull requests from forks don't play nicely with Travis encrypted files](https://docs.travis-ci.com/user/encrypting-files/), so your Travis build fails. I've tested it locally.

We've not had any outside contributors yet, so you may find a number of other hurdles on the way to getting started. In particular, loading initial data is going to be a PITA - adding tests and refactoring the [loader software](https://github.com/ebmdatalab/openprescribing-data) is on the cards from next month, but until then you may find it a challenge. Good luck!

PR #376, opened 2017-03-22
Style simplification

sebbacon:
* Drop background wallpaper


sebbacon:
## Before
before

## After
after

HelenCEBM:
I like the no-background, but I'd say the fonts look a bit too standard and boring....

Bengoldacre:
is this to improve consistency on different browsers (system fonts) or just for the look of it? slight preference for original fonts but no strong feelings.

sebbacon:
It is generally just for the look of it -- there was a discussion at a team meeting recently -- but it does have performance benefits too.

The main change requested was the background.

sebbacon:
Updated just to remove background image for the time being.

PR #388, opened 2017-04-04
Alert when there's an inverted ratio w/ list sizes

### LINKED ISSUE #321: Deal with small list sizes
In some cases, list sizes are fictional. In [this case](https://openprescribing.net/analyse/#org=practice&orgIds=Y02873&numIds=0408010AE&denom=total_list_size&selectedTab=summary
), the prescribing is real (149 items) but the list size is false (16 patients).  Therefore it skews the data in terms of cost per patient enormously.

image

This shouldn't be listed as a "4" (GP practice) in our dataset, but it is.  

[According to NHS BSA](https://contactcentreservices.nhsbsa.nhs.uk/selfnhsuktest/template.do;jsessionid=DF983133253CC51D59397C9A225F8D1A?name=What+is+the+procedure+for+setting+up+a+new+practice+or+cost+centre%3F&id=17237), a new GP Practice of Cost Centre must have a patient list size even if there are actually no patients there (e.g. patients are listed elsewhere); and this patient list size can be anything, though 

> we advise that the patient list size does not exceed 100 patients per quarter

Possible approaches are:

* Drop all practices with a list size of 100 or less
* Drop all practices in the bottom centile of list size (or similar)

sebbacon:
When there are small list sizes (e.g. drug dependency specialist
practices) in an analysis, the denominator is sometimes larger than the
numerator, where it would normally be expected to be smaller.

This can skew visualisations. Let a user know when this is possibly
happening, and give them the option to hide such data points.

The decision to hide (or not) is made sticky with a cookie.

Fixes #321


sebbacon:
[example](http://staging.openprescribing.net/analyse/#org=practice&orgIds=Y02873&numIds=0408010AE&denom=total_list_size&selectedTab=summary):

image

lmfrench:
Tested in Chrome with same query as screenshot:

Maps: clicking remove from chart doesn't remove the practice from the map, it changes the items & list size to 0

Text: A lot of people don't know what numerator/denominator means. Could change to  'This chart may be skewed as this data includes some ratios where the denominator is higher than the numerator' so the important bit is at the start, but we're not losing the information.

lmfrench:
Maybe the text should make it clear we're removing practices -'This chart may be skewed. This data includes some practices with ratios where the denominator is higher than the numerator' or could change 'Remove them from chart' to 'Remove these practices from the chart'

sebbacon:
OK, I've changed the logic so it looks for outliers in general, rather than specifically unusual inverted ratios. That means I've been able to make the text clearer (hopefully!). Take a look (deployed to staging)

lmfrench:
Yep prefer the text!

Tested in Chrome, IE11 & Firefox. Still showing on the map when you remove. Switching between is a bit laggy for me in IE11

richiecroker:
We need to be careful that we are alerting to outliers for the "correct" reason (i.e. a small list size), rather than just because they have unusual prescribing behaviours, which we don't want to suggest hiding.

For example, if you undertake this search for co-proxamol at Bristol CCG, three outliers are shown:
image

Looking at the latest data, it identifies AVONMOUTH MEDICAL CENTRE, HELIOS MEDICAL CENTRE and HORFIELD HC as being outliers, although Horfield appears to be 7th highest prescriber/list size in Bristol, and Avonmouth has been shut since January 2016?

richiecroker:
These low volume items (such as co-proxamol) may well always have outlying behaviour, as they are not commonly prescribed items.

sebbacon:
The problem is that there's no single way to reliably identify possibly-unwanted-noise-outliers. If we just go on list size, we could hard code less than 100 but that wouldn't work in many cases. On the other hand, if we just go on a statistical measure of outliers (as implemented here) we'll always pick up genuinely interesting outliers.

The working assumption in what I've implemented so far is just that by pointing out the outliers it's up to the user to interpret and decide if they want to hide them -- we're not advising hiding them, just giving it as an option. Is that too confusing? Could we fix with better wording?

HelenCEBM:
It seems to work pretty well - perhaps it comes across as a bit odd in the text though, because one would (at least initially) think that the outliers are the most interesting ones.. how about something like:

> There appear to be some outliers in this analysis. These may not be interesting if they are caused by an unusually small list size or a very low prescribing volume. Outliers here are: COMPASS HEALTH.  You can _remove it/them_ if you would like to make the rest of the chart easier to read.

(perhaps a bit long!)

lmfrench:
Agree with Helen that saying outlier could makes it sound as though it is more interesting. We could say something like 'The data for 'X' might be making it difficult for you to see the other practices in this CCG. Remove it from the charts' We're then being neutral on the reason while still keeping the text short, and we can link through to a longer explanation in the FAQs

sebbacon:
I'm still not happy with the wording, but I could tweak it for ages, and very few people will ever see it, so I'm deeming it "good enough" for now...

PR #402, opened 2017-04-20
Provide custom error page and text version

### LINKED ISSUE #387: Cloudflare "currently offline" mode is confusing
* User makes a query that takes more than 60 seconds to return
* Our web server times out and returns a 500
* Cloudflare decides the site is down and sets a cookie on the user's browser to indicate this
* The cookie stays on the user's browser for an unpredictable amount of time (I've had it stay for hours)
* As long as the cookie is there, the user can't use the website as they get a Cloudflare "currently offline" message

Ideal solutions might look like one of:

1. Ensure queries never time out:
   1. IIRC more or less impossible as the scan-and-union of many queries is inherently expensive and we can't use more than one CPU for this kind of query (though worth checking)
   2. Use a map-reduce backend instead of postgres, e.g. [Spark](http://spark.apache.org/)
   3. Hybrid: switch to using bigquery for complex queries
2. Gain control over timeout error messages:
   1. Work with Cloudflare support to understand the bug whereby cookies persist for too long
   2. Implement API data caching locally and only use cloudflare for edge caching of assets

sebbacon:
With specific error message for database timeouts

Fixes #387

### DIFF
--- a/openprescribing/frontend/management/commands/generate_presentation_replacements.py
+++ b/openprescribing/frontend/management/commands/generate_presentation_replacements.py
@@ -239,7 +239,7 @@ def write_zero_prescribing_codes_table(level):
 
 
 def get_csv_of_empty_classes_for_level(level):
-    """Using BigQuery, make a CSV of BNG codes at the given level
+    """Using BigQuery, make a CSV of BNF codes at the given level
     (e.g. `section`, `paragraph`) that have never had any prescribing.
 
     Returns a path to the CSV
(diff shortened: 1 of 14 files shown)


sebbacon:
image

lmfrench:
Yes I like it! I'm guessing there isn't a way for me to test it on the staging site?

sebbacon:
Not easily!

openprescribing/frontend/management/commands/generate_presentation_replacements.py
```diff
@@ -218,7 +238,29 @@ def write_temp_code_table(level):
     return table
 
 
+def get_csv_of_empty_classes_for_level(level):
+    """Using BigQuery, make a CSV of BNG codes at the given level
```
inglesp: s/BNG/BNF/ ?

Issue #400, opened 2017-04-19
Remove "worst" reference for neutral measures


sebbacon:
When `low_is_good` is null in a measure, we avoid any value judgement about the sort order.

Currently, we say the following on the page that shows all CCGs for a measure:

> CCGs are ordered by mean percentile over the past six months, with the worst-performing at the top

It should probably just say:

> CCGs are ordered by mean percentile over the past six months

lmfrench:
I think it would be fine to remove 'worst-performing at the top' from all measures rather than doing extra work to make the neutral measures have different text. I can do this if you agree

sebbacon:
Closed by #434

PR #434, opened 2017-05-15
Remove 'prescribing' from title

lmfrench:
Some of the measures already have 'prescribing' in their title, so having text for each measure saying 'prescribing by practices' often makes no sense

e.g. 
Prescribing of pregabalin prescribing by GP practice
Antibiotic stewardship: volume of antibiotic prescribing (KTT9) prescribing by all CCGs

Have removed 'prescribing' from the text - it still makes sense for measures without prescribing in the title e.g. Silver Dressings by all CCGs

### DIFF
--- a/openprescribing/templates/measure_for_practices_in_ccg.html
+++ b/openprescribing/templates/measure_for_practices_in_ccg.html
@@ -2,7 +2,7 @@
 {% load humanize %}
 {% load template_extras %}
 
-{% block title %}{{measure.name}} prescribing by practices in {{ ccg.name }}{% endblock %}
+{% block title %}{{measure.name}} by practices in {{ ccg.name }}{% endblock %}
 {% block active_class %}ccg{% endblock %}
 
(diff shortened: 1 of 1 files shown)


lmfrench:
-Have removed 'prescribing' from the text - it still makes sense for measures without prescribing in the title e.g. Silver Dressings by all CCGs

-Removed first line of text under title, as had same problem & doesn't add anything to page.

-Have also removed 'worst performing practices' as per #400 but @sebbacon might disagree!

richiecroker:
This still seems to be happening?
image

sebbacon:
Would that be because you wrote the words "Prescribing of..." in the measure...?

https://github.com/ebmdatalab/openprescribing/blob/589f2904f54fdc9af4efb7b88665ee3f9a10c856/openprescribing/frontend/management/commands/measure_definitions/libre.json#L7

richiecroker:
No it wouldn’t be

PR #435, opened 2017-05-15
Update measure_for_all_ccgs.html

lmfrench:
Some of the measures already have 'prescribing' in their title, so having text for each measure saying 'prescribing by practices' often makes no sense

e.g.
Prescribing of pregabalin prescribing by GP practice
Antibiotic stewardship: volume of antibiotic prescribing (KTT9) prescribing by all CCG

- Have removed 'prescribing' from the text - it still makes sense for measures without prescribing in the title e.g. Silver Dressings by all CCGs

- Removed first line of text under title, as had same problem & doesn't add anything to page.

- Have also removed 'worst performing practices' as per #400 but @sebbacon might disagree!

### DIFF
--- a/openprescribing/templates/measure_for_all_ccgs.html
+++ b/openprescribing/templates/measure_for_all_ccgs.html
@@ -2,14 +2,12 @@
 {% load humanize %}
 {% load template_extras %}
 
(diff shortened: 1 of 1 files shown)


sebbacon:
Agreed. Closes #400

PR #446, opened 2017-05-27
Add pipeline app for fetching, processing, and importing data

inglesp:
@sebbacon, there's quite a lot here, so you might appreciate a guided tour before reviewing!


sebbacon:
Some little things:
* Just to note, we don't need to keep old data lying around post-run. At a later point we should consider adding a cleaner that gets rid of data older than the month-before-last.  (This is a feature request - not currently handled, so ignore for now!)
* Could rename the pipeline commands to indicate they're fetchers/scrapers? Previously they were in a folder that indicated this 
* Do task objects need documentation field?  I guess the names of the tasks are enough, but it was easier to understand the purpose of various things when they were in one file. Perhaps a bit of tickling the django admin so that you can view a source plus its related tasks on one screen, or similar? Sometimes it is nice to be able to point end users (usually team members) at the manifest to explain where our data comes from.
* We don't need `smoke.sh`, it was superseded by `generate_smoketests`

inglesp:
@sebbacon's approved, so am merging.

Issue #371, opened 2017-03-21
Introduce mobile testing


sebbacon:
We've invested a fair amount of resources in supporting IE8 (for which there are some outstanding issues, see issue #372).

But significantly more people use mobile devices (17% on Safari/iOS, 6% on Chrome/Android). Use [Appium on Saucelabs](https://saucelabs.com/automated_testing/ios-app-testing) to run some tests.

On the other hand, if we restrict to (apparently) NHS visits, we find nearly 100% on Windows, with 39% on IE10, 30% on Chrome, and 9% on IE8.

sebbacon:
Update on stats:

* IE8 is the second most popular IE with 1.18% of sessions overall. On its own this would make it hard to justify the effort of supporting it;
* However, it is the third most used browser in "NHS" sessions at 5.92% -- including IE9 too, that's nearly 9% of that traffic.  (Health warning: we're not exactly sure what "NHS network provider" means in the context of Google Analytics).
* If you look at people who interact with the website via email alerts (this includes emails being forwarded around, and people opening emails but not necessarily going to the website) then IE <= 9 is at about 3.5%

I would take the final measure with a pinch of salt because I think it may include HTML renderers within the email client. However, the second measure rings true.

That said, it's still diminishing returns actively supporting IE8/9 and gracefully removing functionality is probably the way to go.

brianmackenna:
I'd support this as well and we should write a blog explaining why we are doing it (if it hasn't been done already). And also why people should be using newer browsers, with links on how to upgrade.

As regards IE8/9 on NHS computers. My last NHS computer had IE10 or 11 as the default but also had up to date chrome and firefox. A lot of my colleagues just used the default that popped up so this might be what is happening. I actually had to roll my IE back to (I think) 8 in order to run certain programmes. I'm pretty sure it was ePACT1  so a lot of Meds Opt people would've had this setup.

PR #448, opened 2017-05-30
Add ability for end users to download data

### LINKED ISSUE #58: Re-add link to download raw data on analyse chart, add modal
Ben wants this, so we know who's using the data and can ensure we get credit in papers etc.

sebbacon:
Fixes #58 (though without modal)

### DIFF
--- a/openprescribing/media/css/analyse-layout.less
+++ b/openprescribing/media/css/analyse-layout.less
@@ -9,7 +9,7 @@
     width: 100%;
 
     li {
-      width: 33%;
+      width: 32%;
 
       @media (max-width: 1000px) {
         a {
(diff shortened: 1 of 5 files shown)


sebbacon:
Currently works in IE10 and IE11 but not IE9.

frabcus:
Is IE9 a requirement generally for these sites? Commonly enough used in the 
NHS?

In recent years, I've tended to only support IE11 as that is only one 
supported by Microsoft.

Francis

richiecroker:
According to Google Analytics 3.2% of visitors to our site are using IE8, and 1.98% are using IE9.  So only about 1 in 20.

sebbacon:
It may be that within the NHS we do need to support older IEs. See #371 for different stats.

sebbacon:
Per an update I just made to #371, it is likely that up to 10% of NHS users are using IE8/9. Given these are a core target group, I don't think we can just ignore them completely.

sebbacon:
Closing this PR in favour of #453

PR #453, opened 2017-06-08
Add ability for end users to download data

sebbacon:
This PR replaces #448 which had a funky branch name that made various command line operations difficult. See the discussion there.


sebbacon:
@lmfrench I've fixed it in IE10 and above and put a warning for older ones - can you check? thanks

PR #521, opened 2017-07-20
Creation of 17 NHSE low priority measures

richiecroker:


sebbacon:
@richiecroker  - FYI all measures must end in `.json` - I'll fix this.

sebbacon:
@richiecroker, I have just noticed the denominator descriptions are wrong for a number of the measures ([e.g. this one](https://github.com/ebmdatalab/openprescribing/blob/af16dc0489a485be5b711ddce8b8dad4e748f8bb/openprescribing/frontend/management/commands/measure_definitions/lptravelvacs#L14)); they all say "1000 patients" where some should be per section or subsection or whatever.

richiecroker:
@seb, they should all be per 1000 patients.

sebbacon:
This is now ready for review on staging.  Things to think about
* [ ] Precise wording for omnibus measure TBC, including how to describe the numerator / denominator
* [ ] Is the `sum(all_numerators)/sum(all_denominators)` approach correct? (I think so)
* [ ] Layout of CCG page. It's getting quite noisy with "labs", alert signups, and now this (screenshot below).  Ideas? 

image

lmfrench:
Add it as a tab where it says 'sort prescribing measures by' ? And add a sentence to the 'labs' page part instead of it being a separate blue box?

sebbacon:
For the layout issues, I'm going to do something "good enough" and propose a significant upgrade per #567 soon.

Issue #524, opened 2017-07-25
Make pipeline run_id of form YYYY_MM


inglesp:
We can then use this to validate that the right data has been fetched, and that it has been put into the right place on the filesystem.

PR #538, opened 2017-08-05
Specify prescribing month for runner

inglesp:
This addresses #524 .

I don't think it's quite right, but it's a good starting point for discussion.  Specifically, the year and month that we're now passing around everywhere in `runner.py` refer only to the year and month of the prescribing data.


inglesp:
Code Climate issues can be ignored.

PR #539, opened 2017-08-05
Add check that PracticeStatistics and Prescription tables are in sync

### LINKED ISSUE #523: Some kind of sanity check on list size data
A common error when updating the monthly data is for the patient list size data to somehow be out of sync with the prescribing data. 

There should be some kind of assertion that the last date in the patient list size table matches the last date in the prescribing table, and that both of these are for the expected date.

This should happen before any views or measures are computed.

inglesp:
Fixes #523

### DIFF
--- a/openprescribing/pipeline/management/commands/bigquery_upload.py
+++ b/openprescribing/pipeline/management/commands/bigquery_upload.py
@@ -1,14 +1,29 @@
 from __future__ import print_function
 
 from django.core.management import BaseCommand
+from django.db.models import Max
 
 from ebmdatalab import bigquery
+from frontend.models import PracticeStatistics, Prescription
 
 from ...cloud_utils import CloudHandler
 
 
 class Command(BaseCommand):
     def handle(self, *args, **kwargs):
+        # Make sure that PracticeStatistics and Prescription tables both have
+        # latest data.
+        latest_practice_statistic_date = PracticeStatistics.objects\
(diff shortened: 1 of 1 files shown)


openprescribing/pipeline/management/commands/bigquery_upload.py
```diff
@@ -1,14 +1,29 @@
...
 from ...cloud_utils import CloudHandler
 
 
 class Command(BaseCommand):
     def handle(self, *args, **kwargs):
+        # Make sure that PracticeStatistics and Prescription tables both have
+        # latest data.
+        latest_practice_statistic_date = PracticeStatistics.objects\
+            .aggregate(Max('date'))['date__max']
+        latest_prescrption_date = Prescription.objects\
```
sebbacon: can we rename the variable `latest_prescrption_date` -> `latest_prescription_date`

sebbacon:
Looks good - might as well fix that "line too long" pep8 thing though.

As for the other two codeclimate issues - it's a mystery to me how it sometimes decides there are new issues on non-new code...

inglesp:
The long line's caused by a URL in some help text, and it can't be fixed without making the help text look silly.

sebbacon:
I beg to differ / don't quite follow but happy to go with your judgement :-)

inglesp:
Argh, I thought you were talking about the help text in #544.  Sorry about that.

PR #540, opened 2017-08-05
Validate that manually-fetched file is in right place

### LINKED ISSUE #516: Add validation that file is in the right place after manual fetch
After running a manual fetch, the runner could check that an importable file is has been put in the right place.

inglesp:
Fixes #516

### DIFF
--- a/openprescribing/pipeline/runner.py
+++ b/openprescribing/pipeline/runner.py
@@ -159,8 +159,10 @@ class ManualFetchTask(Task):
             print('The following files have been manually fetched:')
             for path in unimported_paths:
                 print(' * {}'.format(path))
-            raw_input('Press return to confirm, or Ctrl+C to cancel '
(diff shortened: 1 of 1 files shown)


inglesp:
I'm not sure what might be useful.  The problems would be one of:

 * file has wrong name
 * file is in a directory belonging to another source
 * file is in a directory with the wrong date

In the first two cases, the file would not be listed by this check, and the user should notice this and be able to work out what's gone wrong.

In the third case, there's no way of the runner knowing what the correct date should be, so this check acts as a trigger to double check what's been done.

Do you have any suggestions?

sebbacon:
I guess just something along the lines of "Files were expected at the following locations but could not be found. Check the file names and directories (including that the datestamp you used is correct)"..?

inglesp:
Good point, I've addressed that.

Issue #517, opened 2017-07-19
Update notes for bnf_codes task


inglesp:
At the moment it says:

    We currently use a version that is correct as of Jan 2016; we should
    probably be importing this quarterly or so. There are monthly changes
    but not that many.

This is not true.

inglesp:
Closed by #544.

PR #544, opened 2017-08-08
Mostly automate fetching of bnf_codes.csv

### LINKED ISSUE #515: (Mostly) automate fetching BNF CSV file
We can use the same approach as [`fetch_prescribing_data.py`](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/pipeline/management/commands/fetch_prescribing_data.py).

inglesp:
Fixes #515 and #517.


inglesp:
Code climate fails because of a long line in the command's help, and because of a couple of totally unrelated things.  It's not letting me mark these as to be ignored.

PR #548, opened 2017-08-09
Smoketests as tasks

inglesp:


inglesp:
Code Climate failures are not relevant to this PR.

PR #559, opened 2017-08-15
BQ refactor

inglesp:
Creating a PR now, as I'd like to check that tests continue to pass.


sebbacon:
This is great. Lovely and readable (actually, easier to read as one big diff than loads of commits) and just what I hoped for (indeed, better!)

Aside from the inline comments:

* We need to delete, deprecate or otherwise retire the original `ebmdatalab` package. Just a big bold title in the README, I guess
* As you mention in `settings`, you've moved it all to a Django app called `ebmdatalab` which doesn't feel satisfactory. 
   (a) Should it actually be a Django app? Isn't it more just a library, and therefore perhaps belongs in `common/` or similar? (I don't have strong feelings about this, but it doesn't feel like an app to me)
  (b) As for name, how about `google_cloud_utils` or similar?

inglesp:
"Aside from the inline comments" -- I don't see these.

inglesp:
Have renamed `ebmdatalab` to `gcutils`.  It needs (I think) to be a Django app, since it has tests and the Django test runner only runs tests of things in `INSTALLED_APPS`.

inglesp:
The only thing left, I think, is a couple of `TODO`s to resolve.

inglesp:
Test failure is unrelated, and codeclimate is being whingy.

Issue #192, opened 2016-09-13
Make raw number searches available through OpenPrescribing


annapowellsmith:
User story from Lisa's testing:
- **As a** commissioner
- **I want** to see (and download) the raw numbers for prescribing of drug X
- **So that** I know whether it's a large enough area to be worth working on. 

Richard Croker has also asked for this

lmfrench:
I think this is a mix of #58 and a separate issue to allow users to search without selecting a denominator

richiecroker:
My view exactly what @lmfrench described above.   Just a quick look at raw numbers, at practice or ccg level.  No denominators, just pure numbers.  Doesn't have to even been graphical, could we use the table that we create for the breakdown?

PR #583, opened 2017-09-21
Searches without denominator #192

sebbacon:


sebbacon:
How does this look @lmfrench?

lmfrench:
Yeah I like it! It might not be obvious to current users that we have removed the requirement for a denominator so could put (optional) by versus

sebbacon:
Superseded by #645

PR #584, opened 2017-09-21
Truncate very long denominator/numerator titles in charts

### LINKED ISSUE #254: Wrap text box for information on graphs
If you're searching a number of different items, the hover text box showing the information goes over the side of the box e.g.

image

This means you can't see the numbers, so can't be sure how big a problem it is.

Can we fix this by wrapping the text, or maybe by saying something like (numerator) and (denominator)

sebbacon:
Fixes #254

### DIFF
--- a/openprescribing/media/js/src/chart_formatters.js
+++ b/openprescribing/media/js/src/chart_formatters.js
@@ -1,4 +1,5 @@
 var _ = require('underscore');
+var humanize = require('humanize');
 
 var formatters = {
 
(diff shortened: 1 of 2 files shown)


sebbacon:
Per the original issue, I don't think wrapping is a good solution as it will always be very hard to read. I prefer truncation, as you can see the full list in the original form: 
image

sebbacon:
@HelenCEBM suggested 

> Tableau will shorten long strings of selected items to "A, B, C and X more", would that be an improvement?

We've agreed it probably is, but this is Good Enough for now.

PR #589, opened 2017-09-27
[Not for merging] BQ refactor spike

inglesp:
This PR is not for merging, but contains the commits from my first go at refactoring BQ.  The code is more-or-less as they will be in the final PR (#559), but commits need to be broken up to help with reviewing.

I've created a new PR because I find the GitHub UI better than anything I've got locally.


inglesp:
We can close this!!

PR #591, opened 2017-09-27
Break down denominators

### LINKED ISSUE #491: Create a "what does this mean" resource for GPs
Currently we provide a number of standard measures to GP practices and CCGs showing their performance.  What these measures don't show, however, is the specific drugs in the numerator that mean the practices/CCG are "poor".

It would be good to have a link on the measures page to show exactly the 10 highest items so that GPs know what to review.  It's also a useful check to make sure that the coding of drugs is correct.

It would be good if we could get it to be inline, looking a bit like this:
image

although a simple CSV download would be OK at this stage.

We've created some SQL for individual measures, e.g.
```sql
SELECT
  practice,
  bnf_name,
  SUM(items) AS items,
  sum(actual_cost) AS actual_cost
FROM (
  SELECT
    practice,
    bnf_name,
    SUM(items) AS items,
    sum(actual_cost) AS actual_cost,
    ROW_NUMBER() OVER (PARTITION BY practice ORDER BY SUM(items) DESC) AS pos
  FROM
    ebmdatalab.testdata.RC_prescribing_2016_Q4
  WHERE
    bnf_code LIKE '0407020Q0%AI'
    OR bnf_code LIKE '0407020Q0%DC'
    OR bnf_code LIKE '0407020Q0%EH'
    OR bnf_code LIKE '0407020Q0%DU'
    OR bnf_code LIKE '0407020Q0%AH'
    OR bnf_code LIKE '0407020Q0%DD'
    OR bnf_code LIKE '0407020Q0%EB'
    OR bnf_code LIKE '0407020Q0%DV'
    OR bnf_code LIKE '0407020Q0%DW'
    OR bnf_code LIKE '0407020Q0%AG'
    OR bnf_code LIKE '0407020Q0%DE'
    OR bnf_code LIKE '0407020Q0%EI'
    OR bnf_code LIKE '0407020AD%AR'
    OR bnf_code LIKE '0407020AD%AH'
    OR bnf_code LIKE '0407020AD%AI'
    OR bnf_code LIKE '0407020AD%AP'
    OR bnf_code LIKE '0407020AD%AQ'
    OR bnf_code LIKE '0407020AF%AD'
    OR bnf_code LIKE '0407020AG%AF'
    OR bnf_code LIKE '0407020AG%AG'
    OR bnf_code LIKE '0407020A0%AF'
    OR bnf_code LIKE '0407020A0%AG'
    OR bnf_code LIKE '0407020A0%AH'
    OR bnf_code LIKE '0407020B0%AF'
    OR bnf_code LIKE '0407020B0%AG'
  GROUP BY
    practice,
    bnf_name )
WHERE
  pos <=10
  and 
GROUP BY
  practice,
  bnf_name
ORDER BY
  practice,
  items DESC
```
It would be good to know if we could integrate somehow into the JSON for the measures so we ensure consistency in the BNF codes.

sebbacon:
Fixes #491


lmfrench:
Tested on Chrome, FireFox & IE11

A few small issues: 

- The text on table continues to say 'Loading' even after everything has loaded (when the chart is actually loading, 'Loading' appears twice)
image

- Can't sort by Presentation Name

- Could add a link to it on the 'all practices' pages for the measures too under 'How this CCG is doing': e.g. http://staging.openprescribing.net/ccg/02Q/dipyridamole/

PR #602, opened 2017-10-11
Add task to back up prescribing table

### LINKED ISSUE #580: Back up prescribing table monthly
As part of the pipeline, we should have a task that exports the prescribing table to google cloud.

I would suggest writing this so it keeps 6 months of historic data. Or possibly, given it's so cheap, indefinite amounts...

inglesp:
Fixes #580.


sebbacon:
Looks good. Strange error in Travis though...

inglesp:
Hm, flaky tests.  Re-ran the build, and it passed.

PR #608, opened 2017-10-17
Add fetcher and scraper for NCSO concessions

### LINKED ISSUE #599: Import NCSO data monthly
Monthly shortage lists are published [here](http://psnc.org.uk/dispensing-supply/supply-chain/generic-shortages/).  We would like the ability to mark any presentation in any given month as in that list, along with its price.

This could be achieved by updating a table, monthly, linked to our dmd tables (see below).

A presentation is defined by the generic name + dose + form + pack size, i.e. the first two columns of that list.

## Relevant resources

* We should probably link this to our the DMD data we import, if possible. An alternative is to link it to BNF codes but the huge difference in naming in the BNF scheme is likely to make this much harder.
  * In our database, the DMD tables are all prefaced `dmd_` (and some but not all have corresponding models in the `dmd` app in `openprescribing`).
  * The entity we'd map with is `dmd_vmpp`, so we could make a table with a foreign key on `dmd_vmpp.vppid` for the NCSO data. The `nm` field should be relatively matchable
* @richiecroker has done some quick analysis by hand and says the majority match very well, and a minority don't - so we should do that a bit more systematically and see if we can fix the misses herustically
  * if not, we need to devise a workflow involving google sheets for someone to (e.g.) copy/paste and fix by hand monthly
* online version of DMD [here](http://dmd.medicines.org.uk/)

Here's a sample query against the db on `largeweb2` illustrating a name match:
```
127.0.0.1 prescribing@prescribing=> select * from dmd_vmpp where nm like '%Terbinafine 250mg tablets%';
┌───────────────────┬─────────┬─────────────────────────────────────┬──────────┬───────────┬────────┬───────────┬────────────┐
│       vppid       │ invalid │                 nm                  │ abbrevnm │   vpid    │ qtyval │ qty_uomcd │ combpackcd │
├───────────────────┼─────────┼─────────────────────────────────────┼──────────┼───────────┼────────┼───────────┼────────────┤
│ 18490211000001108 │       ¤ │ Terbinafine 250mg tablets 8 tablet  │ ¤        │ 324718007 │      8 │ 428673006 │          ¤ │
│ 14755511000001101 │       ¤ │ Terbinafine 250mg tablets 16 tablet │ ¤        │ 324718007 │     16 │ 428673006 │          ¤ │
│  1122811000001109 │       ¤ │ Terbinafine 250mg tablets 28 tablet │ ¤        │ 324718007 │     28 │ 428673006 │          ¤ │
│  1313611000001101 │       ¤ │ Terbinafine 250mg tablets 14 tablet │ ¤        │ 324718007 │     14 │ 428673006 │          ¤ │
└───────────────────┴─────────┴─────────────────────────────────────┴──────────┴───────────┴────────┴───────────┴────────────┘
```

inglesp:
There are some outstanding issues:

* For any concessions whose drug name cannot be matched against dm+d
  data, we create a NCSOConcession object but do not match it against a
  DMDVmpp object.  We should decide what to do here.
* The NCSOConcession model has a price_concession_pence attribute.  Is
  this the right way to record prices?
* The DMDVmpp table already exists in the production database, so the
  migration that creates it should be skipped in production.  (Unless
  there's a better solution...)

Fixes #599.


inglesp:
Using the Django admin for reconciliation isn't going to work, as there are ~30k entries in the dmd_vmpp table to match against.

A command line tool would be simplest to write.  It would look something like:

    $ ./manage.py reconcile_ncso_concessions

    There are 7 unreconciled NCSO concessions
    ~~~
    
    Unreconciled concession:
        drug: Duloxetine 40mg gastro-resistant capsules
        pack_size: 56

    Enter search term (case insensitive):
    > duloxetine

    Found 6 matching VMPPs:

     1. Duloxetine 20mg gastro-resistant capsules 28 capsule
     2. Duloxetine 20mg gastro-resistant capsules 56 capsule
     3. Duloxetine 30mg gastro-resistant capsules 28 capsule
     4. Duloxetine 40mg gastro-resistant capsules 56 capsule
     5. Duloxetine 60mg gastro-resistant capsules 28 capsule
     6. Duloxetine 60mg gastro-resistant capsules 84 capsule

    Enter number of matching VMPP, or 0 to search again:
    > 4

    Matching against:
        Duloxetine 40mg gastro-resistant capsules 56 capsule

    Please confirm [yN]:
    > y

    There are 6 unreconciled NCSO concessions
    ~~~

    ...

Thoughts?

sebbacon:
I think this would be great. The only question is - who would be doing this? Making it a django management command kind of means it would have to be you or me, or we train others in a whole new infrastructure.

Right now, the only alternative I can think of is to add DMD search to the API, and then make a noddy frontend onto that.

There are advantages to this - exposing DMD (and our mappings) would be useful for the community more widely, and could feed into improvements in our UI.  

On the other hand, it is more work.

On the other other hand, if it's only for our benefit, it could be a really noddy frontend and might not be *that* much work.

Let's discuss with voices soon?

sebbacon:
On the subject of a web ui, a8450d5 was a (discarded) attempt to add a useful API endpoint which included para, section, chapter in the responses, and used DMD names. Might be handy if we go down that route.

inglesp:
Have added a management command for manual reconciliation in 4f87774.

I think it would make sense for one of us to do it after an import.  It shouldn't be a massive job, and as it's almost always going to be simplish string matching, it shouldn't require any pharmacological expertise.

sebbacon:
OK. How should we enforce it in the pipeline?

inglesp:
@sebbacon Could you review again please?

The codeclimate complaints are about long lines in migrations.

sebbacon:
Looks great! I wonder if we should merge this now or have a conversation about how we actually use the data and include that in this PR.

I'll talk with @richiecroker today about what we'll do with it...

sebbacon:
We should merge this ASAP and run the matcher as @richiecroker wants to do some research with the data in any case.

PR #611, opened 2017-10-17
Run yapf over codebase

inglesp:
This is just an experiment to see what using an automated formatter would do to our code.

I've chosen [yapf](https://github.com/google/yapf) because I haven't heard bad things about it.

I'm not aghast at the choices that it's made (it's certainly made some improvements) and I think I could get used to running this automatically as a pre-commit hook.  I could certainly get used to not worrying about code formatting myself.

The only downside I see is that it makes the git history a bit harder to deal with.

Thoughts @sebbacon, @frabcus?


sebbacon:
In general the changes improve readability, but mostly not in cases where the readability was already particularly poor.

My only concern is that these changes are a superset of flake8 but there is (as far as I can see) no flycheck integration for the ruleset, which means I can't learn it (flycheck+flake8 alerting me as I type). Thus there is no option other than always running it in a precommit hook (or similar) which means you will always get ugly stuff in your git commit history.

On the other hand, I think it is easy to fetishise git commit history. I use it maybe twice a month, and even with messy commit chains it rarely takes me more than 10 mins to answer any question I might have.

So, having thought through it aloud, I'm on the fence, but maybe leaning slightly to "yes": I'm relaxed about commit history and in favour of readability. I'm just slightly uncomfortable with never learning the style that the tool that I'm using enforces.

inglesp:
Closing in favour of #677.

PR #613, opened 2017-10-17
Run tests under coverage

inglesp:


inglesp:
Builds fail because `flake8` exits with non-zero, I think.

sebbacon:
Where does the report end up and how shall we incorporate it into our workflow? Fail if coverage %age decreases?

inglesp:
This is ready to merge.  The next step is to enable something like Coveralls[https://coveralls.io/] to report coverage as a comment on each PR.  Later on we can decide whether to make a reduction in coverage a build failure.  I think that only @sebbacon can enable Coveralls.

PR #617, opened 2017-11-01
Make hscic_prescribing a no-op

inglesp:
The data has moved but we don't need to import it right now.


inglesp:
The failed test is a flaky one, so I'll merge and deploy.

Issue #620, opened 2017-11-02
August alert mailouts failed


sebbacon:
This tends to happen every month, but usually only one or two mailouts are affected. Check how many were affected this month, then make improvements to code so we don't have to check so manually every time.

```
(.venv)hello@largeweb2:/webapps/openprescribing/openprescribing$ python manage.py send_monthly_alerts --settings=openprescribing.settings.production
Traceback (most recent call last):
  File "manage.py", line 12, in <module>
    execute_from_command_line(sys.argv)
  File "/webapps/openprescribing/.venv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/webapps/openprescribing/.venv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/webapps/openprescribing/.venv/local/lib/python2.7/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/webapps/openprescribing/.venv/local/lib/python2.7/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/webapps/openprescribing/openprescribing/frontend/management/commands/send_monthly_alerts.py", line 161, in handle
    email_retrier.try_email(callback)
  File "/webapps/openprescribing/openprescribing/frontend/management/commands/send_monthly_alerts.py", line 204, in __exit__
    self.exceptions[-1][2])
frontend.management.commands.send_monthly_alerts.BatchedEmailErrors: Encountered 4 mail exceptions (showing last traceback only): `CalledProcessError: Command '/usr/local/bin/phantomjs /webapps/openprescr
ibing/openprescribing/frontend/management/commands/grab_chart.js "https://openprescribing.net/ccg/11X/" /mnt/database/tmp/tmpHaHxE2.png "#lplidocaine" 1024x1024 1000' returned non-zero exit status 1, Call
edProcessError: Command '/usr/local/bin/phantomjs /webapps/openprescribing/openprescribing/frontend/management/commands/grab_chart.js "https://openprescribing.net/ccg/02Q/" /mnt/database/tmp/tmpGNRkh5.png
[...]

PR #621, opened 2017-11-02
Avoid N+1-type query issue in measures

sebbacon:
The previous implementation involved 2400+ separate select queries,
just to load the Measure model for a single field.

This is also more readable.

Fixes part of #620

### DIFF
--- a/openprescribing/api/views_measures.py
+++ b/openprescribing/api/views_measures.py
@@ -4,6 +4,7 @@ from rest_framework.response import Response
 
 from frontend.models import ImportLog
 from frontend.models import Measure
+from frontend.models import MeasureGlobal
 
 import view_utils as utils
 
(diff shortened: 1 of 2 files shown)


openprescribing/api/views_measures.py
```diff
@@ -16,62 +17,52 @@ class MissingParameter(APIException):
...
-    query += 'mg.denominator, mg.measure_id, '
-    query += 'mg.calc_value, mg.percentiles, mg.cost_savings, '
-    query += 'ms.name, ms.title, ms.description, '
-    query += 'ms.why_it_matters, ms.tags_focus, '
-    query += ' ms.denominator_short, ms.numerator_short, '
-    query += 'ms.url, ms.is_cost_based, ms.is_percentage, '
-    query += 'ms.low_is_good '
-    query += "FROM frontend_measureglobal mg "
-    query += "JOIN frontend_measure ms ON mg.measure_id=ms.id "
+    tags = request.query_params.getlist('tags', None)
```
inglesp: I don't think this does what you think it does.

If a request is to `/url/?q=1,2,3` then `request.GET.getlist('q')` will give `['1,2,3']`.  But it looks like you're expecting it to give `['1', '2', '3']`.

sebbacon: aha, thanks

PR #624, opened 2017-11-07
Ensure measures are parsed in alphabetical order

### LINKED ISSUE #619: August LP omnibus data looks wrong
image

The thousand patients are too low, e.g. dropping from 170 to 66

inglesp:
Fixes #619


sebbacon:
Hmm. The test isn't passing! But it all looks fine to me. I assume `list(orderedlist)` does what one would expect....

PR #629, opened 2017-11-08
Core measures only

sebbacon:
Alerts were running across all measures. This was causing breakage in the alert-sending process. Also, Low Priority alerts are still considered "labs" so we won't include them directly (only via the omnibus version) for the time being.


sebbacon:
Bad branching. Closing.

PR #632, opened 2017-11-08
Remove {% if not debug %} from base.html

inglesp:
As far as I can tell, debug is never set in a template's context.  This
has been in the code since day one, and is responsible for a lot of
noise in the test output on Travis, where we're logging everything at
debug level.

(Also: do we ever use errorception?)


sebbacon:
Errorception also removed.  In theory this is covered in new relic, though in practice I've not had time to address any of the issues there, either.

PR #647, opened 2017-11-19
Refactor convert hscic prescribing

inglesp:


inglesp:
Merging after discussion with Seb.

PR #672, opened 2017-11-29
Improve use of raw SQL in API views

inglesp:


openprescribing/api/views_measures.py
```diff
@@ -149,144 +162,141 @@ def measure_numerators_by_org(request, format=None):
...
+        qs = qs.filter(measure_id=measure_id)
+
+    if tags:
+        valid_tags = ['core', 'lowpriority']
+        clauses = []
+        for tag in tags:
+            assert tag in valid_tags
+            clauses.append('{} = ANY(frontend_measure.tags)'.format(tag))
+        extra_sql = ' AND '.join(clauses)
+        qs.qs.extra(where=[extra_sql])
```
sebbacon: Is this right? `qs.extra`, no?

inglesp: Oof, no.  It should be `qs = qs.extra(...)`.

openprescribing/api/views_measures.py
```diff
@@ -149,144 +162,141 @@ def measure_numerators_by_org(request, format=None):
...
+            pct__close_date__isnull=True,
+            practice_id__isnull=True,
+        ).\
+        order_by('pct_id', 'measure_id', 'month')
+
+    if measure_id:
+        qs = qs.filter(measure_id=measure_id)
+
+    if tags:
+        valid_tags = ['core', 'lowpriority']
```
sebbacon: Could probably not bother for now, but given (a) it's repeated and (b) we have a live plan to add more tags, it would be nicer to either have this in settings, or check against values in `Measure.objects.values_list('tags').distinct()` or whatever it would be.

I'm worried we won't pick this up in tests when we add new tags and everything will break.

inglesp: Good point.  I think we need better testing here full stop.

inglesp:
This is ready to merge.  Code Climate is complaining about complexity.

PR #677, opened 2017-12-05
Pre commit hooks

inglesp:
Set up various pre-commit hooks.

Each developer will need to run `pip install -r requirements/local.txt` followed by `pre-commit install` to get the benefit of the hooks.

If it turns out any hooks are unhelpful, we can remove them!


inglesp:
I'm going to merge this and try it out myself for a few days, and then we can decide whether we should encourage everyone (ie @sebbacon) to use it too.

PR #674, opened 2017-11-30
Tariff explorer

sebbacon:
Closes #671

PR #679, opened 2017-12-06
Tariff explorer

sebbacon:
Cleaned up version of #674


openprescribing/dmd/management/commands/bulk_import_drug_tariff.py
```diff
@@ -0,0 +1,72 @@
...
+class Command(BaseCommand):
+    args = ''
+    help = 'Imports a CSV of historic tariff prices'
+
+    def add_arguments(self, parser):
+        parser.add_argument('--csv')
+
+    def handle(self, *args, **options):
+        '''
+        '''
```
inglesp: No need for an empty docstring.

openprescribing/dmd/management/commands/bulk_import_drug_tariff.py
```diff
@@ -0,0 +1,72 @@
...
+        with open(options['csv'], 'r') as f:
```
inglesp: `'r'` is redundant here.

openprescribing/dmd/management/commands/bulk_import_drug_tariff.py
```diff
@@ -0,0 +1,72 @@
...
+            with transaction.atomic():
+                month = None
+                counter = 0
+                for row in csv.DictReader(f):
```
inglesp: Unless the file is massive, I'd prefer:

```
with open(path) as f:
    rows = list(csv.DictReader(f)

for row in rows:
    ...
```

This reduces nesting.

openprescribing/dmd/management/commands/bulk_import_drug_tariff.py
```diff
@@ -0,0 +1,72 @@
...
+                    month = datetime.strptime(row['Month'], '%d/%m/%Y')
+                    counter += 1
+                    if 'Category A' in row['DT Cat']:
+                        tariff_category = 1
+                    elif 'Category C' in row['DT Cat']:
+                        tariff_category = 3
+                    elif 'Category M' in row['DT Cat']:
+                        tariff_category = 11
+                    else:
+                        raise
```
inglesp: A bare `raise` isn't quite valid Python -- it'll actually raise a `TypeError`.  (You can take the coder out of Ruby, etc etc.)  It's harmless here, but a pattern I'm finding myself using is `assert False, msg`, where `msg` is something useful -- in this case, probably the value of `row['DT Cat']`.

openprescribing/dmd/management/commands/fetch_drug_tariff.py
```diff
@@ -0,0 +1,97 @@
...
+                 'drug-tariff/drug-tariff-part-viii/')
+        rsp = requests.get(index)
+        doc = bs4.BeautifulSoup(rsp.content, 'html.parser')
+
+        month_names = [x.lower() for x in calendar.month_name]
+
+        for a in doc.findAll('a', class_='excel', title=re.compile('VIIIA')):
+            words = re.split(
+                r'[ -]+',
+                urllib.unquote(os.path.splitext(os.path.basename(a.attrs['href']))[0]))
```
inglesp: An example of what `a.attrs['href']` and `words` are would be helpful.

openprescribing/dmd/management/commands/fetch_drug_tariff.py
```diff
@@ -0,0 +1,97 @@
...
+            )
+
+            if os.path.exists(dir_path):
+                continue
+
+            mkdir_p(dir_path)
+            xls_url = urljoin(index, a.attrs['href'])
+            xls_file = StringIO(requests.get(xls_url).content)
+            xls = pd.read_excel(xls_file, skiprows=2).dropna()
+            # drop rows with nulls
```
inglesp: This comment belongs to the line above, doesn't it?

openprescribing/dmd/management/commands/fetch_drug_tariff.py
```diff
@@ -0,0 +1,97 @@
...
+                year_and_month
+            )
+
+            if os.path.exists(dir_path):
+                continue
+
+            mkdir_p(dir_path)
+            xls_url = urljoin(index, a.attrs['href'])
+            xls_file = StringIO(requests.get(xls_url).content)
+            xls = pd.read_excel(xls_file, skiprows=2).dropna()
```
inglesp: Could we call this `df` not `xls`, as it's a `DataFrame`?

openprescribing/dmd/management/commands/import_drug_tariff.py
```diff
@@ -0,0 +1,71 @@
...
+                        date=date,
+                        vmpp_id=vmpp,
+                        product_id=product_id,
+                        tariff_category_id=dt_cat,
+                        price_pence=price_pence
+                    )
+                ImportLog.objects.create(
+                    category='tariff',
+                    filename=kwargs['filename'],
+                    current_at=date)
```
inglesp: Similar comments to bulk_import_drug_tariff.py.

openprescribing/dmd/migrations/0004_auto_20171129_1341_squashed_0008_auto_20171129_1354.py
```diff
@@ -0,0 +1,48 @@
+# -*- coding: utf-8 -*-
```
inglesp: A small thing for neatness: unless this has already been applied, you could delete this and the next migration and re-run `makemigrations`.  You'd need to roll back any migrations locally.

openprescribing/dmd/models.py
```diff
@@ -164,6 +164,10 @@ class DMDVmpp(models.Model):
     qty_uomcd = models.BigIntegerField(blank=True, null=True)
     combpackcd = models.BigIntegerField(blank=True, null=True)
 
+    @property
+    def products(self):
+        return DMDProduct.objects.filter(vpid=self.vpid)
```
inglesp: I'm a bit wary of putting database access behind a `@property`.  Would it be possible to add an explicit `ForeignKey`?

openprescribing/dmd/models.py
```diff
@@ -183,3 +187,28 @@ def unreconciled(self):
...
+    vmpp = models.ForeignKey(DMDVmpp)
+    product = models.ForeignKey(DMDProduct)
+    # 1: Category A
+    # 3: Category C
+    # 11: Category M
+    tariff_category = models.ForeignKey(TariffCategory)
+    price_pence = models.IntegerField()
+
+    @property
+    def concession(self):
```
inglesp: We can't have a foreign key here, but could you change the property to a method called `get_concession`?

openprescribing/media/js/src/tariff-charts.js
```diff
@@ -0,0 +1,265 @@
...
+      // style found
+    };
+    return hasStyle;
+  },
+
+  addDummySeriesForCategoryLabels: function(data) {
+    // Given complete data series, return an array of strings
+    // indicating any DT Categories that have been used in any of the
+    // series. We use this array to decide which extra legend items to
+    // display (e.g. to indicate that a dotted line means "Categoriy
```
inglesp: s/Categoriy/Category/

openprescribing/dmd/management/commands/bulk_import_drug_tariff.py
sebbacon: agreed but ignoring as the entire file becomes redundant after deploy!

openprescribing/dmd/management/commands/bulk_import_drug_tariff.py
sebbacon: Nice!

openprescribing/dmd/models.py
sebbacon: Actually, it doesn't seem to be used anywhere, so I'll delete it

PR #682, opened 2017-12-07
Add command for fetching and importing all NCSO concessions

inglesp:


inglesp:
This is ready for merging.  The Code Climate issues are about complexity (can we turn these off?) and a couple of long URLs (which I'm relaxed about).

openprescribing/dmd/tests.py
```diff
@@ -84,53 +89,83 @@ def test_import_dmd_snomed(self):
...
+        base_path = os.path.join(settings.SITE_ROOT, 'dmd', 'tests', 'pages')
+
+        with open(os.path.join(base_path, 'ncso-archive.html')) as f:
+            archive_doc = bs4.BeautifulSoup(f.read(), 'html.parser')
+
+        with open(os.path.join(base_path, 'ncso-current.html')) as f:
+            current_doc = bs4.BeautifulSoup(f.read(), 'html.parser')
+
+        patch_path = 'dmd.management.commands.fetch_and_import_ncso_concessions'
+        with nested(
```
sebbacon: Isn't the `nested` redundant here?

inglesp: Don't think so -- see http://www.voidspace.org.uk/python/mock/patch.html#nesting-patch-decorators.  (That said, I find the `mock` API pretty confusing so I may be missing a simpler solution.)

sebbacon: I'm pretty sure that's out of date - in Python 2.7 you can do `with X as a, Y as b:`.

From [the python docs](https://docs.python.org/2/library/contextlib.html#contextlib.nested):
> This function has been deprecated in favour of the multiple manager form of the with statement.

PR #689, opened 2017-12-13
Refactor javascript 

sebbacon:
Build process:
- Implement browserification and related processes as javascript to improve readability and reduce code duplication
- More consistent debugging output in `watch` mode

Javascript organisation:
- Move all JS source to `src/` subdirectory
- expose jquery as `$` using browserify-shim
- prepend all analyse-form-related modules with `analyse-`
- use `domready` npm package instead of jquery `document.ready` idiom
- delete unused scatter-chart module
- only include javascript for specific pages on pages to which it pertains
- factor out common javascript to `common.js` module (using a browserify plugin)


sebbacon:
Things to test:

* [ ] Dropdown menu
* [ ] Doorbell popover
* [ ] Measures (by CCG, by measure, by practice, etc)
* [ ] Analyse form (all charts)
* [ ] Filtering by practice name

openprescribing/media/js/package.json
```diff
@@ -44,6 +44,9 @@
       "jquery": true,
       "browser": true,
       "mocha": true
+    },
+    "rules": {
+      "no-var": 0
```
inglesp: Why's there duplication between this and `eslintrc.json`?

sebbacon: This version is for your own eslint which you would install using npm per the README.  The other version is for travis per .travis.yml.

sebbacon:
I suspect the failing travis test could be a race condition, so have [turned off concurrency](https://travis-ci.org/ebmdatalab/openprescribing/settings) to test this.

If that doesn't fix it, then I should probably see about rebuilding the base image.

Issue #698, opened 2017-12-18
BQ resource limits mean sorting prescribing dataset is impossible


sebbacon:
In #663, we sort the prescribing data as a easy alternative to CLUSTERing it in postgres on load.  This is critical for ensuring that our analyse queries are performant.

However, it turns out that BigQuery resource limits mean this isn't possible:

```
convert_hscic_prescribing --filename /home/hello/openprescribing-data/data/prescribing/2017_10/20171217_1513507724667_Detailed_Prescribing_Information.csv
...
 File "/webapps/openprescribing/openprescribing/frontend/management/commands/convert_hscic_prescribing.py", line 154, in handle
    fmtd_data_table.insert_rows_from_query(sql, legacy=True)
...
gcutils.bigquery.JobError: [{u'reason': u'resourcesExceeded', u'message': u'Resources exceeded during query execution: The query could not be executed in the allotted memory. ORDER BY operator used too much memory..'}]
```
BigQuery documentation is inadequate to answer the question "how big a dataset can be sorted"? The best I can find is [this](https://cloud.google.com/bigquery/docs/best-practices-performance-output#use_a_limit_clause_with_large_sorts) which suggests using `LIMIT` to reduce the size of the dataset.

sebbacon:
To debug possible in-database solutions, I've saved the following query (the failing query minus the sort) `tmp_eu.issue_698`:

```
SELECT
  LEFT(PCO_Code, 3) AS pct_id,
  Practice_Code AS practice_code,
  BNF_Code AS presentation_code,
  SUM(Items) AS total_items,
  SUM(NIC) AS net_cost,
  SUM(Actual_Cost) AS actual_cost,
  SUM(Quantity * Items) AS quantity,
  '2017-10-01' AS processing_date,
FROM
  tmp_eu.raw_prescribing_data_2017_10
WHERE
  Practice_Code NOT LIKE '%%998'  -- see issue #349
GROUP BY
  presentation_code,
  pct_id,
  practice_code
```

sebbacon:
It has  9948733 rows, the query takes 3.19 GB of processing.

![image](https://user-images.githubusercontent.com/211271/34101028-43254bac-e3dc-11e7-95b7-46504c18740e.png)

sebbacon:
So we can reproduce the problem by attempting to ORDER the temporary table I created above.

```sql
SELECT
  *
FROM
  [ebmdatalab:tmp_eu.issue_698]
ORDER BY
[...]

PR #712, opened 2018-01-03
Don't use BQ for sorting in create_views

### LINKED ISSUE #711: Sorting in create_views fails
When running `create_views`, we try to sort a large amount of data in BQ.  This fails because of resource limitations.

For the last import (2017_10), we finessed things by doing the sort on the filesystem by hand.

inglesp:
Fixes #711.

### DIFF
--- a/openprescribing/frontend/management/commands/create_views.py
+++ b/openprescribing/frontend/management/commands/create_views.py
@@ -2,6 +2,7 @@ from multiprocessing.pool import Pool
 import glob
 import logging
 import os
+import subprocess
 import tempfile
 import traceback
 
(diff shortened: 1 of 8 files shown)


openprescribing/frontend/management/commands/create_views.py
```diff
@@ -88,17 +90,39 @@ def fill_views(self):
...
+            settings.LARGE_TMP_DIR,
+            '{}-sorted.csv'.format(filename_base)
+        )
+
+        self.log('Downloading {} to {}'.format(table.name, raw_path))
+        with open(raw_path, 'w') as f:
             exporter.download_from_storage_and_unzip(f)
-            f.seek(0)
 
+        self.log('Sorting {} to {}'.format(table.name, sorted_path))
```
sebbacon: We should add a comment here explaining why we sort and why we do it this way. e.g. with a link to https://github.com/ebmdatalab/openprescribing/issues/698 (or just a summary of it).

openprescribing/frontend/management/commands/create_views.py
```diff
@@ -88,17 +90,39 @@ def fill_views(self):
...
+            self.download_and_import(table, prescribing_date)
             self.log("-------------")
 
-    def download_and_import(self, table):
+    def download_and_import(self, table, prescribing_date):
         storage_prefix = 'hscic/views/{}-'.format(table.name)
         exporter = TableExporter(table, storage_prefix)
 
-        with tempfile.NamedTemporaryFile(mode='r+') as f:
+        filename_base = '{}-{}'.format(table.name, prescribing_date)
```
sebbacon: Simply using `tempfile` should work, because we set `export TMPDIR=/mnt/database/tmp/` in the production environment.

As the location of large volumes is environment-specific, I think that's the right way of doing this?

(Though ISTR there was an issue with a disk filling up when you were trying this out before...?)

PR #722, opened 2018-01-10
Upload various DMD tables to BQ

inglesp:


inglesp:
Code climate warning are spurious.

PR #750, opened 2018-01-30
Warn when NCSO concession has been removed from source

inglesp:


inglesp:
Failing tests come from hasty merge of #748.  Irrelevant here.

PR #751, opened 2018-01-30
Don't try to render all tariffs on page load

### LINKED ISSUE #747: Tariff page very slow
We have intermittently seen the tariff page grind to a halt today.

I have also seen it look like it's graphing every single tariff:

image

inglesp:
Fixes #747 properly.


inglesp:
@sebbacon Could you look at this?  There might be a nicer way to fix the problem.

Issue #701, opened 2017-12-21
Routinely load latest month's data into BigQuery


HelenCEBM:
to avoid querying whole prescribing table

sebbacon:
To clarify: it costs more money (and takes longer) to query all prescribing data ever. Often, e.g. to develop new analytic approaches, it suffices to query the most recent month.

It is therefore useful to have a table of only the most recent month's data available. At the moment the team do it ad hoc; we should automate it.

inglesp:
We have a bunch of tables `prescribing_with_merged_codes_yyyy_mm`, and presumably these are manually created.  @HelenCEBM can you tell me how you create them?

HelenCEBM:
No sorry I don't know where those come from! They are probably similar to what I was asking for, though not all fields from the prescribing table are in them, like `bnf_name` ...

inglesp:
OK, thanks.  So to clarify: do we want a `hscic.prescribing_yyyy_mm` for each month we have prescribing data for, containing exactly the columns in `hscic.prescribing`?

HelenCEBM:
Yes, we used to have a copy of just the latest month for running tests on, i don't think we need to keep them forever (maybe having just the latest 2-3 months could be helpful though)

richiecroker:
Ideally we need items as well as quantity.

inglesp:
> Ideally we need items as well as quantity.

@richiecroker We've got items and quantity in [`hscic.prescribing`](https://bigquery.cloud.google.com/table/ebmdatalab:hscic.prescribing) -- are you referring to something else?

richiecroker:
No items in hscic.prescribing_yyyy__mm

inglesp:
[hscic.prescribing_2017_11](https://bigquery.cloud.google.com/table/ebmdatalab:hscic.prescribing_2017_11?tab=preview) is now up, and once #753 is deployed, we'll always add the latest month's data during an import.

@HelenCEBM, @richiecroker: Do we want data going back before 2017_11?

HelenCEBM:
Can someone clarify what is `ebmdatalab.hscic.normalised_prescribing_standard_latest_month` ? This is perfect for test queries (often easier than the named ones e.g. `_2017_12` because we don't have to manually insert the latest month's date every time); however BQ is telling me it uses 81GB to query this table, the same as the full table `ebmdatalab.hscic.normalised_prescribing_standard` which contains all the months!

richiecroker:
[...]

PR #753, opened 2018-01-30
Load latest month's prescribing data into BQ table

inglesp:
Addresses #701


inglesp:
Test failure fixed in #751.

PR #773, opened 2018-02-14
Refactor import_measures

inglesp:
I think these changes make `import_measures` easier to follow.  In particular, there's quite a bit less indirection.

There's quite a lot of churn here, and it won't be that easy to review.  (I've split things into logical commits as best I can.)


sebbacon:
The removal of in direction is welcome! But I'll have to spend some time properly reading this which won't happen til next week - sorry.

PR #812, opened 2018-03-20
Fix import_dmd_snomed

inglesp:


sebbacon:
Oh - one comment I thought of after approving - probably worth adding a quote to a row in the input CSV  fixture and changing capitalisation of one of the columns to maintain test coverage

PR #816, opened 2018-03-21
Update docker environment for local tests

rebkwok:
As it currently stands, the README instructions for running tests locally in docker with `docker-compose run test` need a couple of extra things to get them working.

-  The `SAUCE_HOST` environment variable is required (but not actually used afaict, unless the env variable `USE_SAUCELABS` is True.  It needs a value though.  I've update the docker-compose version to v2.1 so that it will accept a default value.  This also means updating minimum requirements of Compose and Docker Engine.
 
- Selenium tests don't work because they depend on firefox, geckodriver and xvfb.  
I've added install lines to `Dockerfile-test`, on the assumption that you'll want this to be updated in the image.  For testing it locally, I added those to the end of the docker-setup-sh.  Note that apt-get install of firefox on Jessie installs version 52.X which doesn’t work with later versions of geckoworker.

- `google-credentials.json` needs to be decrypted.  I added a comment to the README to that effect - but I guess these will just be expected test failures for external contributors.  Perhaps there could be a flag to pass to the test script that skips tests needing google auth credentials?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 env:
   global:
-    DOCKER_COMPOSE_VERSION: 1.7.1
+    DOCKER_COMPOSE_VERSION: 1.9.0
   matrix:
     - TEST_SUITE=nonfunctional
     - TEST_SUITE=functional BROWSER="firefox:latest:Windows 2012"
--- a/Dockerfile-test
+++ b/Dockerfile-test
@@ -1,4 +1,7 @@
 FROM ebmdatalab/openprescribing-base:latest
 
+wget -qO- https://github.com/mozilla/geckodriver/releases/download/v0.16.1/geckodriver-v0.16.1-linux64.tar.gz | tar xvz -C /usr/bin
+apt-get update && apt-get install -y firefox-esr xvfb
+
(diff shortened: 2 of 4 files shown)


sebbacon:
This all makes sense. Unfortunately it breaks the Travis build. I'm not sure why; the error's not making much sense to me.

One thing to note is that the [Travis environment's docker-compose](https://github.com/rebkwok/openprescribing/blob/f06a6d25e0395eaa815832b6aa1deacbb142456f/.travis.yml#L3) would also need updating, though I don't see how that would be related to this error.

sebbacon:
Oh, I see that odd error is appearing in your other PR which is definitely not related to anything Travis might care about. Maybe the bug was introduced elsewhere.

sebbacon:
Ah, of course, it's because our environment isn't available in your fork.  Not used to getting PRs from outside the organisation :)

rebkwok:
I've updated the travis docker-compose version now.  Sorry I didn't comment in advance on the reason for the Travis error - I saw it mentioned in another PR from an external contributor, so I'd assumed it was a known thing :)

PR #819, opened 2018-03-28
Update docs and uglifyjs requirement

rebkwok:
A few updates to the README for issues that I came across while setting up a non-docker environment.  

- pip and setuptools need to be up to date otherwise some of the requirements fail to install properly
- various JS dependencies complain about the node version (some wanting node > v8.0)
- uglifyjs is deprecated and `npm run build` fails. Updated package.json with uglify-js instead
- `npm install -g less` is also required for `npm run build` to complete successfully
- postgres 9.5 is required for the `replace_matviews` and `import_dmd` management commands that use `CREATE INDEX IF NOT EXISTS` syntax (not available before 9.5)


sebbacon:
This is great, thanks.

FYI there's [a PR](https://github.com/ebmdatalab/openprescribing/pull/689) which is starting to get very stale which refactors a lot of the javascript.

rebkwok:
Ah, yes, I saw that there were a bunch of deprecation warnings about the javascript, but I only changed the uglify-js for this PR because it was the only one that actually broke things for me

PR #820, opened 2018-03-29
Local development setup with vagrant

rebkwok:
This may not be something that you want, or that you want to keep within the main repo, but as I have set it up locally, I thought I would contribute it in case it is useful.

This allows you to set up a local development environment which is hopefully as close as possible to the production environment, using Vagrant and VirtualBox, and provisioning the box using ansible.  It may be particularly useful for anyone working on OSX.

The base box is Debian Jessie because the README specified install dependencies for Jessie.  That requires a bit of a workaround for postgres because Jessie doesn't come with 9.5.


sebbacon:
thanks @rebkwok! Have finally merged this (by hand, hence closing the PR). Hope you're well!

rebkwok:
@sebbacon Glad it  was useful 🙂

PR #826, opened 2018-04-05
Measures SQL refactors

inglesp:
Before merging, I'd like to ensure that these changes don't produce different results when used against real data.  I need to think carefully about how to do this.

### DIFF
--- a/openprescribing/frontend/management/commands/import_measures.py
+++ b/openprescribing/frontend/management/commands/import_measures.py
@@ -367,9 +367,9 @@ class MeasureCalculation(object):
     def calculate_global_centiles_for_practices(self):
         """Compute overall sums and centiles for each practice."""
         extra_fields = []
-        # Add prefixes to the select columns so we can reference the
-        # joined tables (bigquery legacy SQL flattens columns names
-        # from subqueries using table alias + underscore)
+        # Add prefixes to the select columns so we can reference the joined
+        # tables (bigquery flattens columns names from subqueries using table
+        # alias + underscore)
         for col in self._get_col_aliases('numerator'):
             extra_fields.append("num_" + col)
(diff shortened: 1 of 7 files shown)


inglesp:
I have satisfied myself that the new SQL produces the same results as before.  (Or at least: close enough.  See below.)

To do this, I created a handful of new measures by copying existing measures and appending `test_` to the name.  I did this for `ace`, `lpcoprox`, `opioidper1000`, which include both cost-based and non-cost-based measures.  (I also did it for `ktt9_uti_antibiotics` and hit a problem.  I'll open a separate issue for this.)

I also modified `import_measures.py`so that it did not write to the local database after doing some calculations in BQ.  (This involved commenting out calls to the `write_X_to_database` methods.)

For each new measure, I then ran `import_measures` for a single month (2017_01), downloaded BQ tables for both the new measure and the original measure using `download_measure_tables` and then compared them using `verify_measure_tables_match` (see attached scripts).

The tables match up to minor differences in floating point values.  Specifically, all numerical values satisfy:

    (abs(v1) < epsilon and abs(v2) < epsilon) or abs(1 - (v1 / v2)) < epsilon

for `epsilon = 1e-10`.  I'm not quite sure why the values don't match exactly, as the calculations should have been the same, but I'm not too interested in why they're not.  (And actually, re-running the calculations with the original SQL produces slightly different results each time too, but with a much smaller `epsilon = 1e-15`.

[download_measure_tables.py.txt](https://github.com/ebmdatalab/openprescribing/files/1957864/download_measure_tables.py.txt)
[verify_measure_tables_match.py.txt](https://github.com/ebmdatalab/openprescribing/files/1957865/verify_measure_tables_match.py.txt)

inglesp:
@sebbacon Can you review this please?

Issue #831, opened 2018-04-11
Produce visualisation of data flow through OpenPrescribing


inglesp:
The way that data goes from its source to a page on the website is complex, and after working on it for over a year I still don't have everything straight in my head.  A visualisation might help here.

The visualisation should cover:

* Sources of data
* Data in the database
* Data in BigQuery
* API endpoints
* Pages on the website

HelenCEBM:
Progress @inglesp ?

PR #835, opened 2018-04-20
Add beginnings of data flow visualisations

inglesp:
Addresses #831.

Lets us produce graphs like

image

which demonstrates how the org_location API endpoint is used and where it gets its data from.

Further work required to include BQ tables, source data, and management commands.

@sebbacon, feel free to review if you're interested.


inglesp:
I'm going to merge this as I think it's useful (at least for now).  I'll take responsibility for keeping it up to date.

Issue #739, opened 2018-01-24
Prepare for April CCG mergers


sebbacon:
As of the 1st of April, Aylesbury Vale CCG (`10Y`) and Chiltern CCG (`10H`) will merge to create  Buckinghamshire CCG (`14Y`).

Review what we had to do last time, write it up properly (in the Wiki), be ready for when the data starts coming through (June/July?)

richiecroker:
I think there are 18 CCGs involved in mergers in April

richiecroker:
https://wilmingtonhealthcare.com/ccgs-in-rush-to-merge/

HelenCEBM:
Incorporating STPs as mentioned above is now covered by a separate issue (#808)

inglesp:
Whenever we show historical data (on analyse page, measures pages, and for PPU) for a CCG, we want to include data for all practices that are currently in the CCG, including (if possible) closed practices that would be in the CCG if they were open.

To do this, there are a number of different, related tasks.

* Ensure that we always use a practice's current CCG membership in calculations.  This has two parts:
  * Ensure that derived tables always use current CCG membership.  These tables are:
    * Measures tables
    * `vw__` tables
    * PPUSavings table
  * Ensure that all on-the-fly calculations always use current CCG membership.  We'll need to do a full audit of all API endpoints.  One endpoint that doesn't use current CCG membership is `spending_by_practice`, used in on the analyse page.
* Work out how to assign closed practices to a CCG
* Display something sensible for closed CCGs
* Display something sensible for closed practices

inglesp:
Current status:

- [ ] Ensure that derived tables always use current CCG membership. These tables are: 
   - [ ] Measures tables
   - [ ] `vw__` tables
   - [ ] PPUSavings table
- [ ] Ensure that all on-the-fly calculations always use current CCG membership
- [ ] Work out how to assign closed practices to a CCG
- [ ] Display something sensible for closed CCGs
- [ ] Display something sensible for closed practices

sebbacon:
See #891 for discussion of closed/dormant options

PR #851, opened 2018-05-16
Spending API SQL refactors

inglesp:
This is with an eye to ensuring that API queries always use current CCG membership, as part of #739.

### DIFF
--- a/openprescribing/api/views_spending.py
+++ b/openprescribing/api/views_spending.py
@@ -467,10 +467,7 @@ def _get_query_for_total_spending(codes):
     ORDER BY date;"""
 
     if codes:
-        code_clauses = [
-            'presentation_code LIKE %s'
-            for _ in range(len(codes))
-        ]
+        code_clauses = ['presentation_code LIKE %s'] * len(codes)
     else:
         code_clauses = None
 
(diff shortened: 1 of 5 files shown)


openprescribing/api/views_spending.py
```diff
@@ -359,24 +359,23 @@ def tariff(request, format=None):
...
     spending_type = utils.get_spending_type(codes)
     if spending_type is False:
         err = CODE_LENGTH_ERROR
         return Response(err, status=400)
 
-    if not spending_type or spending_type == 'bnf-section' \
-       or spending_type == 'chemical':
-        query = _get_query_for_chemicals_or_sections_by_ccg(codes, orgs,
-                                                            spending_type)
+    if spending_type in ['product', 'presentation']:
```
sebbacon: This is a good change but might be even better if we enumerate all the possible spending types as constants somewhere, so they can all be seen in one place, and no string typos possible?

openprescribing/api/views_spending.py
```diff
@@ -385,266 +384,341 @@ def spending_by_ccg(request, format=None):
...
+    JOIN frontend_pct pc ON pr.pct_id=pc.code
+    AND pc.org_type='CCG'
+    WHERE %s
+    GROUP BY pc.code, pc.name, date
+    ORDER BY date, pc.code
+    '''
+
+    if spending_type == 'bnf-section':
+        chemical_clauses = [
+            'pr.chemical_id LIKE %s'
```
sebbacon: Took me a minute to parse this in my head. Might `['pr.chemical_id LIKE %s'] * len(codes)` be a tad clearer?

inglesp:
Thanks for comments.  Both now addressed.

PR #877, opened 2018-06-12
Allow manage.py to be called without arguments

evansd:
This is needed to support tab-completion for management commands. See:
https://github.com/django/django/blob/2bc014750adb093131f77e4c20bc17ba64b75cac/extras/django_bash_completion


inglesp:
:+1:

PR #879, opened 2018-06-12
Add command to load data for local development

evansd:
At present this just loads a bunch of test fixtures (those recommended
by @inglesp) but there's obviously scope for doing something else in
future. The main thing is that there's a single command to run which
does all this and we can update what that command actually does at our
leisure.

Thinking more broadly, it might make sense to make it a convention
across all projects that they have such a command.


inglesp:
Great!

PR #880, opened 2018-06-12
Handle invalid dates (somewhat) more gracefully

### LINKED ISSUE #832: Don't 500 on invalid dates for PPU
URLs like /ccg/07H/price_per_unit/?date=123 are being spidered, and are producing a lot of error alert noise.

These should return a 200 to the user, with an error message that the date is not valid.

evansd:
Previously, requesting certain pages with an invalid date parameter would trigger an uncaught exception. Sometimes we'd get a large number of malformed requests from bots which would trigger a large number of errors and flooding our error logs.

We now handle invalid dates more gracefully and return a 400 Bad Request instead.

Closes #832

### DIFF
--- a/openprescribing/common/utils.py
+++ b/openprescribing/common/utils.py
@@ -158,11 +158,15 @@ def constraint_and_index_reconstructor(table_name):
             logger.info("CLUSTERED %s" % table_name)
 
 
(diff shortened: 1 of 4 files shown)


inglesp:
Good stuff.

Issue #479, opened 2017-06-30
Stop old CCGs appearing as outliers on Analyse results


lmfrench:
Since April old CCGs (Newcastle, Manchester) are appearing as outliers on some searches on the Analyse page. 

![image](https://user-images.githubusercontent.com/15875191/27733952-a15c7b64-5d90-11e7-9c1a-63fb80a7fa41.png)
https://openprescribing.net/analyse/#org=CCG&numIds=0703021Q0BB&denomIds=0703021Q0&selectedTab=summary

This appears to be caused by the closed practices still in the CCG:
![image](https://user-images.githubusercontent.com/15875191/27733964-b2bce47a-5d90-11e7-84d1-1f7f66651ea5.png)
https://openprescribing.net/analyse/#org=practice&orgIds=00F&numIds=0703021Q0BB&denomIds=0703021Q0&selectedTab=chart

The above search mentions Manchester CCGs in the screenshot as outliers even though it is a search for practices in Gateshead. If you click the link above it doesn't have this message at the bottom. This only appears if you first run this search, and then without refreshing change the search to practices:Gateshead CCG.

This message wasn't appearing on analyse before the April upload (I run the Cerazette/Desogestrel search in demos). 

Other examples:
https://openprescribing.net/analyse/#org=CCG&numIds=0601060D0BHABA0,0601060D0CUABA0&denomIds=6.1.6&selectedTab=summary
https://openprescribing.net/analyse/#org=CCG&orgIds=01Y&numIds=090402000BBQLA0,090402000BBRBA0,090402000BBRAA0,090402000BBQKA0,090402000BBQMA0&denomIds=9.4.2&selectedTab=chart

sebbacon:
Relates to #408

sebbacon:
Wrapped up into #891

Issue #408, opened 2017-04-25
Add other old Newcastle CCGs to Newcastle Gateshead (13T)


lmfrench:
Newcastle CCGs merged. We currently only show Newcastle Gateshead on our dashboards (ignoring the error where Gateshead is still showing). 

A new code was not created when the CCGs merged, 13T was instead used as the code for the merged CCGs. For data prior to their merge, we only show data for the original 13T, and after the merge it is the data for all of them. It isn't clear to the user that this is what we've done. We should add the CCG data from the old CCGs to the CCG total. We also need to make a note somewhere (either on dashboard or FAQs) that this is what we've done

lmfrench:
There were also changes to other CCGs in April:

- Cumbria CCG will become North Cumbria, Lancashire North will be Morecombe Bay. Both are keeping their old codes. Some practices in these CCGs are moving. User has been in touch to tell us which practices are moving into Morecombe Bay, and this should also be reflected in the ODS Quarterly update (May 26th)
- Manchester CCGs  (00W, 01M, 01N) are merging, new CCG will be Manchester CCG (14L)

Some GP Practices are also moving CCGs, full details here: https://digital.nhs.uk/organisation-data-service/news-and-updates and should be in the May 26th update.

sebbacon:
Statement of problem:

```
┌──────┬──────────────────────────────────┬────────────┬────────────┐
│ code │               name               │ open_date  │ close_date │
├──────┼──────────────────────────────────┼────────────┼────────────┤
│ 00F  │ NHS GATESHEAD CCG                │ 2013-04-01 │ 2015-03-31 │
│ 00G  │ NHS NEWCASTLE NORTH AND EAST CCG │ 2013-04-01 │ 2015-03-31 │
│ 00H  │ NHS NEWCASTLE WEST CCG           │ 2013-04-01 │ 2015-03-31 │
│ 13T  │ NHS NEWCASTLE GATESHEAD CCG      │ 2014-11-26 │ ¤          │
└──────┴──────────────────────────────────┴────────────┴────────────┘

```

Currently:

* Measures use the current membership of a practice
* Analysis uses the membership of a practice as it was at the time

As I understand it, the solution is to change analysis to use current membership, as per Measures. To achieve this:

[...]

PR #882, opened 2018-06-13
Use current CCG membership in presentations by practice API query

inglesp:
This will fix the chart at https://openprescribing.net/analyse/#org=practice&orgIds=14L&numIds=0703021Q0BB&denom=nothing&selectedTab=chart

This has no new tests -- updating the test fixtures is too much work as it breaks lots of other tests

### DIFF
--- a/openprescribing/api/views_spending.py
+++ b/openprescribing/api/views_spending.py
@@ -648,9 +648,9 @@ def _get_presentations_by_practice(codes, org_ids, date):
     if org_ids:
         for org_id in org_ids:
             if len(org_id) == 3:
-                org_clauses.append('pr.pct_id = %s')
+                org_clauses.append('pc.ccg_id = %s')
             else:
-                org_clauses.append('pr.practice_id = %s')
+                org_clauses.append('pc.code = %s')
 
     if date:
         date_clause = 'pr.processing_date = %s'


sebbacon:
Genuine test error:

```
ERROR: test_spending_by_practice_on_multiple_presentations (frontend.tests.test_api_spending.TestSpendingByPractice)
  File "/code/openprescribing/frontend/tests/test_api_spending.py", line 530, in test_spending_by_practice_on_multiple_presentations
    rows = self._get_rows(params)
<snip>
ProgrammingError: column pc.pct_id does not exist
LINE 13: ... pr.presentation_code LIKE '0202010B0AAABAB') AND pc.pct_id ...
                                                              ^
HINT:  Perhaps you meant to reference the column "pr.pct_id".
```

I agree with the hint - although it could also be `pc.code`, but I that would *probably* result in a less performant query.


I also just spotted this in the test output, not sure how long that's been going on:

```
  File "./scripts/wait_for_postgres.py", line 16, in <module>
    host=environ['DB_HOST'],
  File "/usr/local/lib/python2.7/UserDict.py", line 40, in __getitem__
    raise KeyError(key)
KeyError: 'DB_HOST'
```

inglesp:
Gah, amateur hour!  Have fixed by changing `.pct_id` to `.ccg_id`.  (In my defence, it is a *bit* confusing...)

sebbacon:
Per my concern about performance, this query has conditions on two different tables. It is likely to be more performant (as it could use one index) to apply compound conditions to the largest table first, i.e. against `pr` rather than `pc`.

I've not given this masses of thought so might be wrong. Would at least be a good idea to try out on staging before merging, if you've not already done so, and compare performance.

sebbacon:
I am starting to review outstanding issues properly in the issue tracker, and noticed what I think describes this problem [here](https://github.com/ebmdatalab/openprescribing/issues/408#issuecomment-305133917) - worth reviewing to check you agree they're the same issue and that this PR would fix that issue

inglesp:
Have benchmarked the queries, and the new query is faster (~520ms vs ~750ms, best of three).  The `EXPLAIN` output is 231 rows and so I haven't dug into why!

You're right that [this comment](https://github.com/ebmdatalab/openprescribing/issues/408#issuecomment-305133917) and #479 are related.  They will be fixed in some cases by this PR, and in other cases by #846 (which is merged but we haven't done an import since the merge), depending on whether the analysis is for chemicals or presentations.

sebbacon:
I'm happy if you want to merge & deploy then

inglesp:
Before:

image


After:

image

Issue #891, opened 2018-06-18
Handle dormant and closed practices


sebbacon:
A dormant practice is a practice that should not be doing any prescribing (typically because it's closing, but sometimes this is temporary and it becomes active again).  However, practices that are not actively prescribing **can** (and do) appear in prescribing data -- because prescribing data is recorded at the time a script is fulfilled by a dispenser, which can happen many months after the prescription is made (see #522 for more detail on this).

When a practice is dormant, we know it (at least sometimes) [can still move CCGs](https://github.com/ebmdatalab/openprescribing/issues/513#issuecomment-317433101).  As a specific example, `P84676` was in `00W` (Central Manchester) which merged with `14L` (Manchester) in April 2017.  The data for that month showed `14L` in the prescribing data but still showed `00W` in the practices table.

In our [recently-FOIed mergers data](https://github.com/ebmdatalab/openprescribing/issues/408#issuecomment-392811309), we can also see that `P84676` merged into `P84068` in October 2016; with data being assigned to it right until April 2017.

Per #401, #215, and #479, (all?) closed and (some?) dormant practices appear in data as assigned to old CCGs.

Suggested steps: 
* [ ] `epraccur` is quarterly, so gets out of date and should only be used for CCG membership as a fallback where no other data is available. Check that we use the **most recent** CCG membership for a practice, i.e. prefer membership information given in the latest month's prescribing data. The `P84676` example above suggests we do not do this consistently, or didn't a year ago.  XXX add a new issue and link to it from here. 
* Define an **orphan practice** as a closed or dormant practice which belongs to a CCG that no longer exists, even after (a) having applied available merger information, and (b) using the current month of prescribing data for CCG membership rather than the `epraccur` data (if the previous step suggests this is the right approach)
* Report on, and discuss, the impact of excluding orphan practices:
  * [ ] Make a report of "orphan" practices by month, going back 5 years, (a) any practices assigned to a closed CCG (#527 has handy resource listing these, should also be in our PCT model), (b) if we have merge information or not for each practice, (c) its status (closed, open, dormant). Assert that there are no open orphan practices.
  * [ ] Discuss the report, bearing in mind our merger data will, at least for the time being, get out of date (currently trying to establish the best way of getting the data monthly in the FOI linked above)
* If we think it's going to be OK to exclude orphan practices from our analyses, incorporate merger data into our dataset:
  * [ ] Assign all historic prescribing to the surviving practice, per our FOI spreadsheet
[...]

PR #897, opened 2018-06-19
Flag up Closed/Dormant practices on the frontend

### LINKED ISSUE #889: Label dormant / closed practices
In the list of practices at the top of a CCG page([example](https://openprescribing.net/ccg/08C/)), and at the top of a Practice page ([example](https://openprescribing.net/practice/E85124/)), we should clearly label Closed and Dormant practices as such (with the word "Closed" or "Dormant").

In analyses, where someone is searching for organisations the practice name should also indicate "(closed)" or "(dormant)".  

Presumably a safe way of doing this is to add it to the generated title of a practice.

It may be worth linking somewhere to [our explanation of "dormant"](https://ebmdatalab.net/what-is-a-dormant-gp-practice-and-why-are-they-prescribing/) if it's possible to do discreetly.

For further background, see #215 and #522.

The current status is stored in our Practice model.

evansd:
Closes #889

### DIFF
--- a/openprescribing/api/views_measures.py
+++ b/openprescribing/api/views_measures.py
@@ -208,7 +208,10 @@ def _roll_up_measure_values(measure_values, practice_or_ccg):
         if practice_or_ccg == 'practice':
             measure_value_data.update({
                 'practice_id': measure_value.practice_id,
-                'practice_name': measure_value.practice.name,
+                'practice_name': u'{}{}'.format(
+                    measure_value.practice.name,
+                    measure_value.practice.inactive_status_suffix()
+                ),
             })
         elif practice_or_ccg == 'ccg':
             measure_value_data.update({
--- a/openprescribing/api/views_org_codes.py
+++ b/openprescribing/api/views_org_codes.py
@@ -65,14 +65,11 @@ def _get_practices_like_code(q):
             'setting': p.setting,
             'setting_name': None,
             'type': 'Practice',
-            'ccg': None,
-            'inactive_status': None
+            'ccg': None
         }
         data['setting_name'] = p.get_setting_display()
         if p.ccg:
             data['ccg'] = p.ccg.code
-        if p.has_inactive_status():
-            data['inactive_status'] = p.get_status_code_display()
         results.append(data)
     return results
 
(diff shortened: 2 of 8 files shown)


evansd:
Now deployed to staging, see e.g.
http://staging.openprescribing.net/practice/A89038/
http://staging.openprescribing.net/practice/B84614/
http://staging.openprescribing.net/practice/

In trying to test my changes to the Analyse page I've discovered that the search only returns active practices in any case so my changes here are redundant. See:
https://github.com/ebmdatalab/openprescribing/blob/24110f5a190eab8bc5d01e235fb2de812980913d/openprescribing/api/views_org_codes.py#L53

Should I back them out? Or should we change the search to return dormant practices now that we can flag them as such in the results?

sebbacon:
I think we should back them out.

How we handle closed/dormant practices is still [an ongoing discussion](https://github.com/ebmdatalab/openprescribing/issues/891), but I believe @richiecroker's preference has always been that we should merge historic data about closed practices into the surviving practice, if we can. 

If it turns out this is possible, then analyses should always be against current practices (just as we always treat their CCG membership as always having pertained to their current membership), meaning we should filter out closed/dormant practices there (but still keep them on CCG / Practice index pages).

If it's not possible, we can always revisit this PR. 

Once that's backed out, happy to merge.

sebbacon:
Some more thoughts on exploring further.

Two other places we should consider adding the status:

1. It would be nice for it to appear in the page `<title>` (thus in search engines)
2. [This measures page](http://staging.openprescribing.net/ccg/02T/nebivolol/) image

Per my previous comment, it's probable that point (2) will become irrelevant when we resolve practice mergers, but that may take some time, so it's worth doing for now, I think.

Finally, the "what does this mean" link only pertains to **dormant** practices. We probably don't need to explain what "closed" means.

evansd:
@sebbacon I've updated this PR to address the various points you raised. I can't deploy it to staging at the moment though because it's being used to asses the new measures.

evansd:
Deployed to staging now. See:
http://staging.openprescribing.net/ccg/11X/ciclosporin/#L85045 (scroll up a bit)
http://staging.openprescribing.net/practice/L85602/

PR #900, opened 2018-06-20
Respect dates specified on the command line for measure generation

### LINKED ISSUE #899: Measure importing command ignores specified start and end dates
Specifying dates is useful when generating measures on the staging server. The default is to use the most recent import date, but we may not have updated the prescribing data in the staging database for some time. 

As measures are generated from BigQuery, we can use specified dates to target recent data without having to update all the other data on the staging site.

sebbacon:
Closes #899

### DIFF
--- a/openprescribing/frontend/management/commands/import_measures.py
+++ b/openprescribing/frontend/management/commands/import_measures.py
@@ -94,8 +94,11 @@ class Command(BaseCommand):
     def parse_options(self, options):
         """Parse command line options
         """
-        if len([x for x in (options['start_date'], options['end_date']) if x is not None]) == 1:
-            raise CommandError('--start_date and --end_date must be given together')
+        if len([
+                x for x in (options['start_date'], options['end_date'])
+                if x is not None]) == 1:
+            raise CommandError(
+                '--start_date and --end_date must be given together')
         if 'measure' in options and options['measure']:
             if "," in options['measure']:
                 options['measure_ids'] = options['measure'].split(',')
@@ -111,14 +114,16 @@ class Command(BaseCommand):
                  and options['month_from_prescribing_filename']:
                 filename = options['month_from_prescribing_filename']
                 date_part = re.findall(r'/(\d{4}_\d{2})/', filename)[0]
-                month = datetime.datetime.strptime(date_part + "_01", "%Y_%m_%d")
+                month = datetime.datetime.strptime(
+                    date_part + "_01", "%Y_%m_%d")
 
                 options['start_date'] = options['end_date'] = \
                     month.strftime('%Y-%m-01')
             else:
                 l = ImportLog.objects.latest_in_category('prescribing')
                 options['start_date'] = "%s-%02d-%02d" % (
-                    l.current_at.year - 5, l.current_at.month, l.current_at.day)
+                    l.current_at.year - 5,
+                    l.current_at.month, l.current_at.day)
                 options['end_date'] = l.current_at.strftime('%Y-%m-%d')
         # validate the date format
         datetime.datetime.strptime(options['start_date'], "%Y-%m-%d")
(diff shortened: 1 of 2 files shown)


openprescribing/frontend/management/commands/import_measures.py
```diff
@@ -94,6 +94,11 @@ def add_arguments(self, parser):
     def parse_options(self, options):
         """Parse command line options
         """
+        if len([
+                x for x in (options['start_date'], options['end_date'])
+                if x is not None]) == 1:
+            raise CommandError(
+                '--start_date and --end_date must be given together')
```
inglesp: What about:

```
if bool(options['start_date']) != bool(options['end_date']):
    raise CommandError(...)
```

openprescribing/frontend/management/commands/import_measures.py
```diff
@@ -102,22 +107,24 @@ def parse_options(self, options):
...
-
-            options['start_date'] = options['end_date'] = \
-                month.strftime('%Y-%m-01')
-        else:
-            l = ImportLog.objects.latest_in_category('prescribing')
-            options['start_date'] = "%s-%02d-%02d" % (
-                l.current_at.year - 5, l.current_at.month, l.current_at.day)
-            options['end_date'] = l.current_at.strftime('%Y-%m-%d')
+        if not options['start_date']:
+            if 'month' in options and options['month']:
```
inglesp: When would `'month' in options` be true and `options['month']` not be?

openprescribing/frontend/management/commands/import_measures.py
```diff
@@ -102,22 +107,24 @@ def parse_options(self, options):
...
+                options['start_date'] = options['end_date'] = options['month']
+            elif 'month_from_prescribing_filename' in options \
+                 and options['month_from_prescribing_filename']:
```
inglesp: I can't see us ever using `month_from_prescribing_filename` -- can we drop it?

openprescribing/frontend/tests/commands/test_import_measures.py
```diff
@@ -38,22 +39,45 @@ def test_measures():
     }
 
 
+def parse_args(*opts_args):
+        parser = argparse.ArgumentParser()
+        cmd = Command()
+        parser = cmd.create_parser("import_measures", "")
+        options = parser.parse_args(opts_args)
+        return cmd.parse_options(options.__dict__)
```
inglesp: This function is double-indented.

Also, why is the `.__dict__` necessary in `cmd.parse_options(options.__dict__)`?

openprescribing/frontend/management/commands/import_measures.py
inglesp: Or even, for more clarity:

```
if options['start_date']:
    if not options['end_date']:
        raise ...

if options['end_date']:
    if not options['start_date']:
        raise ...
```

openprescribing/frontend/management/commands/import_measures.py
sebbacon: I wondered about this but left it, because I wondered if we'd need it to restore from old files.

But you're probably right, on reflection

openprescribing/frontend/management/commands/import_measures.py
sebbacon: I prefer your first suggestion as it's clearer than what I wrote but still doesn't involve repetition. The CommandError itself clarifies the intention sufficiently, IMO

openprescribing/frontend/management/commands/import_measures.py
sebbacon: IIRC to test, we manually run through argparse, which behaves slightly differently from Django, and makes it possible for `options['month']` to be `None`. But I can't reproduce now!

openprescribing/frontend/tests/commands/test_import_measures.py
sebbacon: To the latter question: because that's what `call_command` does in django, and there's no subcommand we can call to replicate that.  Will add a comment.

PR #901, opened 2018-06-20
Upgrade jQuery

sebbacon:
Addresses XSS vulnerability CVE-2015-9251

### DIFF
--- a/openprescribing/media/js/package.json
+++ b/openprescribing/media/js/package.json
@@ -59,7 +59,7 @@
     "factor-bundle": "^2.5.0",
     "handlebars": "4.0.5",
     "humanize": "^0.0.9",
-    "jquery": "^1.11.3",
+    "jquery": "^3.0.0",
     "mapbox.js": "^2.2.1",
     "moment": "^2.10.3",
     "npm-run-all": "^4.1.2",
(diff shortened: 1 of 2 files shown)


sebbacon:
This PR needs some reasonably exhaustive manual testing on staging before deploying.

sebbacon:
Errors discovered when using jquery-migrate with jquery 3:
* On analyse form, event bindings `.click(function() {})`, `.scroll(function() {})` are deprecated and should be `.on("click", function() {})` etc. `.click()` should be `.trigger("click")`.
* The [tariff viewer](http://staging.openprescribing.net/tariff/) also has `jQuery.fn.offset() requires an element connected to a document`
* A bunch of errors that I think derive in Highcharts (hard to say with our `common.js` bundling), including `jQuery.fn.bind() is deprecated`, `jQuery.fn.removeAttr no longer sets boolean properties: disabled`, visible [here](http://staging.openprescribing.net/analyse/#org=CCG&numIds=0212000AA&denomIds=2.12&selectedTab=summary) for example
* [This map]( http://staging.openprescribing.net/analyse/#org=CCG&numIds=0212000AA&denomIds=2.12&selectedTab=chart) produces error `Invalid LatLng object: (52.939894110413846, NaN)`

PR #908, opened 2018-06-28
Delete old measure values

### LINKED ISSUE #907: Measures are only recalculated for last five years
When importing measures each month, we calculate (or recalculate) measure values for the last five years.

This causes two problems:

1. For new CCGs (eg [Bucks](https://openprescribing.net/ccg/14Y/)) we can only show five years of data, but have percentiles going back to May 2012.

image

2. For CCGs that have been around more than five years, any measures older than five years do not reflect the current composition of practices in the CCG, but the composition five years later.

Some possible solutions, with quick thoughts about what each would involve:

1. Only ever show five years of measures.  This would be easiest.
2. Recompute all months' measures each month.  We currently only keep five years of prescribing data (presumably to save space), so this would involve (a) keeping prescribing data indefinitely; (b) reinstating prescribing data from more than five years ago.  We _should_ have backups of all the source prescribing data, but it may be fiddly to reimport.
3. Reconstruct CCG measure values from practice measure values.  We have practice measure values going back to May 2012, and as the CCG measure values aggregate these, it _may_ be possible to reconstruct the CCG measure values from the practice measure values.  However we wouldn't be able to cope with cases where the measure definition changes.

This lazy programmer strongly favours 1, but there may be a good business case for exploring 2 or 3, and @sebbacon may have a brilliant idea for 4.

inglesp:
Fixes #907

### DIFF
--- a/openprescribing/frontend/tests/commands/test_import_measures.py
+++ b/openprescribing/frontend/tests/commands/test_import_measures.py
@@ -332,7 +332,7 @@ class BigqueryFunctionalTests(TestCase):
         }
         self._assertExpectedMeasureValue(measure, month, expected)
 
-    def test_import_measureglobal(self):  # failing
+    def test_import_measureglobal(self):
         month = '2015-09-01'
         measure = Measure.objects.get(id='cerazette')
         expected = {
@@ -420,7 +420,6 @@ class BigqueryFunctionalTests(TestCase):
                                 name='BASSETLAW DRUG & ALCOHOL SERVICE',
                                 setting=1)
 
-        args = []
         measure = Measure.objects.create(
             id='cerazette',
             name='Cerazette vs. Desogestrel',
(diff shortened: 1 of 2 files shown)


openprescribing/frontend/tests/commands/test_import_measures.py
```diff
@@ -109,10 +110,120 @@ def test_reconstructor_called_when_no_measures_specified(self, conn):
 class BigqueryFunctionalTests(TestCase):
     @classmethod
     def setUpTestData(cls):
-        cls.env = patch.dict(
```
sebbacon: ISTR this was necessary because we were calling a management command which didn't know it should be working on the test database; perhaps via a shell which we're no longer doing?  Anyway, if it all seems to work it's no longer necessary - just wanted to mention it.

inglesp: Pretty sure that by the time we try patching the environment, the settings have already been loaded and `DATABASES` will already have been set.

sebbacon: Not if subsequent steps call via a subshell? Anyway, no matter. Either way I clearly remember this being necessary at some point, and taking be ages to get right, so just wanted to wave a flag to make it ring a bell if necessary in the future.

inglesp: > Not if subsequent steps call via a subshell?

The patching here doesn't change the environment -- it just changes what `os.environ` returns.

PR #914, opened 2018-07-02
Update lptramadolpara.json

brianmackenna:
Added in extra products


sebbacon:
Question: should we be updating the description every time the codes change?

inglesp:
These are on staging at http://staging.openprescribing.net/measure/lptramadolpara/ and http://staging.openprescribing.net/measure/lprubefacients/.

sebbacon:
Bundled into #1081

Issue #849, opened 2018-05-11
Allow linking to / grouping of related measures


HelenCEBM:
For some measures such as overall antibiotic rate and % broad spectrum, it is helpful to view/interpret the two measures in combination; similarly for opioids measures (currently x3).

Could create some grouping or linkage for each of these measures, e.g.
- "see also X" which links directly to the other measure(s). E.g. if on a practice/ccg dashboard, `see also opioid prescribing rate per 1000` could link to [practice_id]/#opioidper1000 to highlight the related measure.
- "see all related measures" linking to a page displaying the group of related measures. However, note that for antibiotics there is a wider group of measures on the same topic, some less closely related than others, so multiple options may be needed.

sebbacon:
See also #882

sebbacon:
#876 I mean

sebbacon:
I think there are also some other issues relating to this that I'd like to gather together before acting.

Agreed @evansd will review existing code first, then we'll return for more discussion as there are a few high-level UI options here.

sebbacon:
Related issues:

* #810 is related insofar as it's a measure we didn't want to appear with other measures, so we tagged it `paper`, and then we link to it explicitly in papers like https://openprescribing.net/ccg/09C/?tags=paper
* #847 is the second instance of us wanting to do this, i.e. a measure that we want to maintain as a living dataset linked to from a paper
* #876 is a bug in the current implementation, which is in any case quite hard-coded and needs further generalisation

As an example of how we might generalise this - there could be a tabbed interface to measures showing "Core measures", "Cost saving measures", "Efficacy measures", "Safety measures", "Measures mentioned in papers".  This would only address Helen's second bullet point in the OP.

evansd:
Also relevant is Helen's categorisation of measures here:
https://docs.google.com/spreadsheets/d/1Yx9_dWnjscN6FNfes5Q3LT2E2yRdYbQVOY1Q-aZRFQw/edit?usp=sharing

evansd:
## Proposed solution (for discussion)

### 1. Switch from JSON to YAML for measure definitions?

YAML has the advantage that it's easier to write for humans e.g. it's not fussy about trailing commas, it supports mutliline strings and comments. Embedding chunks of text (and even HTML) is [very easy](https://en.wikipedia.org/wiki/YAML#Indented_delimiting).

[...]

PR #915, opened 2018-07-02
Improve tag behaviour to support grouping/linking measures

evansd:
This addresses most of what we need, code-wise, to make it easy to navigate groups of measures on the site (see #849). However there are only two tags currently in use (`core` and `lowpriority`) so this feature will not be particularly useful until we define some more.

At the point when we have a reasonably rich set of tags on our measures we can think about adding a tabbed interface to expose the most useful of these.


openprescribing/frontend/tests/test_views.py
```diff
@@ -394,7 +395,7 @@ def test_all_measures(self):
         self.assertContains(response, 'Cerazette')
 
     def test_all_measures_with_tag_filter(self):
-        response = self.client.get('/measure/?tags=frob')
+        response = self.client.get('/measure/?tags=lowpriority')
```
inglesp: Worth adding a test for `tags=core` and checking that Cerazette _is_ returned?

openprescribing/frontend/tests/test_views.py
```diff
@@ -453,3 +454,18 @@ def test_price_per_unit_histogram_with_practice(self):
...
+        with self.assertRaises(BadRequestError):
+            _get_measure_tag_filter({'tags': 'nosuchtag'})
+
+    def test_filters_on_core_tag_by_default(self):
+        tag_filter = _get_measure_tag_filter({})
+        self.assertEqual(tag_filter['tags'], ['core'])
+
+    def test_returns_tag_name(self):
+        tag_filter = _get_measure_tag_filter({'tags': 'lowpriority'})
+        self.assertEqual(tag_filter['name'], 'Low Priority')
```
inglesp: Could you add tests for the `show_all_by_default` flag?

inglesp: And the `show_message` key in the return value?

evansd: Yes, of course. Sorry this PR is horribly light on tests at the moment. I thought it would be better to get eyes on it and feedback before I spent too much time writing tests. (I'm am very much not a TDD person I'm afraid -- much as I would like would to be :)

sebbacon:
I suggest having a conversation right nowish with the team about other tags as we could profitably define a few straight away

evansd:
This is up on staging now if anyone wants to have a look:
http://staging.openprescribing.net/ccg/02N/?tags=lowpriority

sebbacon:
With respect to `core` tag, it's meaning is twofold: "include in alerts" and "show in main list". I wonder if we should take the opportunity to refactor to make this clearer? Just a thought.

PR #919, opened 2018-07-03
Attempt to integrate Coveralls for test coverage

### LINKED ISSUE #685: Set up coveralls
Use https://coveralls.io/ to generate reports on builds.

Consider making a reduction in coverage a failure.

evansd:
Closes #685


evansd:
@sebbacon I've finally got this working (see https://coveralls.io/github/ebmdatalab/openprescribing) but I don't have admin access on this repo to set up the integration in the other direction (i.e. for coveralls to update the status of the pull request).

evansd:
I think I also need more access on coveralls.io itself. It can't find the associated source files because it doesn't have the right "repo_root" set, but it doesn't give me an option to change this and I think this is a permissions thing.

sebbacon:
I think Coveralls.io is implemented as an [OAuth service rather than a Github App](https://developer.github.com/apps/differences-between-apps/#about-github-apps/). I can't find any permissions settings on Coveralls.io, so I guess this is related to the permissions of the access token?

So - I've given you admin rights on openprescribing. That's the only permissions setting I can see anywhere that *might* be pertinent. Does it help?!

evansd:
:+1: That's done the trick. In theory coveralls should now update the status of PRs with no minimum coverage threshold, but with any decrease counting as a failure.

I might wait until Monday when I'm next officially working before merging this though, in case there are unforeseen breakages.

PR #929, opened 2018-07-09
Update and rename lpglutenfree.json to glutenfree.json

richiecroker:
Move from LMV to core, as no longer part of NHSE consultation.

### DIFF
--- a/openprescribing/frontend/management/commands/measure_definitions/glutenfree.json
+++ b/openprescribing/frontend/management/commands/measure_definitions/glutenfree.json
@@ -1,5 +1,5 @@
 {
-  "name": "NHS England Low Priority Treatment - gluten free products",
+  "name": "Prescribing of gluten free products",
   "title": [
     "Gluten-free products: cost per 1000 patients"
   ],
@@ -17,8 +17,8 @@
   "is_cost_based": false,
   "low_is_good": true,
   "tags": [
-    "cost",
-    "lowpriority"
+    "core",
+    "cost"
   ],
   "numerator_columns": [
     "SUM(actual_cost) AS numerator, "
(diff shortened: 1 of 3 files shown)


inglesp:
This should either rename (or most likely remove) `lpglutenfree` from [practice_data_all_low_priority.sql
](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/frontend/management/commands/measure_sql/practice_data_all_low_priority.sql#L45-L52).

richiecroker:
yes, remove.

evansd:
@inglesp / @sebbacon Just want to double check the procedure for deploying this change. One approach would be:

 * Import the `glutenfree` measure as if it was a brand new measure
 * Manually delete the old `lpglutenfree` measure and its associated data

Alternatively, I could modify all references to `lpglutenfree` to be `glutenfree` and then just import the updated measure definition.

Presumably though, something needs to happen to recalculate the overall Low Priority scores. Do I need to do anything special to make this happen?

sebbacon:
Sounds fine.

For the omnibus measure, you'll want to edit the `practice_data_all_low_priority` view, and recalculating the `lpzomnibus` measure.

inglesp:
> I could modify all references to lpglutenfree to be glutenfree and then just import the updated measure definition.

Presumably you mean something like `MeasureValue.objects.filter(measure_id='lpglutenfree').update(measure_id='glutenfree')` (and also for `MeasureGlobal`)?

That should work, although you'd need to import the new `Measure` first otherwise you'll get an `IntegrityError`.

> Presumably though, something needs to happen to recalculate the overall Low Priority scores. Do I need to do anything special to make this happen?

Yes, you'll need to run `import_measures` on the `lpzomnibus` measure.  But before that, you'll need to update the [measure definition](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/frontend/management/commands/measure_definitions/lpzomnibus.json) and change `18` to `17`, since there are now 17 low-priority measures.

Does this make sense?

evansd:
Perfect, that all makes sense. Thanks @inglesp

PR #930, opened 2018-07-09
Remove August 2013 limit for CCG analyse queries

### LINKED ISSUE #920: Analyse queries highlighting CCGs cut off at August 2013
Analyse queries that highlight a CCG or CCGs ([example](https://openprescribing.net/analyse/#org=CCG&orgIds=99P&numIds=0703021Q0BB&denom=nothing&selectedTab=chart)) only show data going back to August 2013, while all other queries ([example](https://openprescribing.net/analyse/#org=practice&orgIds=99P&numIds=0703021Q0BB&denom=nothing&selectedTab=chart)) show data going back five years.

This is deliberate although possibly mistaken.  [This comment](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/media/js/src/chart_utils.js#L273-L276) says that CCGs were formed in August 2013, but my understanding from [here](https://www.nhscc.org/ccgs/) is that they were formed in April 2013.

@richiecroker Can you think of any reason not to show data for CCGs from before August 2013?

@sebbacon You added the check for August 2013 in f8690afb.  When we discussed this a couple of months ago, you thought this might be a mistake.  Can you confirm?

inglesp:
The limit should have been April 2013 (as this is when CCGs were formed)
but April and August are easy to mix up!

We can remove the limit completeyl now as we only hold five years of
data, and we have now imported data from April 2018.

Fixes #920.

### DIFF
--- a/openprescribing/media/js/src/chart_utils.js
+++ b/openprescribing/media/js/src/chart_utils.js
@@ -269,13 +269,6 @@ var utils = {
     var monthRange = [];
     if (combinedData.length > 0) {
       var firstMonth = combinedData[0].date;
-      if (options.org === 'CCG') {
-        // CCGs were formed in Aug 2013. This special-casing can be
-        // removed after Aug 2018, as we only deal with 5 years of
-        // data.
-        var firstCCGDate = '2013-08-01';
-        firstMonth = firstMonth > firstCCGDate ? firstMonth : firstCCGDate;
-      }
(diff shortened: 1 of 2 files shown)


inglesp:
Wait for @richiecroker to respond to query on #920 before merging.

richiecroker:
Nothing special about August 2013!  April 2013 on the other hand was the first month of CCGS.

Issue #924, opened 2018-07-05
Missing BNF codes in DT download


richiecroker:
There are still some missing codes in the DT download, most obvious is Vit B co tabs.

![image](https://user-images.githubusercontent.com/18683239/42313263-358897a6-803a-11e8-9410-494eac1cbcaf.png)

Is this related to #818?

inglesp:
Have just re-run BNF/SNOMED mapping import, and now all but 4 of the records in the CSV file have a BNF code.

The exceptions are all for VMPP "Ibuprofen 200mg / Phenylephrine 6.1mg tablets 16 tablet".  This VMPP (34874111000001109) isn't referenced in the mapping file (although its AMPP (34875011000001107)) is.

sebbacon:
I've written to BSA about the Ibuprofen example.

sebbacon:
Answer: it's a newer product than the mapping.

> In relation to the product Ibuprofen 200mg / Phenylephrine 6.1mg tablets 16 tablet, this item was only added to our CDR database in April this year and the June mapping file relates to data that was available in CDR in March 2018. The next mapping file due to be published in September will be based on data that was available in CDR for June 2018 and as such this product will be included in this dataset.

inglesp:
And if it's a new product in April we won't see it in the prescribing data until July, so it should only be without a BNF code for a couple of months.

sebbacon:
See #1029 for a systematic analysis of missing codes

Issue #818, opened 2018-03-28
Problem with latest BNF/SNOMED mapping import


richiecroker:
For certain drugs (possibly where there has been a change in Category) the charts do not render:

Selecting:

![image](https://user-images.githubusercontent.com/18683239/38024585-661a9974-327d-11e8-810a-1a4e0c44f828.png)

gives:

![image](https://user-images.githubusercontent.com/18683239/38024591-6d7998fa-327d-11e8-9387-9199bcfd2b2c.png)
 (and the wrong URL for the code at the top)

Entering the correct code manually into the URL also doesn't work:

![image](https://user-images.githubusercontent.com/18683239/38024630-82c60518-327d-11e8-9116-0cbdf9461e9f.png)

inglesp:
Almost certainly caused by problems with the SNOMED mapping import.

There are a lot of `Product`s with the BNF code in question:

```
# select count(*) from dmd_product where bnf_code = '190201000AABLBL';
┌───────┐
│ count │
├───────┤
│  1424 │
└───────┘
(1 row)

Time: 30.053 ms
```

---

The problem's not with how we imported the data, but with the data itself: 2142 rows in the March spreadsheet have BNF code `190201000AABLBL`.  They also all have `Unspec Drug Code_` as in the column`MDR: Product Description`.  (And they're the only rows with that value in that column.)

Compare this with the January data, where only two rows have that BNF code:

```
190201000AABLBL,Unspec Drug Code_,321894004,Amoxapine 100mg tablets,
190201000AABLBL,Unspec Drug Code_,,Exception Handler Unspecified Item,
```

Note the curious value of the last field (this is `DMD: Product Description`): `Exception Handler Unspecified Item`.

---

It's only `190201000AABLBL` that has a massive number of rows in the March spreadsheet, but there are over 100 BNF codes that appear in more than 100 rows (compared with 38 BNF codes that appear in more than 100 rows in the January spreadsheet).

@richiecroker, how did you find the problem with `190201000AABLBL`?

---

I'm going to reimport the January data, and we should follow up with NHS BSA to try to get better data for March.

inglesp:
Part of my confusion last week was because my spreadsheet program (Numbers) struggles with large spreadsheets, and can only export ~65k lines of a spreadsheet to a CSV file, while the January and March spreadsheets contain 1,048,575 and 142,521 records respectively.  I've since exported the spreadsheets to CSV using Python, which has made analysis easier.

[...]

PR #943, opened 2018-07-23
Update import_dmd_snomed to handle June spreadsheet

inglesp:
Refs #818.


inglesp:
I forgot we even had tests for this!

inglesp:
I've run the version at 115b56c09b6a12ae9b4d5b5529dd81a653c71a9f against the production data, and this has fixed #924.

I have not yet updated the tests, as the test data is not self-consistent (eg there's an AMPP with no corresponding VMPP) and I'd like to take this opportunity to make it more so.  But I've got a bit bogged down trying to work out whether there's a relationship between a VMP having zero/one/many BNF codes, and whether corresponding AMPs have zero/one/many BNF codes.

PR #944, opened 2018-07-23
Support downloading measure data

### LINKED ISSUE #917: Allow download of measures data as CSV
This is a common reason for people requesting a custom extract; a lot of measures are fiddly or impossible to recreate in the analyse page.

evansd:
Closes #917


evansd:
Some particular points for review:
1. Are we happy with the [columns](https://github.com/ebmdatalab/openprescribing/blob/251b6f9e2d16511c4bb0617d8a6ed2669eee436b/openprescribing/media/js/src/measure_utils.js#L521-L523) I've chosen? Is this the data people actually want?
2. Are we happy with the [Google Analytics](https://github.com/ebmdatalab/openprescribing/blob/251b6f9e2d16511c4bb0617d8a6ed2669eee436b/openprescribing/media/js/src/measures.js#L197-L202) tracking?

openprescribing/media/js/src/measure_utils.js
```diff
@@ -499,6 +500,79 @@ var utils = {
...
+    return table.map(this.formatRowAsCSV.bind(this)).join('\n');
+  },
+
+  formatRowAsCSV: function(row) {
+    return row.map(this.formatCellAsCSV.bind(this)).join(',');
+  },
+
+  formatCellAsCSV: function(cell) {
+    cell = cell ? cell.toString() : '';
+    if (cell.match(/[,"\r\n]/)) {
```
sebbacon: Is this actually required? I guess it's useful future proofing even if not.

evansd: No, it's not required at the moment. The most likely use case would be if we start putting something more descriptive in the column headings. But in any case, my OCD won't let me write a function called "format as CSV" if it only formats as CSV by accident :)

PR #968, opened 2018-08-02
Add client for Sentry error tracking

evansd:


requirements/base.txt
```diff
@@ -38,6 +39,7 @@ googleapis-common-protos==1.5.3  # via google-api-core
 html2text==2016.9.19
 httplib2==0.10.3          # via google-api-python-client, oauth2client
 idna==2.6                 # via cryptography, requests
+ipaddress==1.0.22         # via cryptography
```
inglesp: Any idea why this line has appeared now?

evansd: I don't. It's slightly mysterious, but looks fairly harmless.

PR #983, opened 2018-08-11
WIP: Upgrade to Python3

inglesp:
I've run `2to3` over the codebase and made some changes to get the majority of tests to pass.


inglesp:
The work so far has taken about three hours of developer time.  My current sticking point is getting Travis to run the tests with Python 3.

The following steps (at least) are still required:

- [ ] Get tests running on Travis
- [ ] Investigate `python3-openid 3.1.0 has requirement defusedxml<=0.4.1, but you'll have defusedxml 0.5.0 which is incompatible.` in Travis output
- [ ] Fix remaining few tests
- [ ] Review untested code, and check that files are opened with correct permissions
- [ ] Install Python 3 (alongside Python 2) on server (with `pyenv`?)
- [ ] Deploy this branch to staging
  - [ ] Kick tyres of website
  - [ ] Run end-to-end pipeline tests
- [ ] Merge into master and deploy to production
- [ ] Feel warm and fuzzy

sebbacon:
I expect Travis isn't using python 3 because the tests use their own docker image which has to be rebuilt to include the new python3 base image...?

Search for "Any time you change the npm or pip dependencies, you should rebuild the docker image used by the tests" in the README.

sebbacon:
So the ansible tests are failing because they depend on the Debian-provided `python`, rather than `python3`.  Installed around here:

https://github.com/ebmdatalab/openprescribing/blob/539ed950b2ba9766d2788f659f2cfe26fff466b3/ansible/roles/setup/tasks/main.yml#L40

Virtualenv set up here:

https://github.com/ebmdatalab/openprescribing/blob/539ed950b2ba9766d2788f659f2cfe26fff466b3/ansible/roles/app/tasks/main.yml#L2-L9

As we're using `jessie` the newest Python is 3.4.2. We could either use that, or change the ansible scripts to use pyenv instead (e.g. [this](https://github.com/dylanninin/ansible-pyenv)).

3.4.2 is rather old, but we get security fixes for free.

inglesp:
I'm going to close this as it's going to get harder and harder to keep in sync with changes to master.

I would like to revisit this sometime in the new year.

PR #994, opened 2018-08-15
Pick worse measure in the same way the listing page does

### LINKED ISSUE #989: Dashboards: Measures  selected as having the most potential for improvement are inconsistent with ranked list
may be because the measure is selected based on only one month of data (unlike the ranked list which takes into account 6 months of data). E.g. this practice gets opioids highlighted (97th percentile last month) but at the top of the ranked list is glaucoma (96th percentile last month) https://openprescribing.net/practice/E81030/measures/
- But sometimes they are measures with no data, e.g. https://openprescribing.net/practice/F85024/
image
- This also applies to the CCG dashboards.

### LINKED ISSUE #991: Dashboards: problem when the worst measure is the low-priority omnibus measure
image

(this issues is broken out from #988)

sebbacon:
Fixes #989

Also refactor this logic so it's identical for CCGs and
Practices, and in doing so, fix #991

Needs tests.

### DIFF
--- a/openprescribing/frontend/views/views.py
+++ b/openprescribing/frontend/views/views.py
@@ -378,7 +378,7 @@ def _home_page_context_for_entity(request, entity):
     entity_type = type(entity).lower()
     mv_filter["{}_id".format(entity_type)] = entity.code
     # find the core measurevalue that is most outlierish
-    extreme_measurevalue = MeasureValue.objects.\filter(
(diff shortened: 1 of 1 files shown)


sebbacon:
Good suggestion, but I'm going to merge on the grounds no-one's around much at the moment, lots to do, and it's broken on live...

PR #1001, opened 2018-08-22
WIP: Check in gunicorn config

inglesp:
This PR contains an updated version of `gunicorn_start` that can be checked in.

It is agnostic about the environment in which it runs, which is now passed in as an argument.

Before merging and deploying, I would like to talk through this at the next sysadmin session.

Deployment checklist:

- [x] Back up `/etc/profile.d/environment.sh`!
- [x] Review `/etc/profile.d/environment.sh` and remove any variables which are unused

For each of staging and production:

- [ ] Back up `gunicorn_start`
- [ ] Deploy
- [ ] Move contents of `/etc/profile.d/environment.sh` to `environment` files
- [ ] Update `environment` with the variables listed under "Only required on the server" in `environment-sample`
- [ ] Remove `source /etc/profile.d/openprescribing.sh` from `.venv/bin/activate`
- [x] Update `/etc/supervisor/conf.d/openprescribing.conf ` and `/etc/supervisor/conf.d/openprescribing_staging.conf` to pass environment (ie `staging` or `production`) to `gunicorn_start`

### DIFF
--- a/bin/gunicorn_start
+++ b/bin/gunicorn_start
@@ -1,0 +1,41 @@
+#!/bin/bash
+ 
+NAME="openprescribing_prod"                                  # Name of the application
+DJANGODIR=/webapps/openprescribing/openprescribing             # Django project directory
+SOCKFILE=/webapps/openprescribing/run/gunicorn.sock  # we will communicte using this unix socket
+USER=hello                                        # the user to run as
+GROUP=www-data                                    # the group to run as
+NUM_WORKERS=6                                     # how many worker processes should Gunicorn spawn
+DJANGO_SETTINGS_MODULE=openprescribing.settings.production             # which settings file should Django use
+DJANGO_WSGI_MODULE=openprescribing.wsgi                     # WSGI module name
+LOGFILE=/webapps/openprescribing/logs/gunicorn-error.log
+ACCESS_LOGFILE=/webapps/openprescribing/logs/gunicorn-access.log
+TIMEOUT=60
+ 
+echo "Starting $NAME as `whoami`"
+echo "Env variable $GMAIL_PASS"
+echo "DB_USER $DB_USER"
+echo "Test env variable $TEST1" 
+# Activate the virtual environment
+cd $DJANGODIR
+source ../.venv/bin/activate
+export DJANGO_SETTINGS_MODULE=$DJANGO_SETTINGS_MODULE
+export PYTHONPATH=$DJANGODIR:$PYTHONPATH
+export NEW_RELIC_CONFIG_FILE=/webapps/openprescribing/newrelic.ini 
+export NEW_RELIC_ENVIRONMENT=production
+export NEW_RELIC_LICENSE_KEY=ZZZZZZ
+# Create the run directory if it doesn't exist
+RUNDIR=$(dirname $SOCKFILE)
+test -d $RUNDIR || mkdir -p $RUNDIR
+ 
+# Start your Django Unicorn
+# Programs meant to be run under supervisor should not daemonize themselves (do not use --daemon)
+exec newrelic-admin run-program gunicorn ${DJANGO_WSGI_MODULE}:application \
+  --name $NAME \
+  --workers $NUM_WORKERS \
+  --timeout $TIMEOUT \
+  --user=$USER --group=$GROUP \
+  --bind=unix:$SOCKFILE \
+  --log-level=warn \
+  --log-file=$LOGFILE \
+  --access-logfile=$ACCESS_LOGFILE
--- a/bin/gunicorn_start
+++ b/bin/gunicorn_start
@@ -33,6 +33,9 @@ echo "DB_USER: $DB_USER"
 echo "GUNICORN_USER: $GUNICORN_USER"
 echo "PYTHONPATH: $PYTHONPATH"
 
+# Source environment file
+source "$INSTALL_ROOT/environment"
+
 # Activate the virtual environment
 source "$INSTALL_ROOT/.venv/bin/activate"
  
--- a/bin/gunicorn_start
+++ b/bin/gunicorn_start
@@ -1,7 +1,14 @@
 #!/bin/bash
 
-ENVIRONMENT=$1
-NAME="openprescribing_{$ENVIRONMENT}"
+ENV=$1
+
+if [[ $ENV != "production" && $ENV != "staging" ]]
+then
+  echo '$ENV must be "production" or "staging"'
+  exit 1
+fi
+
+NAME="openprescribing_{$ENV}"
 
 # INSTALL_ROOT and SITE_ROOT match definitions in settings
 INSTALL_ROOT=$(dirname $(dirname $0))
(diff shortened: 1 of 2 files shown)


bin/gunicorn_start
```diff
@@ -0,0 +1,52 @@
...
+LOGDIR="$INSTALL_ROOT/logs"
+
+# Create the run and logs directories if they don't exist
+mkdir -p $RUNDIR
+mkdir -p $LOGDIR
+
+SOCKFILE="$RUNDIR/gunicorn.sock"
+LOGFILE="$LOGDIR/gunicorn-error.log"
+ACCESS_LOGFILE="$LOGDIR/gunicorn-access.log"
+
```
sebbacon: Maybe assert environment is one of the allowed values here?  e.g.

    [[ $ENV == "production" || $ENV == "staging" ]] || (echo '$ENV must be "production" or "staging"' && exit 1)

environment-sample
```diff
@@ -23,3 +23,12 @@ SLACK_GENERAL_POST_KEY=slack_general_post_key
 CF_API_KEY=cf_api_key
 CF_API_EMAIL=cf_api_email
 NEWRELIC_API_KEY=newrelic_api_key
+
```
sebbacon: It feels to me that these variables belong elsewhere:

(a) they're not to do with the application's environment (hence the "only required on the server")
(b) they're not about secrets (which this file is supposed to be about, really)

Feels to me like they might as well actually be literals in the gunicorn script...?

bin/gunicorn_start
inglesp: Good idea, fixed.  (Although it turns out that `(` and `)` start a subshell, and the `exit 1` exits from the subshell, so I've rewritten slightly.)

environment-sample
inglesp: For `GUNICORN_TIMEOUT` and `GUNICORN_LOG_LEVEL`, there are different values in staging and production.  And for these, and `GUNICORN_NUM_WORKERS`, we might want to change the values without committing and deploying.

sebbacon: Ah OK!

inglesp:
On staging we got as far as deploying, updating the environment file, and updating `openprescribing_staging.conf`.  But after updating `openprescribing_staging.conf` we ran `sudo /etc/init.d/supervisor restart`, which failed:

```
$ sudo /etc/init.d/supervisor restart
[....] Restarting supervisor (via systemctl): supervisor.serviceJob for supervisor.service failed. See 'systemctl status supervisor.service' and 'journalctl -xn' for details.
 failed!
```

Moreover, this brought down the production site.

The problem is definitely with the config file.  With the line:

```
command = /webapps/openprescribing_staging/bin/gunicorn_start
```

supervisor can start.

But with:


```
command = /webapps/openprescribing_staging/bin/gunicorn_start staging
```

it can't start.

Here are the relevant logs:

<details>
<summary>/var/log/supervisor/supervisord.log</summary>
2018-08-23 10:31:27,087 CRIT Supervisor running as root (no user in config file)
2018-08-23 10:31:27,087 WARN Included extra file "/etc/supervisor/conf.d/antibioticsrct-staging.conf" during parsing
2018-08-23 10:31:27,087 WARN Included extra file "/etc/supervisor/conf.d/antibioticsrct-live.conf" during parsing
2018-08-23 10:31:27,088 WARN Included extra file "/etc/supervisor/conf.d/openprescribing.conf" during parsing
2018-08-23 10:31:27,088 WARN Included extra file "/etc/supervisor/conf.d/openprescribing_staging.conf" during parsing
2018-08-23 10:31:27,106 INFO RPC interface 'supervisor' initialized
2018-08-23 10:31:27,106 CRIT Server 'unix_http_server' running without any HTTP authentication checking
2018-08-23 10:31:27,109 INFO daemonizing the supervisord process
2018-08-23 10:31:27,109 INFO supervisord started with pid 7514
2018-08-23 10:31:28,113 INFO spawned: 'antibioticsrct' with pid 7522
2018-08-23 10:31:28,115 INFO spawned: 'openprescribing_staging' with pid 7523
2018-08-23 10:31:28,117 INFO spawned: 'antibioticsrct-staging' with pid 7524
2018-08-23 10:31:28,119 INFO spawned: 'openprescribing' with pid 7525
2018-08-23 10:31:28,124 INFO exited: openprescribing_staging (exit status 1; not expected)
2018-08-23 10:31:29,591 INFO success: antibioticsrct entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2018-08-23 10:31:29,593 INFO spawned: 'openprescribing_staging' with pid 7565
2018-08-23 10:31:29,593 INFO success: antibioticsrct-staging entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2018-08-23 10:31:29,593 INFO success: openprescribing entered RUNNING state, process has stayed up for > than 1 seconds (startsecs)
2018-08-23 10:31:29,604 INFO exited: openprescribing_staging (exit status 1; not expected)
2018-08-23 10:31:31,611 INFO spawned: 'openprescribing_staging' with pid 7617
2018-08-23 10:31:31,672 INFO exited: openprescribing_staging (exit status 1; not expected)
2018-08-23 10:31:34,678 INFO spawned: 'openprescribing_staging' with pid 7652
2018-08-23 10:31:34,687 INFO exited: openprescribing_staging (exit status 1; not expected)
2018-08-23 10:31:35,688 INFO gave up: openprescribing_staging entered FATAL state, too many start retries too quickly
2018-08-23 10:32:25,676 WARN received SIGTERM indicating exit request
2018-08-23 10:32:25,676 INFO waiting for antibioticsrct, antibioticsrct-staging, openprescribing to die
2018-08-23 10:32:27,735 INFO stopped: openprescribing (exit status 0)
2018-08-23 10:32:27,753 INFO stopped: antibioticsrct-staging (terminated by SIGKILL)
2018-08-23 10:32:28,754 INFO waiting for antibioticsrct to die
2018-08-23 10:32:28,758 INFO stopped: antibioticsrct (terminated by SIGKILL)
</details>


The only difference from starting supervisor successfully is the line:

```
WARN received SIGTERM indicating exit request
```

At the moment, staging is unusable because I didn't back up the old `gunicorn_start` file.

inglesp:
To debug, we can run supervisord in the foreground with the config file that's causing problems, via `supervisord --nodaemon --configuration /etc/supervisor/conf.d/openprescribing_staging.conf`.

evansd:
The "fix" here is that the new config must be introduced via the following incantation:
```
sudo supervisorctl reread
sudo supervisorctl update
```
Simply restarting supervisor will not work. I _think_ this is because, in the absence of an explicitly set process name, supervisor is using the entire command line as a way of keeping track of the processes and has a freak out when the command names in its config don't match what's actually running.

Stopping every process and then restarting supervisor would probably work, but is a bit draconian.

All this reinforces my prior belief (based on mere prejudice) that supervisor is Not Great Software, and we'd be better off moving to systemd which has the advantage of already being there.

evansd:
I have now updated both `openprescribing.conf` and `openprescribing_staging.conf` `/etc/supervisor/conf.d/` so that they pass the appropriate environment as an argument e.g.:
```
command = /webapps/openprescribing/bin/gunicorn_start production
```
This extra argument is currently safely ignored by the production gunicorn script, but it means that this PR can be merged and deployed without any further configuration changes.

inglesp:
> Simply restarting supervisor will not work. I *think* this is because, in the absence of an explicitly set process name, supervisor is using the entire command line as a way of keeping track of the processes and has a freak out when the command names in its config don't match what's actually running.

Wow.

PR #1020, opened 2018-08-30
Include settings in environment

inglesp:
This will mean we won't have to use `--settings` when running management commands.

Before merging, we should update the `environment` file for both staging and production.


sebbacon:
Could you update the README as well?

Issue #1002, opened 2018-08-23
Check for unrecorded Python dependencies in production


inglesp:
We know that `newrelic` was not included in `requirements.txt` (until 7944f9) but we should check for other unrecorded dependencies.

This is a dependency for upgrading to Python 3, as that will involve creating a brand new virtualenv and we'll need to know what to put in it!

inglesp:
Here's the diff of the installed dependencies on the server, vs those in a clean virtualenv after running `pip install -r requirements.txt`:


```
20d19
< cycler==0.10.0
22d20
< dill==0.2.6
32,33d29
< django-leaflet==0.19.0
< django-rest-swagger==0.3.5
37d32
< ebmdatalab-python==0.0.26
44d38
< future==0.16.0
46,56d39
< gapic-google-cloud-datastore-v1==0.15.3
< gapic-google-cloud-error-reporting-v1beta1==0.15.3
< gapic-google-cloud-logging-v2==0.91.3
< gapic-google-cloud-pubsub-v1==0.15.3
< gapic-google-cloud-spanner-admin-database-v1==0.15.3
< gapic-google-cloud-spanner-admin-instance-v1==0.15.3
< gapic-google-cloud-spanner-v1==0.15.3
< gapic-google-cloud-speech-v1beta1==0.15.3
< gapic-google-cloud-vision-v1==0.90.3
< gapic-google-logging-v2==0.10.1
< gapic-google-pubsub-v1==0.10.1
60d42
< google-auth-httplib2==0.0.2
62d43
< google-cloud==0.27.0
64d44
< google-cloud-bigtable==0.26.0
66,77d45
< google-cloud-datastore==1.2.0
< google-cloud-dns==0.26.0
< google-cloud-error-reporting==0.26.0
< google-cloud-happybase==0.20.0
< google-cloud-language==0.27.0
< google-cloud-logging==1.2.0
< google-cloud-monitoring==0.26.0
< google-cloud-pubsub==0.27.0
< google-cloud-resource-manager==0.26.0
< google-cloud-runtimeconfig==0.26.0
< google-cloud-spanner==0.26.0
< google-cloud-speech==0.28.0
79,82d46
< google-cloud-translate==1.1.0
< google-cloud-videointelligence==0.25.0
< google-cloud-vision==0.26.0
< google-gax==0.15.15
86,91d49
< grpc-google-iam-v1==0.11.1
< grpc-google-logging-v2==0.10.1
[...]

PR #1033, opened 2018-09-06
Check that gunicorn starts

inglesp:
We have renamed the `*_ROOT` variables and they are now consistent between the Ansible config and the rest of the code.

To do:

* Make virtualenv path an environment variable in `environment`
* Change `gunicorn_start` to source activate script
* Change README to be right WRT gunicorn
* Add line to `test_playbook.sh` to ensure that gunicorn can start
* Add `gunicorn` and (possibly?) `newrelic-plugin-agent` to requirements

Closes https://github.com/ebmdatalab/sysadmin/issues/26 and #1002.


evansd:
Currently this is failing because we're passing the `--user` and `--group` flags to gunicorn, and the user `hello` doesn't exist inside Travis.

I'd suggest that the this is the wrong place to be trying to switch user in any case. This is already handled by supervisord. Shall we just remove those args?

sebbacon:
Good spot, makes sense

inglesp:
Resolving the conflict should be a case of deleting everything between lines 9 and 17 inclusive at https://github.com/ebmdatalab/openprescribing/pull/1033/conflicts, but as the author of 2b09786c, could you confirm @sebbacon?

PR #1051, opened 2018-09-18
Upgrade highcharts

### LINKED ISSUE #1012: PPU: Text on x axis sometimes overlaps
image
https://openprescribing.net/practice/E81061/0105010B0AAAAAA/price_per_unit/?date=2017-05-01
(Firefox, Chrome, IE11 tested)

(issue extracted from #555)

evansd:
This takes us from v4.1.9 to v6.1.3 which is the current latest version. If the Highcharts [changelog](https://www.highcharts.com/documentation/changelog) is to be trusted then this shouldn't involve any breaking changes. But it could do with a thorough tire-kicking, particularly in older IE versions.

This doesn't automatically fix #1012 as I'd hope it might, but it seems worth doing in any case.


evansd:
This seems to work without issue, even in IE9. IE8 was broken beforehand anyway and we've decided a friendly warning will suffice for that.

PR #1052, opened 2018-09-18
Fix stray numeric labels in PPU charts

### LINKED ISSUE #1050: PPU charts sometimes have stray numeric labels on the x-axis
For example:
https://openprescribing.net/practice/F83032/0601060D0AAA0A0/price_per_unit/?date=2018-06-01

image

These occur because the chart gets automatically expanded to fit the bubbles which means that the x-axis extends beyond the number of category labels we have supplied for it. So Highcharts fills in with numeric labels. We can work around this by supplying our own list of labels which we pad with empty string values.

evansd:
Closes #1050

### DIFF
--- a/openprescribing/media/js/src/bubble.js
+++ b/openprescribing/media/js/src/bubble.js
@@ -42,6 +42,14 @@ domready(function() {
           style: {
             textOverflow: 'none',
           },
+          formatter: function() {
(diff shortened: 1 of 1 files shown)


sebbacon:
Regarding IE testing, it may be  useful to note this (from README):

image

Credentials in `pass`.

PR #1061, opened 2018-09-20
Add warning banner for IE8 and below

evansd:
Our charts currently don't work on IE8. While I did manage to get them
working it involved downgrading some libraries (e.g. the `domready`
package) and polyfilling various various missing Array methods.
Maintaining IE8 compatability therefore looks like it will be a
non-negligable future maintenance burden.

Currently IE8 represents about 1% of our total visitors and 1.7% of our
NHS visitors. So it doesn't seem worthwhile to spend our limited
resources on this. (IE9, by contrast, continues to work without issue.)

Example of the warning in action:
ie8-warning

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ If required, you can run individual Django tests as follows:
 
     python manage.py test frontend.tests.test_api_views
 
-We support IE8 and above. We have a free account for testing across
+We support IE9 and above. We have a free account for testing across
 multiple browsers, thanks to [BrowserStack](www.browserstack.com).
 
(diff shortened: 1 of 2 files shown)


sebbacon:
Looks good, but I think given the site is basically useless in IE8 it might as well be much shoutier - maybe the red-coloured error class, and double height or similar - what do you think?

evansd:
Yes, you're right. I was being far too British about it. How about this?
localhost_8000_

sebbacon:
Much better!

PR #1069, opened 2018-09-23
Ensure constraint_and_index_reconstructor is actually called

inglesp:
constraint_and_index_reconstructor was not used correctly in
import_measures, because although it was called and yielded, the result
was never iterated over.  This PR changes it to be called and used as a
context manager, which was the original intention.

Although this code is exercised in tests, this bug was not detected,
because a test asserted that an empty list (the list of calls) was
greater than zero.  This returns True in Python 2 (where comparing
values of different types is allowed) but raises a TypeError in Python
3.  (Additionally, the wrong thing was being patched.)

This PR fixes the code and the tests to reflect the intended behaviour,
but before merging we should decide whether this is what we actually
want, as we have survived without it for a long time.

### DIFF
--- a/openprescribing/frontend/management/commands/import_measures.py
+++ b/openprescribing/frontend/management/commands/import_measures.py
@@ -655,4 +655,5 @@ def conditional_constraint_and_index_reconstructor(options):
         # updating the entire table.
         yield
     else:
-        yield utils.constraint_and_index_reconstructor('frontend_measurevalue')
+        with utils.constraint_and_index_reconstructor('frontend_measurevalue'):
+            yield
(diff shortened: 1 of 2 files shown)


sebbacon:
Yuck! Subtle bug!

As the comment says, it's an optimisation. For a large table like `MeasureValues` it can be much faster to re-create an index from scratch than do lots of inserts.

I don't know what the potential speedup is in this case. I do seem to recall that Measures is one of the things that takes the longest.  It's a bit of a tricky call: on the one hand, we manage fine without it; on the other hand, the quicker an import finishes, the better for catching odd problems etc.

My instinct is that we should keep this in - the context manager itself works as it's used elsewhere (in the views code) so it doesn't feel like much of a risk to introduce it for the first time here, and the potential win is useful for the reasons above.  I guess without benchmarking there's also a risk it actually worsening performance: do we have data about how long each import step takes somewhere, so we can monitor it?

PR #1072, opened 2018-09-24
Improve measures layout

### LINKED ISSUE #1039: Rejig measure layouts
image

Ensure that alert emails also include the descriptive chart title

And add a legend at the top (and also in emails):

image

### LINKED ISSUE #1040: Remove space under "Explore" in measures text
image

Probably belongs with #1039

sebbacon:
Looks like this:

image

Note that 2 or 3 of the descriptions (the ones that include BNF codes - the second in that screenshot) are unpleasantly long to function as chart titles. I propose that we deploy anyway (it's an improvement overall) and create a new issue to show a list of the BNF codes used in the numerator and denominator as a mouseover or similar (so we're consistent across all measures)

Closes #1039, and closes #1040

### DIFF
--- a/openprescribing/frontend/views/bookmark_utils.py
+++ b/openprescribing/frontend/views/bookmark_utils.py
@@ -622,6 +622,9 @@ def make_email_with_campaign(bookmark, campaign_source):
     return msg
 
 
+def get_chart_id(measure_id):
+    return "#{}-with-title".format(measure_id)
+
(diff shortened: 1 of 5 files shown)


alexwalkercebm:
The new key looks great, but just noticed that you can click on the items to make them disappear on the graph. This works fine apart from for the deciles, as only the first decile is associated with the key item. I imagine this is a non-trivial fix? Can the key be non-interactive?

PR #1073, opened 2018-09-24
Update ktt9_cephalosporins.json

richiecroker:
removed codes from description, added hyperlink and tidied up code


sebbacon:
Bundled into #1081

PR #1080, opened 2018-09-27
Add Sentry's JavaScript error tracking

evansd:
This involves addding the Sentry error tracking library to our JS and
then passing through the account identfier which allows Sentry to
associated these errors with the correct application.

As this identifier is exposed to the world at large it needs to be the
public one, not not the private one which contains our secret key. Sentry
provide the `raven` templatetag library which generates this public ID
automatically for us.

### DIFF
--- a/openprescribing/templates/base.html
+++ b/openprescribing/templates/base.html
@@ -1,10 +1,12 @@
 {% load staticfiles %}
 {% load template_extras %}
+{% load raven %}
 
 <!DOCTYPE html>
 <html lang="en">
   <head>
     <meta charset="utf-8">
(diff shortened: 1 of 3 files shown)


openprescribing/templates/base.html
```diff
@@ -1,10 +1,12 @@
 {% load staticfiles %}
 {% load template_extras %}
+{% load raven %}
```
sebbacon: Could we add a comment here to help soothe my bafflement regarding what `raven` and `sentry_public_dsn` are doing here?  e.g. a link to the docs.  Either as a comment or in the commit message.

evansd:
@sebbacon Just to confirm that you're happy with this now? If so, I'll merge (test failure is just the ansible thing).

PR #1083, opened 2018-09-27
Fix bugs arising from #1053

sebbacon:
* Correct schemas for affected bigquery tables
* Address our bigquery client results correctly


inglesp:
Looks like a sensible addition to the test suite!

PR #1085, opened 2018-10-01
All England dasboard

evansd:


evansd:
@inglesp Yes, I 100% agree. This started out as a hack to be able to get something on screen without having to understand what every bit of the code (particularly the frontend code) did. The struggle has been to get _anything_ working, so I haven't given a great deal of thought yet to what the best way of implementing this would be. Ditto for how we actually test it.

At this stage though, I'm not sure what the best way to proceed is. If we want to do make some more wholesale changes then maybe Brian could demo the dashboard on staging and we could delay merging into production for a few more weeks while I work on it (though I'm slightly running low on steam/enthusiasm tbh)? Alternatively, I could fix up the most egregious bits of hackery (of which faking the MeasureValue objects is probably the worst, I agree :) and we can improve things at a later stage.

Maybe @sebbacon has thoughts here?

By the way, did you make any inline comments on the code? If so, I fear I may have lost them through my over-enthusiastic use of rebase (although I thought github generally hung on to comments, even on outdated commits).

sebbacon:
I think this PR is nice and clear, with the exception of the pseudo org stuff as per @inglesp's comments. I'm happy with the pseudo-org concept but there are opportunities to make it a bit more readable, which I think it's worth taking.

Having said that I also appreciate your flagging enthusiasm for it @evansd!

Refactoring would be much easier (even fun?!) with non-weary eyes, in which case I vote for deploying to staging, parking, and returning on a set day with fresh eyes and a maximum of one day's effort, then shipping it.

evansd:
I have the headline Low Priority savings calculated, but not managed to do anything client side for the individual measures. (I feel like I've lost a fair chunk of the day to deploying and redeploying measure tweaks to staging -- but that's a discussion for another day.)

Below are two screenshots of the dashboard, one at CCG level and the other at practice level. Given that Brian is now going to be demoing the staging dashboard, rather than us trying to deploy anything in time, I wonder if it would make sense to switch the default to practice level? The slow performance shouldn't matter as much and the numbers are certainly more impressive.

localhost_8000_all-england_

localhost_8000_all-england__entity_type practice

sebbacon:
Thanks, I'll run it past Brian tomorrow.

Regarding making measures less painful - I think the solution is the ability for anyone to make measures through the web - not through code. Related to "deciles on demand".

evansd:
@inglesp This is now significantly less hacky than before -- although perhaps not entirely hack-free. Could you take a quick look at some point?

There are now more tests in place, but I need to generate some new test data to properly test the low priority cost saving stuff.

inglesp:
I've had a 20-minute skim, and it looks in decent shape.  There's still an increase in complexity, but it's managed, commented, and tested (a bit?), and I think the complexity is probably unavoidable.  `aggregate_by_measure_and_month` is neat.

Are there any parts you'd appreciate a detailed review on?

PR #1086, opened 2018-10-01
Change all Low Priority measures to be cost based

### LINKED ISSUE #986: Calculate cost savings for all cost-related measures
Similar to #927 which only relates to low-priority measures, should we / can we calculate cost savings for all cost-related measures?

evansd:
Closes #986


evansd:
Closing as this approach won't work. See:
https://github.com/ebmdatalab/openprescribing/issues/986#issuecomment-426215975

PR #1093, opened 2018-10-02
Assume DJANGO_SETTINGS_MODULE is set in the environment

### LINKED ISSUE #1091: Include settings in environment
It would be nice if, rather than having to pass `--settings` every time we run a management command, the settings could be picked up from the environment.  So, when running locally we'd use `openprescribing.settings.local`, in production, we'd use `openprescribing.settings.production`, and so on.  And the tests should only use `openprescribing.settings.test`.

This would be possible by setting `DJANGO_SETTINGS_MODULE` appropriately in `environment`.

I began to work on this in #1020, but there are at least two problems with my approach:

---

#1020 changed various scripts, including `fabfile.py` and scripts run by cron jobs, to not pass `--settings`.  Before deploying, I'd set `DJANGO_SETTINGS_MODULE` in `/webapps/openprescribing/environment`.  But deploying failed because `manage.py collectstatic` was using `antibioticsrct.settings.staging` as its settings file.  I did not have time to investigate, so reverted #1020 in #1058.

This happened because `dotenv.read_dotenv` does not by default override existing environment variables, and `DJANGO_SETTINGS_MODULE` is set in `/etc/profile.d/antibioticsrct_staging.sh` which somehow gets sourced automatically.

The fix is to tweak the call to `dotenv.read_dotenv` to prefer values in the file over values already in the environment.

---

We also need to change `manage.py` to ensure that the tests always run with the correct settings file, because it currently prefers to take `DJANGO_SETTINGS_MODULE` from the environment.  This led to accidental truncation of the dmd tables in BQ.

---

I will open a new PR that addresses these issues, and appreciate a thoughtful review!

inglesp:
Fixes #1091.

The only difference between this and #1020 is that we prefer environment variables in the `environment` file to those already in the environment.

@evansd you suggested blowing up if running the tests with `DJANGO_SETTINGS_MODULE` already set to something that's not `openprescribing.settings.test`, but this wouldn't be compatible with setting `DJANGO_SETTINGS_MODULE` in the `environment` file, which was the goal of this.  Do you agree?


evansd:
Yes, I agree that my suggestion wouldn't work in that case and I think this implementation looks fine.

I suppose I'm slightly against the whole dotenv thing in principle (and anything where applications pull in their config, rather than having it pushed down to them) but that's part of a wider question and certainly in the current context this PR makes sense.

PR #1109, opened 2018-10-08
Fix bar charts broken by Highcharts update

### LINKED ISSUE #1107: Chemical Trend Tool - Displaying same number for Items and Cost
From doorbell
>  - the item value and the spend value seems to be the same figure when I search using the Trend tool. Could you tell me if this is an issue at my end or with the tool please?
[](https://openprescribing.net/chemical/)
> 
This does appear to be the case. 
For example on [Lactulose](https://openprescribing.net/chemical/0106040G0/) the items in July = 238 440 and the spend = £238 440.

The same issue is happening on the BNF sections trend tool. 
See [Proton Pump Inhibitors page](https://openprescribing.net/bnf/010305/)

evansd:
These charts were broken by a change in Highcharts v6.1.0. Specifically
the change:

 * Added feature to animate the series when updating data through
   Series.setData, Series.update or Chart.update. The new logic looks
   for matching X values and can be disabled through setData's
   updatePoints option.

https://www.highcharts.com/documentation/changelog#highcharts-v6.1.0

The fix is to disable the new `updatePoints` logic as described.

Closes #1107

### DIFF
--- a/openprescribing/media/js/src/bar-charts.js
+++ b/openprescribing/media/js/src/bar-charts.js
@@ -84,7 +84,9 @@ var barChart = {
         };
         chart.yAxis[0].update(newYAxisOptions, false);
         var newData = _this.getYValueOfData(data, graphType);
-        chart.series[0].setData(newData, false);
+        // See: https://api.highcharts.com/class-reference/Highcharts.Series#setData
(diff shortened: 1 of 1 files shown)


inglesp:
Unit test failures are unrelated.

PR #1110, opened 2018-10-09
Set settings for e2e tests correctly

inglesp:


inglesp:
The e2e tests haven't been running, because of [this check](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/pipeline/management/commands/pipeline_e2e_tests.py#L25-L27).

evansd:
I see codeclimate is moaning. Should we try to bump it's line length limit to 88 as [recommended](https://github.com/ambv/black#line-length) by Black, which I'd like to start using at some point anyway?

inglesp:
> I see codeclimate is moaning. Should we try to bump it's line length limit to 88 as recommended by Black, which I'd like to start using at some point anyway?

I'd be happy to do that, but I think only Seb can change the codeclimate config.

As for Black: I'm keen too, but it requires Python 3 (although you can run it over a Python 2 codebase) so it'd be easier to wait until #983 is merged.

PR #1115, opened 2018-10-11
Replace doorbell

inglesp:


inglesp:
> I'm happy to take this on if you don't have time at the moment.

I'd appreciate that, thanks.

As for anti-bot defenses, it might be worth looking at what MailChimp do too.

evansd:
I think this is ready to go now, subject to a counter-review from @inglesp

inglesp:
Thanks for taking this on!  I'm happy with all your additions.

We should hold off on merging though until we're sure the whole team can access the shared mailbox -- I have a feeling @brianmackenna can't yet!

brianmackenna:
I can't access @LydiaBerry  has chased IT so hopefully I get in soon.

inglesp:
One question: now that we've changed `SUPPORT_EMAIL` to `"ebmdatalab@phc.ox.ax.uk"`, I suppose we should change `ANYMAIL["MAILGUN_SENDER_DOMAIN"]`.  But then we have `SERVER_EMAIL` set to `"errors@openprescribing.net"`, so presumably we should change that too.  But to what?

evansd:
I would suggest we leave that as it is. We can't send mail as phc.ox.ac.uk so it will have to come from openprescribing.net.

openprescribing/frontend/feedback.py
```diff
@@ -0,0 +1,27 @@
...
+        'message': message,
+    }
+
+    template = get_template('feedback_email.txt')
+    body = template.render(ctx)
+
+    email = EmailMultiAlternatives(
+        subject='OpenPrescribing Feedback: {}'.format(subject),
+        body=body,
+        from_email='{} <{}>'.format(user_name, user_email_addr),
```
inglesp: Is this `from_email` right?  (I know, I wrote it...)

As we're setting `reply_to`, any replies to the email will go to the right place.

And we want to include the sender's name in `from_email`, so that the name shows up in the right place in the inbox UI.

But is using `user_email_addr` the right thing to do here?  We send email through Mailgun, and in the settings, we have:

    ANYMAIL["MAILGUN_SENDER_DOMAIN"] = "openprescribing.net"

I think that means we should be sending from an "@openprescribing.net" address.  @evansd do you agree?

evansd: This is a good point. I think it will work in the sense that the mailgun won't reject the mail (`sender` and `from` don't need to match), but it may not help our spam issues.

Would an alternative be to use just their name in the `from` field, and then include their full email address in the body of the email for searchability?

inglesp: Yes, there are definitely spam considerations:


```
Sender: <peter.inglesby=gmail.com@openprescribing.net>
Message-ID: <20181010183214.1.BC2175AA63CDED91@openprescribing.net>
Reply-To: <peter.inglesby@gmail.com>
To: <ebmdatalab@phc.ox.ac.uk>
From: Peter Inglesby <peter.inglesby@gmail.com>
Subject: Test 4
Content-Type: text/plain; charset="ascii"
MIME-Version: 1.0
X-NTG-DKIM-verify: pass 
X-Oxmail-Spam-Status: score=0.5 tests=FREEMAIL_FORGED_FROMDOMAIN,FREEMAIL_FROM,HEADER_FROM_DIFFERENT_DOMAINS
  0.0 FREEMAIL_FROM          Sender email is commonly abused enduser mail provider
                             (peter.inglesby[at]gmail.com)
  0.2 HEADER_FROM_DIFFERENT_DOMAINS From and EnvelopeFrom 2nd level mail
                             domains are different
  0.2 FREEMAIL_FORGED_FROMDOMAIN 2nd level domains in From and EnvelopeFrom
                              freemail headers are different
```

And note how the sender email address has been changed.  (I'm not sure what's done that.)

> Would an alternative be to use just their name in the from field, and then include their full email address in the body of the email for searchability?

Good idea, I'll do that.

inglesp:
@evansd One final review please?

evansd:
@inglesp Good approach I think. All looks good to me.

sebbacon:
Sorry for drive-by commenting without actually reading the code... but my brain doesn't want to read code at the moment....

When thinking about the right behaviour we should consider that the team will be using the Shared Folder in Outlook. One of the things that we'll find useful is threading messages, which means we want people to use the "reply" button to preserve the message id. This also means we can use the IMAP "unreplied messages" filter.

IIUC the current approach will require the user to hit reply and then change the recipient address? This might be Good Enough but just wanted to check we think this is adequate UX.

inglesp:
> IIUC the current approach will require the user to hit reply and then change the recipient address?

We set `reply_to` to the user's email address:

    email = EmailMultiAlternatives(
        subject='OpenPrescribing Feedback: {}'.format(subject),
        body=body,
        from_email='{} <{}>'.format(user_name, settings.SUPPORT_FROM_EMAIL),
        to=[settings.SUPPORT_TO_EMAIL],
        reply_to=[user_email_addr],
        connection=connection
    )

Does that address your concern, or have I misunderstood?

evansd:
I think most mailgun features can be toggled on and off by including special x-mailgun headers with the email. Let me look into it a bit more.

evansd:
Yep, here we go: https://documentation.mailgun.com/en/latest/user_manual.html#sending-via-smtp

I think `X-Mailgun-Track: no` is what we need.

inglesp:
> I think `X-Mailgun-Track: no` is what we need.

:wizard_hat:

inglesp:
Feedback from @HelenCEBM:

- [ ] Add feedback button at bottom of each page (Doorbell added this automatically)
- [ ] Send copy of feedback to user
- [ ] Show confirmation, eg "Thanks for sending your feedback/query! A copy of the message has been sent to you at joe.bloggs@nhs.net. Please check your spam folder for our reply."
- [ ] Pre-fill the email box for newsletter with the email address the person just entered

sebbacon:
:+1:

inglesp:
@evansd one last review please?

PR #1118, opened 2018-10-15
Fix error when resetting zoom on Analyse bar charts

### LINKED ISSUE #1116: Reset zoom is broken on Analyse form bar charts
The immediate issue is that the x-axis label formatting code is given an int where it expects a string and so blows up [here](https://github.com/ebmdatalab/openprescribing/blob/405a0399247730d5626573e889072be4bd53859c/openprescribing/media/js/src/highcharts-options.js#L182).

This can be worked around by adding a guard clause e.g.:
```js
if ( ! this.value.length) return '';
```

However, while this no longer triggers an error, the chart looks very wrong after resetting the zoom:

localhost_8000_analyse_

I suspect this is probably due to the Highcharts upgrade, or it would have been noticed before.

evansd:
This was caused by the Highcharts upgrade. Specifically, v4.2.7 has this
issue and v4.2.6 doesn't. The solution (or *a* solution) seems to be to
set the x-axis categories explicity rather than relying on each data
point having a `name` property.

It's not entirely obvious how or why this works although the upgrade
contains various changes to the label display logic for category axes so
this fix is at least plausible.

Closes #1116

Highcharts changelog:
https://www.highcharts.com/documentation/changelog#highcharts-v4.2.7

Difference between versions:
https://github.com/highcharts/highcharts/compare/v4.2.6..v4.2.7

### DIFF
--- a/openprescribing/media/js/src/analyse-bar-chart.js
+++ b/openprescribing/media/js/src/analyse-bar-chart.js
@@ -63,7 +63,12 @@ var barChart = {
       barOptions.yAxis.max = globalOptions.maxRatioItems;
     }
     barOptions.series = utils.createChartSeries(dataForMonth);
+    barOptions.xAxis.categories = this._getCategoriesFromSeries(barOptions.series[0]);
     return new Highcharts.Chart(barOptions);
+  },
+
+  _getCategoriesFromSeries: function (series) {
+    return series.data.map(function(d) { return d.name; });
   },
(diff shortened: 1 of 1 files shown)


evansd:
@inglesp That it a reasonable response :) Although an alternative is: let's upgrade Highcharts immediately every time a new release comes out so it's easier to find where the problem is.

PR #1123, opened 2018-10-18
Fix unsubscribe form for users with just one alert

evansd:
The issue was a missing CSRF token which is a single line fix. However I have also added functional tests which now cover this, and also enabled tracking of CSRF failures in Sentry so we get alerted when they happen.

### DIFF
--- a/openprescribing/frontend/tests/test_bookmark_views.py
+++ b/openprescribing/frontend/tests/test_bookmark_views.py
@@ -1,15 +1,8 @@
-import base64
-from mock import patch
-
 from django.test import TransactionTestCase
 from django.core.urlresolvers import reverse
 
 from frontend.models import OrgBookmark
-from frontend.models import SearchBookmark
-from frontend.models import PCT
-from frontend.models import Practice
 from frontend.models import User
-from frontend.models import Measure
 
 
 class TestBookmarkViews(TransactionTestCase):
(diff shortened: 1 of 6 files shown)


openprescribing/frontend/tests/test_bookmark_views.py
```diff
@@ -1,15 +1,8 @@
-import base64
-from mock import patch
```
inglesp: Almost nothing makes me happier than seeing

```
- from mock import patch
```

in red in a diff.

openprescribing/openprescribing/settings/production.py
```diff
@@ -83,15 +83,23 @@
...
+            'class': 'raven.contrib.django.raven_compat.handlers.SentryHandler'
+        },
     },
     'loggers': {
         'django': {
             'handlers': ['gunicorn'],
             'level': 'WARN',
             'propagate': True,
         },
+        'django.security.csrf': {
```
inglesp: Didn't know about this.

openprescribing/frontend/tests/test_bookmark_views.py
evansd: :) It wasn't even used

openprescribing/openprescribing/settings/production.py
evansd: Neither did I until today. But it struck me as really terrible that there's a special magic string you can forget to put in a form and then it just won't work but will never trigger an error so I thought it was worth seeing if Sentry could warn us about it. And it turns out it can quite easily.

PR #1125, opened 2018-10-18
Combine coverage reports from multiple builds

evansd:
At present reported coverage varies wildly, apparently at random (see
https://github.com/ebmdatalab/sysadmin/issues/35). This is probably
because results from the multiple builds aren't being combined and so
there's a race condition.

This is an attempt to configure the repo correctly as described here:
https://docs.coveralls.io/parallel-build-webhook


evansd:
Sadly this doesn't work. And it looks like we're not the only ones having problems:
https://github.com/lemurheavy/coveralls-public/issues/1139
https://github.com/lemurheavy/coveralls-public/issues/1200

I'd like to give [CodeCov](https://codecov.io) a go, but it wants permissions to access the organisation before we use it and because I'm not an admin I can't grant this.

PR #1134, opened 2018-10-30
Create test_strip.json

richiecroker:
prototype for different measure calculation, using a view to do a PPU style filter

### DIFF
--- a/bin/gunicorn_start
+++ b/bin/gunicorn_start
@@ -40,13 +40,10 @@ echo "REPO_ROOT: $REPO_ROOT"
 echo "APPS_ROOT: $APPS_ROOT"
 echo "PYTHONPATH: $PYTHONPATH"
 
-# Environment variables that are set in the "environment" file but
-# need to be marked for export otherwise they won't be passed into
-# child processes
-export NEW_RELIC_LICENSE_KEY
-
-# Source environment file
-source "$REPO_ROOT/environment"
+# Export each variable in the environment file, ignoring comments. Sourcing the
+# file is insufficient as it sets the vars in the current environment but
+# doesn't flag them for export to child processes.
+export $(grep -v '^\s*#' "$REPO_ROOT/environment" | xargs)
 
 # Activate the virtual environment
 source "$VIRTUALENV_PATH/bin/activate"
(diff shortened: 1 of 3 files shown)


inglesp:
Importing the measure fails with `Unrecognized name: p at [39:10]`.

The problem is that the `numerator_where` refers to aliases which are created inside a subquery in `numerator_from`.  I don't really understand SQL's scoping rules, but it looks like aliases created inside a subquery are not visible outside the subquery.

The fix would be to change references to `p` and `r` in the `numerator_where` to `hscic.normalised_prescribing_standard` and `median_prices_per_pack` respectively.

However... `median_prices_per_pack` is a CTE that is also in a subquery, and so I expect that wouldn't be visible either.

Would you be able to create a `median_prices_per_pack` view in BQ, similar to the `pregabalin_total_mg` view?

<details>
  <summary>Here's one of the queries that uses the `numerator_from`, `numerator_where`, etc</summary>

```
SELECT month, practice, SUM(actual_cost) AS numerator 
FROM
  (
    WITH
      prices_per_pack AS (
        SELECT
          month AS date,
          bnf_code,
          IEEE_DIVIDE(net_cost,quantity) AS price_per_pack
      ),
      median_prices_per_pack AS (
        SELECT
          distinct date,
          bnf_code,
          percentile_cont(price_per_pack, 0.5) OVER (PARTITION BY date, bnf_code) AS median_price_per_pack
        FROM prices_per_pack
        order by bnf_code, date
      )

      SELECT
        p.month,
        p.pct,
        p.practice,
        p.net_cost,
        p.actual_cost,
        p.quantity
      FROM
        hscic.normalised_prescribing_standard p
          LEFT JOIN median_prices_per_pack r
            ON p.month=r.date AND p.bnf_code=r.bnf_code
  )
WHERE
  DATE(month) >= '2013-06-01' AND DATE(month) <= '2018-06-01'
  AND
    (  p.bnf_code LIKE '0601060D0%' AND r.median_price_per_pack >=0.2 )
GROUP BY
  practice,
  month
```
</details>

richiecroker:
It was in a view originally, but @sebbacon and I discussed putting into a single query, as a template for others.  I'll leave it @sebbacon and @inglesp to decide what to do!

richiecroker:
Actually, does this answer the question?
https://stackoverflow.com/questions/49571955/error-unrecognized-name-google-bigquery

richiecroker:
@sebbacon  re-written using view

inglesp:
Closing in favour of #1211.

PR #1186, opened 2018-11-13
Redirect CCG/practice homepage requests with "tags" query

### LINKED ISSUE #1182: Create redirects for tag queries on ccg/practice dashboards
Redirect from `/ccg/FOO/?tags=bar` to `/ccg/FOO/measures/?tags=bar` and ditto for practice homepages.

These URLs used to work before the new dashboards and there are links to them in various places which we don't control (see #1181)

evansd:
When the homepages used to show a list of all measures these URLs would
show that list filtered by the selected tags. Now that we only show two
selected measures the URLs don't work as expected. We can fix the
internal links but there are external ones which we don't control and so
it makes sense to set up redirects to handle this.

Closes #1182

### DIFF
--- a/openprescribing/frontend/tests/test_views.py
+++ b/openprescribing/frontend/tests/test_views.py
@@ -488,6 +488,13 @@ class TestFrontendViews(TestCase):
         practices = doc('#practices li')
         self.assertEqual(len(practices), 2)
 
+    def test_ccg_homepage_redirects_with_tags_query(self):
+        response = self.client.get('/ccg/03V/?tags=lowpriority')
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(
+            response['Location'],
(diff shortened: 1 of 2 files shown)


evansd:
I'd be quite strongly against doing this in nginx as it puts an important part of our application logic outside of version control and makes it not reproducible locally.

But your suggestion does make me think that an improvement to this patch would be to pull the redirect decorator out into the `urls.py` file and then just wrap the view functions in there. We could then have a section of the `urls.py` file (or even a new `redirects.py` file which we import) which handles all redirect related stuff and keeps it out of the view logic.

Views on that?

evansd:
@inglesp @sebbacon I've updated the PR with this new approach. I think it's a much nicer separation now, but let me know what you think.

PR #1187, opened 2018-11-13
Use psycopg2 binary installation

### LINKED ISSUE #1183: Switch to using psycopg2-binary package
This shouldn't involve any behavioural changes, it's just a change in the way Psycopg2 is being packaged. See: http://initd.org/psycopg/docs/install.html#binary-install-from-pypi

We want to fix this mainly because it stops the annoying warning message we get at the moment.

evansd:
This is what we're doing currently, but unless we change the package
name we'll have to compile from source once we upgrade to v2.8.

This change is being made because the binary installation can cause
problems under certain circumstances. But this doesn't seem to have
affected us so we're just going to stick to the status quo for now.

For more information see:
http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/

Closes #1183

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -35,7 +35,7 @@ openpyxl==2.3.3
 pandas-gbq==0.3.1
 pandas==0.20.2
 premailer==3.0.1
-psycopg2==2.7.5
+psycopg2-binary==2.7.5
 python-dateutil==2.6.1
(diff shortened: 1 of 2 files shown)


evansd:
This PR is WIP while we decide what we want to do (see discussion in related ticket)

PR #1204, opened 2018-11-19
Ensure all config environment vars are correctly exported

evansd:
This is based on a neat trick/nasty hack from David Winterbottom:
https://twitter.com/codeinthehole/status/1063413713892446208

It is the more general fix for the issue addressed in #1191.

### DIFF
--- a/bin/gunicorn_start
+++ b/bin/gunicorn_start
@@ -40,13 +40,10 @@ echo "REPO_ROOT: $REPO_ROOT"
 echo "APPS_ROOT: $APPS_ROOT"
 echo "PYTHONPATH: $PYTHONPATH"
 
-# Environment variables that are set in the "environment" file but
-# need to be marked for export otherwise they won't be passed into
-# child processes
-export NEW_RELIC_LICENSE_KEY
-
-# Source environment file
-source "$REPO_ROOT/environment"
+# Export each variable in the environment file, ignoring comments. Sourcing the
+# file is insufficient as it sets the vars in the current environment but
+# doesn't flag them for export to child processes.
+export $(grep -v '^\s*#' "$REPO_ROOT/environment" | xargs)
 
 # Activate the virtual environment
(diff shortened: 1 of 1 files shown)


inglesp:
Neato, but is this actually required?  We read the environment file in [`wsgi.py`](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/openprescribing/wsgi.py#L26), so the Django stuff should all have the correct environment.  Is it just the `NEW_RELIC_LICENSE_KEY` that needs to be exported specially?

I suppose it's belt-and-braces, which is probably a good thing.  Is there any way in which reading the environment twice might cause problems?  Not that I can think of...

evansd:
It's certainly not required in the sense that nothing is broken at the moment. The issue with newrelic was that the newrelic-admin command runs outside of the wsgi process and so doesn't get any of its environment variables and exporting that one variable fixed the problem. But I think it's worth doing for the sake of saving future head scratching.

I can't see that setting the environment to the same values twice would cause any issues though I agree it's a bit of smell. But I think the solution to _that_ would be to ditch the dotenv stuff and just set the environment externally. My ideal would be a single entry point script for each environment (i.e. one for staging, one for production) which activates the virtualenv, sets the config env vars and does any other setup required and then `exec`s the rest of its arguments. And then everything -- including the web process, cron jobs, ad hoc management commands -- gets run through that.

So something like:
```
./openprescribing-production gunicorn [...] 
./openprescribing-production ./manage.py import_measures
```
or even, if you need to run several different commands:
```
./openprescribing-production bash
```

Happy to expand more on that if you'd like. But equally happy not to as we've got not shortage of things to do at the moment and it's hardly a top priority :)

inglesp:
I think I like the sound of that.  Reducing number of moving parts is valuable, so what about spending the next couple of sysadmin sessions on implementing it?

As for this PR: :shipit:

PR #1238, opened 2018-11-29
Add RegionalTeam and STP models

inglesp:


openprescribing/frontend/migrations/0030_add_ccg_centroids.py
```diff
@@ -25,5 +25,10 @@ class Migration(migrations.Migration):
             field=django.contrib.gis.db.models.fields.PointField(
                 blank=True, null=True, srid=4326),
         ),
-        migrations.RunPython(set_centroids_without_args),
+
+        # This is now commented out because the SQL generated to execute
+        # set_centroids_without_args includes a reference to fiels which aren't
+        # created until migration 36.
```
sebbacon: So maybe move it to migration 38? Or delete?

openprescribing/frontend/models.py
```diff
@@ -85,6 +105,8 @@ class PCT(models.Model):
     )
     code = models.CharField(max_length=3, primary_key=True,
                             help_text='Primary care trust code')
+    regional_team = models.ForeignKey(RegionalTeam, null=True)
```
sebbacon: I presume you've already thought through why we allow nulls here - might be worth adding a comment explaining.

I want to be sure we don't forget to add STP membership when new CCGs are created.

PR #1244, opened 2018-11-30
Update import_measures for STPs and regional teams

inglesp:


openprescribing/gcutils/bigquery.py
```diff
@@ -188,7 +194,7 @@ def query_into_dataframe(self, sql, legacy=False):
         }
         try:
             return pd.read_gbq(sql, **kwargs)
-        except:
+        except Exception:
             for n, line in enumerate(sql.splitlines()):
```
evansd: Thank you! I have been wanting better BQ error visibility for ages, but not enough to actually sit down and fix it.

PR #1249, opened 2018-12-03
Price concession dashboards

evansd:


evansd:
@inglesp Yes, splitting things into smaller files is definitely a win. I had originally punted on this because there wasn't an obvious existing structure to follow and I didn't feel like inventing one on the fly. But on reflection, even just putting the functions in a dreaded "utils" file at least provides separation and would make any future restructure easier. So I've done that.

PR #1250, opened 2018-12-03
Fix project reference in BQ view SQL

inglesp:


inglesp:
This should fix the e2e tests.

PR #1257, opened 2018-12-04
Preparation for CCG grouping frontend changes

inglesp:


inglesp:
This is now ready for review... but I'm not sure how reviewable it is.

It makes the JS measures code more-or-less agnostic about what kinds of entities it is dealing with, by leaving all the computation to Python code in `views.py`.  This should make it more straightforward to add measure pages for STPs, Regional Teams, and other CCG groupings.

I'll buy a pint for each sensible review comment.

openprescribing/frontend/templatetags/template_extras.py
```diff
@@ -107,3 +108,8 @@ def fancy_join(lst, sep=', ', final_sep=' and '):
         head, tail = lst[:-1], lst[-1]
         lst = [sep.join(head), tail]
     return final_sep.join(lst)
+
+
+@register.filter(is_safe=True)
+def as_json(obj):
+    return mark_safe(json.dumps(obj, indent=2))
```
evansd: This isn't in general safe to do: the characters `<` , `>` and `&` will occur unescaped in the output if they exist in strings inside `obj`. You can fix this by replacing them with their corresponding unicode escape sequences eg. `\u003C`.

I doubt there's a security issue here for us in practice, but it could well be a source of bugs as I can imagine using those characters in description text.

The [json_script](https://github.com/ebmdatalab/openprescribing/blob/be879d180a1b8f350a373e4a2ecf74588020d40d/openprescribing/frontend/templatetags/_json_script_backport.py) tag we backported does almost what you need, however you can't use it as a direct replacement for your `as_json` function because it embeds the whole output in a script tag of type json. This a deliberate decision on Django's part to enforce clean separation of code and data and to allow setting a Content Security Policy header which forbids all inline JS while still embedding data in the page.

Accessing data embedded in this way is slightly more faff, but not loads more (see e.g. [here](https://github.com/ebmdatalab/openprescribing/blob/be879d180a1b8f350a373e4a2ecf74588020d40d/openprescribing/media/js/src/spending-chart.js#L47)) so it might be worth switching to that.

Alternatively you can just copy the escaping code from the `json_script` function.

openprescribing/media/js/src/measure_utils.js
```diff
@@ -342,63 +280,84 @@ var utils = {
     return dataCopy;
   },
 
+  _buildUrl: function(urlTemplate, context) {
```
evansd: :+1:

inglesp:
Thanks.  Will merge and deploy after Brian's meeting.

inglesp:
Well, after Brian's meeting and after I've fixed the failing tests...

PR #1261, opened 2018-12-05
Allow longer lines in PEP8 linter

evansd:
According to CodeClimate, we do this via a "pep8" section in a setup.cfg
file:
https://docs.codeclimate.com/docs/pep8

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[pep8]
+max-line-length = 88


evansd:
The codeclimate failure is due to codeclimate rechecking all line lengths and then re-alerting us to all the overlong lines we had previously ignored.

inglesp:
image

It's well confused.

Issue #1284, opened 2018-12-19
Price Concessions Calculator


brianmackenna:
Some suggestions

- On  the [ALL England page](https://openprescribing.net/all-england/concessions/) the normal cost of GTN is described as £95 for  12,114 pumps.
![image](https://user-images.githubusercontent.com/40460354/50208722-ddd9d200-0369-11e9-8526-bcc90f8cf59a.png)

Tariff prices are [here](https://openprescribing.net/tariff/?codes=0206010F0AACJCJ) but I wonder is it to do with the odd pack sizes?

- Buprenorphine there are significant amount of costs of Buprenorphine that go through "non-CCG routes" as Local Councils are generally responsible for funding it. Local councils and public health departments don't have access to the other NHS data sources commonly used on prescribing (PrescQIPP and ePACT2). My preference would be for us to make this information available in some way either on this dashboard or another method so this data is visible as the Buprenorphine has attracted a huge amount of interest. It is also proportionally a much bigger part of the relevant public health budget and will likely lead to cuts in other services.

Text on page

> Price concessions are a short term agreement by the NHS to pay for more expensive versions of a medicine because the cheaper version has temporarily gone out-of-stock.

Can this be changed please? Its a bit of a pedant point but there is actually stock obtainable when NCSO 
is applied - if there is temporary stock shortage no concession is paid as no product is dispensed.
I suggest
 _Price concessions are a short term agreement by the NHS to pay for more expensive versions of a generic medicine. This occurs because sometimes pharmacy contractors have to dispense an equivalent product that is only available above the set Drug Tariff price. Examples of why this may happen are changes in the value of sterling or manufacturers temporarily suspending production of a generic medicine meaning there is less supply_

- Financial reporting - CCGs will typically report the figures in one of two ways
 a) in the last 12 months we have spent an additional x on NCSO
 b) in the financial year to date we have spent/estimated to spend an additional x in the first x months (only available data and due to the nature of NCSO is it impossible to predict or average to get a 12 month estimate)
 It would be good if a total box could be placed on the page to reflect b) and a) if possible

- Latest month - due to the graphs updating daily we should probably indicate that the latest month is incomplete. Suggest doing it by colouring differently.

- Stray space:
![image](https://user-images.githubusercontent.com/19630/50214133-6874fe00-0377-11e9-9638-b358addd6e8d.png)

- [ ] GTN

- [ ] Buprenorphine

- [x] Text changes

- [x] Financial reporting

[...]

PR #1289, opened 2018-12-19
Add `quantity means pack` field

evansd:
This address the issue raised in #1284 where the calculated tariff costs for "Glyceryl trinitrate 400micrograms/dose pump sublingual spray" (0206010F0AACJCJ) were off by a factor of 200. In this case the quantities were "number of pumps dispensed" but we were treating it as "number of doses" and dividing it by "doses per pump" (which was 200).

This PR doesn't solve the issue in general, but it lays the groundwork for tackling it by adding a field to the presentations table which indicates whether quantity means "packs" or not and updating the NCSO logic to respect this flag.

It also adds a management command which sets this flag for all presentations. At present this just sets it for the one presentation above where we know it's needed and clears it for everything else, but the idea is that we can try various heuristics for setting it more widely or even (dream of dreams) find a canonical data source for it.


evansd:
Incidentally, the random-data-plus-parallel-implementation approach to testing really paid off here. It was trivial to update the data generation to include random values for this flag and to update the parallel implementation to handle it. Whereas I can imagine it being a real pain if we had used handwritten test data and hardcoded expected results.

PR #1314, opened 2019-01-08
Add a temporary hack to block bots from NCSO pages

evansd:
For most of our pages the expensive data loading is done in an API call
triggered by JS. This has the side effect that crawling the site puts
very little load on the server. The NCSO pages are an exception as they
embed their data and this makes them very expensive to crawl to the
extent that it's causing a significant slowdown on the site.

This patch blocks all bots from loading these pages while we figure out
a better solution.


evansd:
@sebbacon It was drive by bug fix. I initially tried rendering the page with an empty context which triggered that bug and I thought I might as well fix it while I was there.

PR #1336, opened 2019-01-16
dm+d browser

inglesp:
I'd like to deploy this today, so that we can point the BSA to pages about things we have questions about.

It's not ready for public consumption.

@evansd / @sebbacon do you want to review?


sebbacon:
Can you deploy to staging also?

inglesp:
> I particularly enjoyed the code generation bits!

Yes, I enjoyed writing it!

> What kind of tests would you like to write, if any?

We should test the importing and the views, using a subset of the data.  There is already a script somewhere in the `dmd` app for generating a self-consistent subset of the TRUD data.

> What is the plan for retiring dmd and moving to dmd2?

There are a number of places where we use the `dmd` models, and we'll need to make sure they use `dmd2` models.  I'll write this up separately.

PR #1338, opened 2019-01-16
Catch correct exception in MailChimp client

inglesp:
This wouldn't have caught the exception class changing when the
dependency was updated (we'd need something like VCR for that) but this
code was untested and now it's not!

### DIFF
--- a/openprescribing/frontend/views/mailchimp_utils.py
+++ b/openprescribing/frontend/views/mailchimp_utils.py
@@ -1,0 +1,48 @@
+import hashlib
+import logging
+
+from django.conf import settings
+from mailchimp3 import MailChimp
+from mailchimp3.mailchimpclient import MailChimpError
+
+from common.utils import get_env_setting
+
+logger = logging.getLogger(__name__)
+
+
+def mailchimp_subscribe(
+        request, email, first_name, last_name,
+        organisation, job_title):
+    """Subscribe `email` to newsletter.
+
+    Returns boolean indicating success
+    """
+    del(request.session['newsletter_email'])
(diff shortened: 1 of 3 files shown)


openprescribing/frontend/views/mailchimp_utils.py
```diff
@@ -0,0 +1,48 @@
...
+        'status': 'subscribed',
+        'merge_fields': {
+            'FNAME': first_name,
+            'LNAME': last_name,
+            'MMERGE3': organisation,
+            'MMERGE4': job_title
+        }
+    }
+    client = MailChimp(
+        mc_user=get_env_setting('MAILCHIMP_USER'),
```
evansd: Do you know why these aren't in `settings`? I'm guessing it's because the MailChimp client used to grab them directly out the environment and then stopped doing that and this was the smallest modification that made it work. That's fine and I don't think it has to be changed immediately, but I think it would be nicer if was done through settings like everything else.

openprescribing/frontend/views/mailchimp_utils.py
```diff
@@ -0,0 +1,48 @@
...
+            subscriber_hash=email_hash)
+        return True
+    except MailChimpError:
+        try:
+            client.lists.members.create(
+                list_id=settings.MAILCHIMP_LIST_ID, data=data)
+            return True
+        except MailChimpError:
+            # things like blacklisted emails, etc
+            logger.warn("Unable to subscribe %s to newsletter", email)
```
evansd: Is it worth getting the reference to the exception object and sticking that in the log message as well so we have some idea as to why the subscription failed?

PR #1350, opened 2019-01-22
(Hopefully) improved layout and wording for dmd product matrix

sebbacon:


sebbacon:
@inglesp I *think* this makes it a wee bit clearer...?

inglesp:
Can you add screenshots?

sebbacon:
image

sebbacon:
What I'd also like to do is add some faded effect (opacity/transparency) to the AMPPs which can't be directly denoted in the dispensing data, i.e. most of them. So the exceptions (AMPPs with their own BNF codes) stand out within the AMPPs sunken area.

Similarly, branded AMPs would be normal whereas generic AMPs would be faded (again because they don't have their own BNF codes).  And then, as the odd exception to the rule, the Becotide 100 AMP [here](https://openprescribing.net/dmd/vmp/320537007/relationships/) would be faded out.

I think that would help make sense of the grid even more, but I'm not sure how it would be implemented, as (if I understand correctly) we're just displaying the mapping they've done for us?

openprescribing/dmd2/views.py
```diff
@@ -194,6 +194,8 @@ def vmp_relationships_view(request, vmp_id):
 
     ctx = {
         'vmp': vmp,
+        'vmpp_cols': len(vmpps),
+        'amp_rows': len(amps),
```
inglesp: Would prefer it if these were called `num_vmpps` and `num_amps`.

openprescribing/dmd2/templates/dmd/vmp_relationships.html
```diff
@@ -35,14 +55,23 @@ <h2>Presentations</h2>
...
-<h2>VMPPs, AMPs, AMPPs</h2>
-<table class="table table-condensed">
+<h2>Related VMPPs, AMPs, AMPPs</h2>
+<p>This table shows how the BNF codes above relate to each other, and to their DM+D equivalents.  Cells are coloured by BNF code; cells without colors are products in DM+D that don't have a corresponding BNF code (and have therefore never appeared in dispensing data).</p>
+<table class="table table-condensed" id="amp-table">
+  <tr>
+    <td></td>
+    <td></td>
+    <th colspan="{{ vmpp_cols }}" class="pp-header">
+      generic product packs
```
inglesp: Except it's not always generic!  I'd stick with "VMPPs".

openprescribing/dmd2/templates/dmd/vmp_relationships.html
```diff
@@ -55,18 +84,19 @@ <h2>VMPPs, AMPs, AMPPs</h2>
   </tr>
   {% else %}
   <tr>
+    {% if forloop.counter0 == 1 %}<th class="amp-header"   rowspan="{{ amp_rows }}"><div class="vertical">AMP/brand/supplier</div></th> {% endif %}
```
inglesp: "AMP/brand/supplier" Again, I think I'd rather this just said "AMP", and describe what this means in the documentation.

openprescribing/dmd2/templates/dmd/vmp_relationships.html
```diff
@@ -1,30 +1,50 @@
...
+  /* vertical line */
+  .vmpp-first-col { box-shadow: 7px 0px 10px -5px #555 inset; }
+  /* vertical(for top left) */
+  .amp-first-row.vmpp-first-col { box-shadow: 7px 7px 10px -5px #555 inset; }
+
+}
+</style>
+
+<h1>VMP {{ vmp.id }} relationships</h1>
+<p>This page shows all the different ways that the <strong>generic product</strong> <span style="background: {{ vmp_colour }}">{{vmp.nm}}</span> can be prescribed.  The generic form is always in <span style="background: {{ vmp_colour }}">this colour</span>. Other BNF codes for the same product (that is, for branded versions or generic product packs) are indicated with other colours.</p>
```
inglesp: VMPs aren't always generic products, although they should be.  Suggest sticking with dm+d terminology (VMP/AMP/VMPP/AMPP) and then explaining what these mean in documentation elsewhere.

openprescribing/dmd2/templates/dmd/vmp_relationships.html
sebbacon: Isn't it? Aren't all VMPPs generic?

PR #1352, opened 2019-01-22
Add links to other measure categories from entity dashboard

### LINKED ISSUE #1333: Show wider range of measures on CCG dashboard
User contacted us re. how to find the Consultation measures. In order to find them you have to go to the standard measures pages and use the drop-down menu; they are not mentioned on the dashboard page at all; i.e. unless coming from the blog post/twitter, users may not be aware they are there at all.

Suggest we mention this on the CCG/practice dashboards, e.g. where highlighted below, **or**, neater but less obvious: after "View the 17 measures" we could say "and the XX currently under consultation"
image

evansd:
Closes #1333

Screenshot:
localhost_8000_ccg_02n_

### DIFF
--- a/openprescribing/frontend/views/views.py
+++ b/openprescribing/frontend/views/views.py
@@ -946,6 +946,11 @@ def _home_page_context_for_entity(request, entity):
         parent_org
     )
 
+    # This will blow up loudly if we remove this tag, at which point we should
+    # also remove the corresponding link from the entity_home_page.html
+    # template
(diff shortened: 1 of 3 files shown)


HelenCEBM:
Great! Though it does sound like the categories are in addition to the measures listed above, which mostly they're not... how about 
> .. plus you can browse all of our measures by topic:

evansd:
Ah yes, thanks @HelenCEBM that's much better

PR #1356, opened 2019-01-23
Fix race condition when sorting by savings

evansd:
If a user hit the "sort by potential savings" button before Highcharts
had finished rendering the charts then the non-cost-saving measure
charts would get removed from the DOM and then Highcharts would attempt
to render to a non-existing element and throw an error.

We fix this by keeping all charts in the DOM and just hiding them when
needed.


inglesp:
Good sleuthing!

Issue #1203, opened 2018-11-18
Several measures don't allow you to break down into individual presentations


brianmackenna:
This was raised by attendees at my talk in Wakelfield who were trying to explore their hypnotics prescribing

Their [dashboard](https://openprescribing.net/ccg/03R/measures/)
![image](https://user-images.githubusercontent.com/40460354/48675414-7a813800-eb50-11e8-86be-23bd40d3b365.png)
 
And the measures

- https://openprescribing.net/ccg/03R/bdzper1000/
https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/frontend/management/commands/measure_definitions/bdzadq.json
- https://openprescribing.net/ccg/03R/bdzadq/
 https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/frontend/management/commands/measure_definitions/bdzadq.json

inglesp:
This affects the following measures:

* bdzadq
* bdzper1000
* ktt9_uti_antibiotics
* lpneedles
* lpzomnibus
* opioidome
* pregabalinmg
* test_strip

This is because the numerators for these measures involve more than aggregating over `hscic.normalised_prescribing_standard`.

There are two classes:

1. numerators that aggregate over a view in BQ
    * lpzomnibus (aggregates `practice_data_all_low_priority`)
    * opioidome (aggregates `opioid_total_ome`)
    * pregabalinmg (aggregates `pregabalin_total_mg`)
2. numerators that aggregate over `hscic.normalised_prescribing_standard` and `JOIN` on another table
    * all the others

In case 2, we can work out which presentations contribute to the numerator with something like:

`SELECT DISTINCT bnf_code FROM {measure.numerator_from} WHERE {measure.numerator_where}`

This is not possible in case 1, as the BQ views do not have `bnf_code` columns.  For these cases, we will have to find another way to work out which presentations contribute to the numerator.  We can either hardcode this (not ideal) or add a field to `Measure` that contains enough SQL to generate the list.

In both cases, we will want to add a `numerator_bnf_codes` column to the `Measure` table, probably an array, which we can then use to generate the list of presentations in `measure_numerators_by_org`.  We may as well do this for measures whose numerator is already queryable, as that should simplify the implementation.

sebbacon:
Sounds good. Could we alter the views in case (1) so they also return a `bnf_code` column?

inglesp:
> Could we alter the views in case (1) so they also return a `bnf_code` column?

[...]

PR #1359, opened 2019-01-23
Add direct links to measure charts

### LINKED ISSUE #1227: Add link to bookmark/share link for individual measures within a page
I propose we add a "Share this" or "Copy link to clipboard" at the bottom of each measure, eg. present Url (which may be CCG, practice or England page) plus `/#[measurename]`.

E.g. as a user I might be looking at the all-england page and see the Opioids measure half way down the page - I want to share this specific measure with someone / on social media, link to it in a paper/blog, or bookmark it for later. However all I can see is the url for the whole page (https://openprescribing.net/all-england/) or a link to breakdown by CCG. 

The average user wouldn't know the short measure name to allow them to do it manually, and it would help us if this were listed on the page as well, so we don't need to remember or look up the short measure name. 

image

Marking as trivial as I think this would be pretty easy! But I may be wrong.

evansd:
We already supported this, but there was no easy way to find out the
correct link for a given chart. This patch adds the link into the panel
heading for each chart and, as an extra usability feature, clicking the
link temporarily replaces it with a text input with the link text
already selected and copies it to the clipboard (where the browser
supports that).

Closes #1227

Screenshot before clicking:
192 168 1 127_8000_ccg_02n_measures__tags eye 2

Screenshot after clicking:
192 168 1 127_8000_ccg_02n_measures__tags eye 1

### DIFF
--- a/openprescribing/media/css/dashboard.less
+++ b/openprescribing/media/css/dashboard.less
@@ -57,6 +57,10 @@
     white-space: nowrap;
     overflow: hidden;
 
+    .panel-heading {
+      white-space: normal;
+    }
+
     .panel-body {
         padding: 5px;
         .inner, .measure-description {
--- a/openprescribing/media/js/src/global.js
+++ b/openprescribing/media/js/src/global.js
@@ -45,4 +45,31 @@ domready(function() {
     $elementsToHide.css('display', 'none');
     $button.appendTo($container);
   });
+
+  // We have to attach the listener to the chart container, rather than to the
+  // individual links, because at this point the charts and their links may not
+  // exist
+  $('#charts').on('click', '.js-shareable-link', function() {
+    var $link   = $(this);
+    var $input = $link.siblings('input');
+    $input.val($link.prop('href'));
+    $input.one('blur', function() {
+      $input.tooltip('destroy');
+      $input.hide();
+      $link.show();
+    });
+    $link.hide();
+    $input.show();
+    $input.select();
+    var copied = false;
+    try {
+      document.execCommand('copy');
+      copied = true;
+    } catch(err) {}
+    if (copied) {
+      $input.tooltip({trigger: 'manual', title: 'Copied to clipboard'});
+      $input.tooltip('show');
+    }
+    return false;
+  });
 });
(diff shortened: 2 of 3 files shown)


evansd:
We could also make the link be automatically copied to the clipboard when the user clicks, but I haven't done this as that gives me a slightly icky feeling. I'd be open to changing my mind on that though if people thought it would be a big usability benefit.

sebbacon:
I think we *should* copy to the clipboard, because it's consistent with [what we do elsewhere](https://github.com/ebmdatalab/openprescribing/commit/c3fe6207641060260872d12409b3cad264c3a92a), icky or not!

One other thought - should we perhaps link to (a version of) the "breakdown by component" page instead of the chart itself? https://openprescribing.net/ccg/99P/ace/. The thing where we shade out all the other measures on a page is cute, but it is also a pretty horrible UI experience if someone's first interaction with the site involves a freezy-loady-75+measures page. Linking to the "by component" means a fast load, plus the opportunity to see other detailed info about the measure on one page (which we could build on in the future - list of "other measures in same categories", for example.

Doing this would also require a couple of extra changes to the "by component" page to make it a more generic "one-measure-view page":

* "Presentations contributing to variation" should be moved below the chart
* That section should also be hidden for measures without the ability to show presentations (q.v. @inglesp's current work)
* Make the "View all measures for X CCG" link more prominent, more like a main navigation element
* I note a bug on that page whereby the text describing total possible savings for all measures appears - which it shouldn't

Thoughts?  @HelenCEBM @evansd ?

evansd:
OK, happy to implement the copy-to-clipboard thing.

I totally agree that our current measure highlighting thing is not great UX and I think your suggestion makes a lot of sense. (I'm assuming you actually meant this URL: https://openprescribing.net/measure/ace/ccg/99P/).

I think we should park this PR temporarily and make a separate issue and PR to improve the breakdown page along the lines you suggest.

HelenCEBM:
good idea, i forgot that the chart was on that page. Should work well with tweaks as described!

inglesp:
One problem with this: not every measure can be broken down by presentation, so we'll have to remove the "Presentations contributing to variation" table in this case.  (Should be easy -- just check `measure.numerator_can_be_queried`.)

And example is https://openprescribing.net/measure/bdzadq/ccg/99P/ which will be fixed (see #1203) but new measures, eg GBGs, may not.

evansd:
This is parked until all #1380 is completed, at which point it can be merged without further work.

PR #1362, opened 2019-01-24
Automatically add banner when import in progress

inglesp:
And, crucially, remove banner when import finishes!


sebbacon:
Oh, and at the beginning!

inglesp:
Good points -- will do so manually today.

PR #1365, opened 2019-01-24
Allow breakdown by presentation for all measures

### LINKED ISSUE #1203: Several measures don't allow you to break down into individual presentations

inglesp:
Fixes #1203.


openprescribing/frontend/management/commands/measure_definitions/lpzomnibus.json
```diff
@@ -30,6 +30,26 @@
   "numerator_where": [
     "1 = 1 "
   ],
+  "numerator_bnf_codes_query": [
```
sebbacon: As this is designed to query the bigquery prescribing dataset (though perhaps it should be querying something smaller like the presentations table?), the column name should be `bnf_code`, not `presentation_code`, I think.

sebbacon:
I would also like to discuss if we should consider allowing a custom "breakdown" page.  It would be nice for the ghost generic breakdown to be linkable in the same way (i.e. a link to https://openprescribing.net/ccg/99P/ghost_generics/ from the ghost generics measure).

That breakdown is sufficiently different from the others (in particular it uses very custom SQL to return custom columns) it feels to me like an exception that is worth supporting - but perhaps with more thought we could generalise it.

openprescribing/frontend/management/commands/measure_definitions/lpzomnibus.json
inglesp: It's querying the `bnf` table, so this is not a problem.

openprescribing/frontend/management/commands/import_measures.py
```diff
@@ -221,54 +226,82 @@ def create_or_update_measure(measure_id):
...
+    measure.is_percentage = v['is_percentage']
+    measure.low_is_good = v['low_is_good']
+    measure.numerator_bnf_codes_query = v.get('numerator_bnf_codes_query')
+    measure.numerator_bnf_codes = get_numerator_bnf_codes(measure)
+    measure.save()
+
     return measure
 
 
+def get_numerator_bnf_codes(measure):
```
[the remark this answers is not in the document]
inglesp: This is tested in `test_import_measures_new` where we check that `numerator_bnf_codes` has been set correctly:

```
self.assertEqual(m.numerator_bnf_codes, ['0703021Q0BBAAAA'])
```

inglesp:
Thanks for the review.  I've made the suggested fixes.

> Slightly concerned about the possible performance impact of the lpzomnibus query which expands to > 600 bnf_codes.

I hadn't really considered this!  8-12 seconds isn't much longer than an analyse query, but we could easily mark this measure as not being breakdownable if it turns out to cause performance degradation elsewhere on the site.

> I would also like to discuss if we should consider allowing a custom "breakdown" page.

What would the GBG breakdown page show?  I think this might be out of scope for this PR.

sebbacon:
>> Slightly concerned about the possible performance impact of the lpzomnibus query which expands to > 600 bnf_codes.

> I hadn't really considered this!  8-12 seconds isn't much longer than an analyse query, but we could easily mark this measure as not being breakdownable if it turns out to cause performance degradation elsewhere on the site.

Fair point. All the same, I think we should just test it on staging first, as I estimated that based on a similar query but not quite the same - it's also going to return a *lot* of rows which could also affect performance.  At the least if it's going to take more than 10 seconds we should consider adding a more obvious wait clue  than "loading please wait" text?

inglesp:
To get this deployed, I'd like to set `numerator_is_list_of_bnf_codes=False` on the `lpzomnibus` measure, and come back to it if there's demand.  In any case, for this measure, it's possible to get a breakdown for each constituent measure.

@sebbacon would you be happy with this?

PR #1393, opened 2019-01-29
Prevent duplicate import warnings from being displayed

evansd:
When a request is redirected the middleware would add an extra warning
message for each hop of the redirect, which could sometimes be three or
more.

### DIFF
--- a/openprescribing/pipeline/middleware.py
+++ b/openprescribing/pipeline/middleware.py
@@ -2,17 +2,32 @@ from django.contrib import messages
 from pipeline.runner import in_progress
 
 
+IMPORT_WARNING_TAG = 'import_warning'
+
+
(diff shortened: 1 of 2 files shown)


openprescribing/pipeline/middleware.py
```diff
@@ -2,17 +2,32 @@
...
         return response
     return middleware
+
+
+def _should_add_import_warning(request):
+    if not in_progress():
+        return False
+    # If we add a message to an API request, then depending on whether the API
+    # requests get made in parallel, Django can end up updating the cookie
+    # multiple times, causing the message to be displayed multiple times.
```
inglesp: As we have the more robust check below, we could change this comment to something like: "no need to set the warning on API requests".

PR #1394, opened 2019-01-29
Use templated import warning instead of contrib.messages

evansd:
The previous approach had issues with showing duplicate warnings
following redirects and this approach, while cruder, is simpler and more
reliable.


openprescribing/openprescribing/settings/base.py
```diff
@@ -151,7 +151,8 @@
                 'frontend.context_processors.support_email',
                 'frontend.context_processors.google_tracking_id',
                 'frontend.context_processors.api_host',
-                'frontend.context_processors.debug'
+                'frontend.context_processors.debug',
+                'pipeline.context_processors.import_in_progress'
```
inglesp: Add a trailing comma?

evansd: Haha! You know I actually did out of habit and then realised that the list didn't have one previously and so removed it for consistency...

evansd: I'm back on another branch now so I think that might be an enhancement for another day

inglesp: 👍

PR #1397, opened 2019-01-29
Automatically approve alert signups

evansd:
We still send verification emails as we're using this as a form of login
mechanism (by verifying an email address you become logged in as that
user) for managing alerts.

With the verification emails disabled you can view and manage anyone's
alerts just by entering their email address, which feels like a step too
far.

### DIFF
--- a/openprescribing/frontend/views/views.py
+++ b/openprescribing/frontend/views/views.py
@@ -1239,23 +1239,16 @@ def _authenticate_possibly_new_user(email):
     return authenticate(key=user.profile.key)
 
 
-def _unapprove_bookmarks_when_different_user(user, request):
-    # An unverified account can only create unapproved bookmarks.
-    # When an account is verified, all its bookmarks are
-    # approved. Whenever someone tries to add a bookmark for
-    # someone else's email address (or they're not logged in),
-    # that email address is marked as unverified again.  In this
-    # way we can allow people who remain logged in to add several
-    # alerts without having to reconfirm by email.
+def _unverify_email_address_when_different_user(user, request):
+    # This is weird. Because entering any email address logs you in as that
+    # user (see force_login function below) we need to prevent accessing
+    # someone's bookmarks just by signing up using their email address. This is
+    # done by unverifying that email address if you're not already logged in as
+    # that user and then forcing a re-verification before you can access
+    # existing bookmarks.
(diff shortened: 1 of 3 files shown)


evansd:
@inglesp Not quite. If a user clicks the relevant link in an alert email then they will be logged in and can manage their bookmarks. And they can do that from as many different machines as they like and be logged in on all of them.

If they (or indeed anyone) sign up for an alert using that email address on a machine on which they are *not* logged in then the email address becomes "unverified" and they will be logged out on all machines.

Once they click the link to verify their email again they will be logged in on that machine but not, I think, on any others.

Make sense?

No, of course it doesn't.

I think this is one of those cases where allauth is hindering us more that's it's helping. In fact, I have a deep aversion to authentication frameworks born of much frustrating experience. If someone wants to package up, say, the OpenID connect protocol in a nice _library_ for me to use then great. But as soon people start providing models and views and two hundred configuration options you know you're in for a bad time.

I'm not even a massive fan of Django's built in auth stuff for just this reason. But it does have the benefit of being fairly minimalist and well understood, and it's improved a lot over the last few releases.

openprescribing/frontend/views/views.py
```diff
@@ -1364,8 +1356,10 @@ def _handle_bookmark_and_newsletter_post(
             request.session['alerts_requested'] = 1
             user = _authenticate_possibly_new_user(email)
             form_args = _make_bookmark_args(user, form, subject_field_ids)
-            form_args['approved'] = _unapprove_bookmarks_when_different_user(
-                user, request)
+            _unverify_email_address_when_different_user(user, request)
+            # We're automatically approving all alert signups from now on
+            # without waiting for the email address to be verified
+            form_args['approved'] = True
```
inglesp: Do we need `approved` on the form anymore?

inglesp:
I (sort of) see, thanks for the detailed explanation.

openprescribing/frontend/views/views.py
evansd: I don't think we can (yet) get rid of the `approved` state because we don't want to retroactively approve all the existing bookmarks and start sending emails to people who have never got them before.

We talked about maybe doing a mailout to all address who never verified and ask if they'd like to start receiving them.

inglesp: I see.

Issue #1380, opened 2019-01-28
Tweaks to the "single measure for one entity" page 


evansd:
Connected to issue #1359 we want to make linking to an individual measure for an individual organisation easier. We already have a page (e.g. https://openprescribing.net/measure/ace/ccg/99P/) that mostly serves this purpose, but at the moment it's too focused around providing a breakdown of presentations which we can't do (yet) for every measure.

However, with some hopefully small tweaks it can be made to do what we need. These tweaks are (as described by @sebbacon):

 - [x] "Presentations contributing to variation" should be moved below the chart
 - [x] That section should also be hidden for measures without the ability to show presentations (q.v. @inglesp's current work)
 - [x] Make the "View all measures for X CCG" link more prominent, more like a main navigation element
 - [x] I note a bug on that page whereby the text describing total possible savings for all measures appears - which it shouldn't

Once this is complete, #1359 can be merged.

PR #1398, opened 2019-01-29
De-duplicate "measure for one entity" pages

evansd:
As a precursor to #1380 we want to make sure there's only one page to tweak, not two nearly-but-not-quite identical ones.


openprescribing/frontend/views/views.py
```diff
@@ -311,32 +311,20 @@ def measure_for_all_ccgs(request, measure):
     return render(request, 'measure_for_all_ccgs.html', context)
 
 
-def measure_for_one_practice(request, measure, practice_code):
-    practice = get_object_or_404(Practice, code=practice_code)
+def measure_for_one_entity(request, measure, entity_code, entity_type=None):
```
inglesp: Why would `entity_type` ever be `None`?

evansd: Ah yes, good spot. I copied that without thinking from a function which handles the "All England" case.

PR #1405, opened 2019-01-30
Change wording on confirmation emails to be less false

evansd:
We stil send confirmation emails even though we don't now require
verification before sending alerts (because if we turn them off
altogether than anyone can view and edit anyone else's alerts just by
entering their email address). But the wording in our confirmation
emails needs to be tweaked so we don't say that you can ignore the email
if you don't want alerts.

### DIFF
--- a/openprescribing/templates/account/email/email_confirmation_signup_message.html
+++ b/openprescribing/templates/account/email/email_confirmation_signup_message.html
@@ -1,11 +1,11 @@
 {% load account %}
 <p>Hello from OpenPrescribing!</p>
 
-<p>You're receiving this e-mail because someone has requested monthly alerts about {{ user.profile.most_recent_bookmark.topic }}.</p>
+<p>You've been subscribed to monthly alerts about {{ user.profile.most_recent_bookmark.topic }}.</p>
 
-<p>To confirm this is correct, please follow <a href="{{ activate_url }}">this link</a>.</p>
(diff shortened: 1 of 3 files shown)


HelenCEBM:
Have you got suggested wording to be checked? I can't see it...

evansd:
@HelenCEBM Oh, sorry. It's in the "Files changed" tab. There's a green "Review changes" button there which will let you add comments and approve/disapprove. And you can also click the blue plus symbol on individual lines if you want to comment on a specific line.

evansd:
Thanks @HelenCEBM. Unfortunately the unsubscribe link isn't available to us at the point that email goes out. I've added the line you suggested about being able to unsubscribe at any point using the link in each email.

The reason for all these limitations, by the way, is that we're using an open-source package that promises to do all this stuff for us -- which it does; so long as you don't try to deviate at all from the way it's designed to work. It breaks my heart knowingly delivering a slightly shoddy user experience like this, but right now we just don't have time to pull it all out and do it ourselves.

inglesp:
It's annoying we can't generate the text of the email from the HTML!

evansd:
One solution I've seen is to write the content in markdown and use that for the plain text and then render to HTML. It comes with it's own pitfalls, but can work quite nicely as long as everything you want to do has a reasonably readable markdown representation.

PR #1412, opened 2019-01-30
Tweaks to single measure page

### LINKED ISSUE #1380: Tweaks to the "single measure for one entity" page

evansd:
This addresses the four requirements we had before we can make this a generic landing page for a single measure on a single organisation:
- "Presentations contributing to variation" should be moved below the chart- 
  That section should also be hidden for measures without the ability to show presentations
-   Make the "View all measures for X CCG" link more prominent, more like a main navigation element
-  I note a bug on that page whereby the text describing total possible savings for all measures appears - which it shouldn't

Closes #1380

Screenshot for a measure which does support numerator breakdown:
localhost_8000_measure_lyrica_ccg_99p_

And screenshot for one that doesn't:
localhost_8000_measure_bdzper1000_ccg_99p_

### DIFF
--- a/openprescribing/templates/measure_for_one_entity.html
+++ b/openprescribing/templates/measure_for_one_entity.html
@@ -10,18 +10,22 @@
 {% endblock extra_css %}
 
 {% block content %}
-  <h1>{{ entity.name }}: {{ measure.name }}</h2>
+  <h1>
+    {{ measure.name }}
+    <br><small>{{ entity.name }}</small>
+  </h1>
+  <p>
+    <a href="{% url measures_url_name entity.code %}" class="btn btn-default">
+      View all measures for this {{ entity_type }} &rarr;
+    </a>
+  </p>
 
   <p>This measure shows how this organisation compares with its peers across NHS England. This is indicative only, and should be approached with caution. <a href='{% url 'faq' %}#measureinterpret'>Read more about measures</a>.</p>
 
   {% include '_measures_heading.html' with entity_type=entity_type %}
   {% include '_measures_panel.html' %}
 
-  <p>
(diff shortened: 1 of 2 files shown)


openprescribing/templates/measure_for_one_entity.html
```diff
@@ -10,44 +10,53 @@
...
 
   {% include '_measures_heading.html' with entity_type=entity_type %}
   {% include '_measures_panel.html' %}
 
-  <p>
-    <a href="{% url measures_url_name entity.code %}">View all measures for {{ entity.name }}</a>
-  </p>
+  <hr>
+
+  {% if measure.numerator_can_be_queried %}
```
evansd: @inglesp Will this conditional still be correct after your numerator breakdown work is merged?

inglesp: No, it should be `measure.numerator_is_list_of_bnf_codes`.

PR #1426, opened 2019-02-04
Update message for resend of confirmation emails

evansd:
We've now stopped requiring confirmation of email address for alerts,
but we're not applying this policy retrospectively so for existing
unconfirmed subscribers we're going to try one last message to get them
to confirm.


sebbacon:
I would add something like "This may have been up to two years ago" just so they're not too surprised!

While you're here, is it worth thinking about maybe documenting this method of resending emails? Or possibly removing it now we don't require confirmation?

inglesp:
> I would add something like "This may have been up to two years ago" just so they're not too surprised!

I agree.  Apart from that, fine.

PR #1428, opened 2019-02-04
Fix link to Ghost Branded Generics measure

evansd:


evansd:
Now I now this isn't a "panic, panic, fix this now" PR I'm minded to spend the little bit of extra time to merge the CCG and practice views along the same lines as we have elsewhere. Thoughts?

sebbacon:
This looks good, and happy to merge.

I agree about fixing it properly though, if you have the energy - there's been some grand work paying down tech debt on measure URLs lately and it would be nice to finish the job.

evansd:
@sebbacon Just done it as it was easy to do now when I'd just been looking at the views and they're still nice and simply. Worth another quick once over from you I think.

sebbacon:
Looks great.

This is a good example of where I should have written a test or been more careful about testing on staging, per the point I raised last week!

Can we deploy to staging and ask @HelenCEBM to test - links both at practice and CCG level?

Issue #1435, opened 2019-02-05
Libre measures display


brianmackenna:
**Bug**
The CCGs boxes are not displaying uniformly in rows of three for the Libre measure
![image](https://user-images.githubusercontent.com/40460354/52299105-81b02900-297c-11e9-8b05-7389a58db097.png)

**User Feedback**

> Why do we display five year when there is only about 1 year of prescribing data?

I thought this was interesting question especially if we start to display a lot of new products and assess uptake.  It sounds like  a bad idea if we want to keep every measure visually comparable across the site. No idea of technical difficulty in altering

brianmackenna:
More pages appear to be affected by box display problem, which I think means them all. Examples
https://openprescribing.net/ccg/03N/statinintensity/
https://openprescribing.net/measure/ktt9_uti_antibiotics/

richiecroker:
I think the CCG name font size has increased - which may be causing the problem

inglesp:
Because, _of course_, `.panel-title` is [defined in bootstrap](https://getbootstrap.com/docs/3.3/components/#panels-heading).

inglesp:
![image](https://user-images.githubusercontent.com/28734/52311360-9dc4c200-299e-11e9-885d-415115efdac2.png)

inglesp:
The font size issue is fixed, which means that panels now line up properly.   (At least, most of the time.  In the past I've seen the occasional odd alignment, but can't find an example now.)

inglesp:
As for:

> Why do we display five year when there is only about 1 year of prescribing data?

My gut feeling is that it'd be fiddly to use different x-axis scales for different measures, and could cause confusion on pages showing all measures for one entity.

Additionally it's potentially useful to see that prescribing for a given measure began recently.

So I'd like to close this issue now -- @brianmackenna, are you happy to do so?

PR #1436, opened 2019-02-05
Avoid clashing with Bootstrap class name

inglesp:
Addresses part of #1435


evansd:
Sorry, I reviewed this and didn't spot the problem because I thought you were deliberately using the bootstrap class to get the bootstrap behaviour!

PR #1438, opened 2019-02-05
Set numerator_bnf_codes_query on time-dependent measures

inglesp:


inglesp:
Before merging, need to test importing these measure on staging, and check that `numerator_bnf_codes` looks sensible.

But @sebbacon could you take a quick look and check this looks reasonable?

inglesp:
Have tested these on staging and they look sensible.

inglesp:
@sebbacon have added a link to https://github.com/ebmdatalab/openprescribing/wiki/Measure-definitions.  Is there anything you'd like to add?

PR #1440, opened 2019-02-06
Add console log

inglesp:
This is in the hope of getting more information out of Sentry about the
TypeErrors occurring in this function.

### DIFF
--- a/openprescribing/media/js/src/highcharts-options.js
+++ b/openprescribing/media/js/src/highcharts-options.js
@@ -179,6 +179,11 @@ barOptions.xAxis.type = 'category';
 barOptions.legend.enabled = false;
 barOptions.xAxis.labels = {
   formatter: function() {
+    if (typeof this.value === 'undefined') {
+      // We hope this will be logged by Sentry to give us some clues about why
+      // this.value is sometimes not defined.
+      console.log(barOptions);
+    };
     var str = this.value.substring(0, 17);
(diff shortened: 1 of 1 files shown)


openprescribing/media/js/src/highcharts-options.js
```diff
@@ -179,6 +179,11 @@ barOptions.xAxis.type = 'category';
 barOptions.legend.enabled = false;
 barOptions.xAxis.labels = {
   formatter: function() {
+    if (typeof this.value === 'undefined') {
+      // We hope this will be logged by Sentry to give us some clues about why
+      // this.value is sometimes not defined.
+      console.log(barOptions);
```
evansd: `console` isn't defined in IE unless the dev tools panel is open so this will throw another error, potentially confusing things.

inglesp: Oof, did not know that!

inglesp:
> When I last looked at these errors they were quite easily reproducible locally so I'm not sure it's more logging data we need to fix them.

We couldn't reproduce them!  Do you have notes in an issue?

evansd:
I didn't get as far as making notes, but I've just reproduced the first instance of that issue I found in Sentry. The trick is to look at the `ui.click` events in the breadcrumbs to see what sequence of events the user followed. In this case it required zooming in and then clicking the Spending/Items toggle.

PR #1443, opened 2019-02-07
Non judgemental tweak

sebbacon:


openprescribing/templates/bookmarks/email_for_measures.html
```diff
@@ -134,14 +134,14 @@ <h3 class="better">Prescribing measures which are outliers</h3>
   {% if stats.interesting %}
       {% for measure in stats.interesting %}
         {% if forloop.first %}
-          <p>Over the last three months, we found that this {{ bookmark.org_type }} was in the top or bottom 10% on
+          <p>Over the last three months, we found that this {{ bookmark.org_type }} deviated a long way from the median practice on
```
evansd: Nitpick, but I think you want `the median {{ bookmark.org_type }}` rather than `the median practice`

Issue #1366, opened 2019-01-24
Add alerts for NCSO concession changes


inglesp:
Users should be able to sign up for alerts to tell them when their organisation is affected by NCSO concession changes.

HelenCEBM:
From team chats so far:
- These will be triggered several times a month, any time that the NCSO list is updated
- Users opting to sign up for alerts for a practice/CCG should then have the following options (phrasing could be improved)
> please select whether you would like to receive alerts for:
> (1) NCSO costs (approx weekly)  - standard measures alerts will be included in each email, updated monthly; OR
> (2) Just standard measures (monthly)
   
- [ ] Question: how to deal with users already signed up for normal alerts who'd like to "upgrade"? Can we supply a link in the routine monthly alerts to let them know this is now an option? Will they have to manually unsubscribe from their "old" alerts?

- In the email, after the NCSO alerts there should be a line under which to show the normal alerts. 
- [ ] We should make it as clear as possible when the normal alerts are updated, e.g. by highlighting prominently which month the regular alerts apply to, or with some kind of "New Month!" message if possible.

inglesp:
(@brianmackenna, @richiecroker)

HelenCEBM:
I'd say remove or edit the first line, (`New price concession data has been published for January 2019.`) because it will come up several times and might look similar each time. 

E.g. if there is a new concession approved mid-month, but this CCG only prescribes a small amount of it, we will presumably trigger an alert but there might be no change to the top 10 list. Do we have a way of handling that? 

If we keep the line we could change it to 
> New or updated price concession data has been published for January 2019

richiecroker:
I would go with:
New or updated price concession data has been published for January 2019

richiecroker:
but apart from that, looks good!

HelenCEBM:
After this:
> We estimate that in January 2019 price concessions for out-of-stock medicines will cost NHS Birmingham and Solihull CCG an additional £285,780. 

Can we say something like:
> - an increase of £XXX since the last price concession list was released.

inglesp:
> New or updated price concession data has been published for January 2019

Good idea.

inglesp:
> Can we say something like:
>>    -- an increase of £XXX since the last price concession list was released.

Not easily -- this would involve keeping track of what it was last time around, and would significantly increase the complexity of the work.

I'd like to get something simpler deployed first, but we could come back and revisit this.

richiecroker:
how about something like:
[...]

PR #1447, opened 2019-02-07
Add alerts for NCSO concessions

inglesp:
Note that this doesn't include code for sending the alert emails!

The form looks like this:

image


openprescribing/frontend/tests/test_spending_views.py
```diff
@@ -107,6 +107,17 @@ def test_spending_views(self):
...
+        url = '/practice/{}/concessions/'.format(self.practice.code)
+        data = {
+            'email': 'alice@example.com',
+            'practice': self.practice.code,
+            'newsletters': ['alerts'],
+        }
+        response = self.client.post(url, data, follow=True)
+        self.assertContains(
+            response, "Check your email and click the confirmation link")
+
```
sebbacon: Can we test a `NCSOConcessionBookmark` is created here? Possibly also with the optional newsletter signup too?

openprescribing/templates/spending_for_one_entity.html
```diff
@@ -33,6 +34,28 @@ <h1>{{ title }}</h1>
 
 <hr/>
 
+{% if form %}
+  {% if not signed_up_for_alert %}
+  <form method="post" class="form form-inline">
+   {% csrf_token %}
+   <p>
+     Get alerted by email when new price concession data is available.
```
sebbacon: add something like "(up to 4 times per month)" (or whatever is accurate) so they know expected frequency?

inglesp:
> I just want to check that I'm understanding correctly that this follows the pattern of the other alerts...

Yes, it all goes through `_handle_bookmark_and_newsletter_post`.

inglesp:
I think this is now done, except that `send_ncso_concessions_alerts` will need to be manually called for the time being.

@evansd are you able to review this week?

evansd:
@inglesp Have re-reviewed. Looks good :+1:

inglesp:
Fixes #1366 and #1458.

PR #1449, opened 2019-02-08
Support passwordless deploying

sebbacon:
This relies on the deploying user being a member of the group
`fabric`, which then has passwordless sudo access to anything in the
folder `deploy/fab_scripts/`

Relates to ebmdatalab/sysadmin#42

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -277,8 +277,8 @@ def clear_cloudflare():
         requests.delete(url % ZONE_ID + '/purge_cache',
                         headers=headers, data=json.dumps(data)).text)
     if result['success']:
-        print "Cloudflare clearing succeeded: %s" % \
-            json.dumps(result, indent=2)
+        print("Cloudflare clearing succeeded: %s" %
+              json.dumps(result, indent=2))
     else:
         warn("Cloudflare clearing failed: %s" %
              json.dumps(result, indent=2))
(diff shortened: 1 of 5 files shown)


fabfile.py
```diff
@@ -36,6 +36,38 @@
...
+        sudo_cmd = 'sudo -u www-data '
+    else:
+        sudo_cmd = 'sudo '
+    return run(sudo_cmd +
+               os.path.join(
+                   env.path,
+                   'deploy/fab_scripts/%s' % script))
+
+
+def setup_sudo():
```
evansd: It's worth being extra cautious here because if we introduce a syntax error when creating that file we will break our whole sudo config which will make it impossible to sudo and this is very tricky to recover from.

The safe way to do this is to first write to a temporary file in `/etc/sudoers.d` making sure that the temporary file has a dot somewhere in the name (any filename containing a dot anywhere is explicitly ignored by the `#includedir` directive). You can then run `visudo -cf <temp-filename>` which will check the syntax of the file and exit with an error code if it's invalid. If it succeeds then rename the temporary file to its final name.

As belt and braces, the first time we run this in production it would be good to do so with a mosh session open that has already got a root shell with `sudo -i` so we can always recover from a borked sudo config.

It's a bit of extra faff but I do think it's worth it in this case.

sebbacon: Yeh I guess you're right!

PR #1468, opened 2019-02-13
Remove resend_confirmation_email command and related templates

evansd:
Now we no longer require confirmation, and now that we've sent the final
round of reminders, this command is no longer required.

Before remove the file, this PR also adds a simple log file to prevent
double-sending which was used in the final round of reminders -- just in
case we ever need to resurrect this command.


inglesp:
There's also `tempate_override` settings.  Could these be removed?

PR #1469, opened 2019-02-13
Handle VPIDs changing between dumps

inglesp:


openprescribing/dmd/management/commands/import_dmd.py
```diff
@@ -295,6 +295,30 @@ def process_datafiles(source_directory):
...
 
 
+def update_product_references():
+    """Update FK references to DMDProducts.
+
+    We need to do this because sometimes a VMP's ID changes between releases of
+    the data.  When this happens, the VMP's previous ID is available in the
+    vpidprev column.
+
+    It is only VMPs that might change their ID, so we do not need to consider
```
sebbacon: Do we know this because its only them that have a `idprev` field?

openprescribing/dmd/tests.py
```diff
@@ -139,11 +127,41 @@ def test_import_dmd(self):
...
+        self.assertFalse(DMDProduct.objects.filter(dmdid=22480211000001104).exists())
+        self.assertQuery(
+            'SELECT count(*) FROM dmd_vmp WHERE vpid = 22480211000001104',
+            0
+        )
+
+        # Check that the TariffPrice has been updated to reference a new
+        # DMDProduct.
+        tp.refresh_from_db()
+        self.assertEqual(tp.product_id, 12345)
```
sebbacon: Maybe check the VMPP foreign key here, too, particularly given we plan to excise DMDProduct in the nearish future?

openprescribing/dmd/management/commands/import_dmd.py
inglesp: Yes.

openprescribing/dmd/tests.py
inglesp: `DMDVmpp` doesn't have a foreigin key to `DMDProduct`.

sebbacon: I mean `self.assertEqual(tp.vmpp_id, 12345)`

inglesp: I don't think you do mean that!  But I've added a test at https://github.com/ebmdatalab/openprescribing/blob/96ca0642373273f990b42fdf08acc93a4fff470d/openprescribing/dmd/tests.py#L165-L171 which should cover what I think you mean.

PR #1476, opened 2019-02-15
Fabric-related tweaks to support ebmbot

sebbacon:


sebbacon:
This *should* fix problems @inglesp had on Thurs, plus a couple of other changes required by the bot

PR #1497, opened 2019-02-21
Add summary totals to the Analyse page

### LINKED ISSUE #1019: Analyse Page - Total for NHS England
On the analyse page it would be helpful for people in national organisations (and others I imagine) to be able to see the "Total Items" and "Total Cost" on queries they do. It links to the All England update as well.

It would be good display as the "NHS England Total" for the 
a) month selected 
b) the last 12 months

Potentially in a box above/below the CCG/Practice graph

evansd:
If no CCGs are highlighted then totals are across all of NHS England. Otherwise totals are for the selected CCGs or practices.

Results appear in a table below the chart like so:
localhost_8000_analyse_

Closes #1019 and #1500

### DIFF
--- a/openprescribing/frontend/tests/functional/test_charts.py
+++ b/openprescribing/frontend/tests/functional/test_charts.py
@@ -65,3 +65,40 @@ class SmallListTest(SeleniumTestCase):
         xlabels = self.find_by_xpath(
             "//*[contains(@class, 'highcharts-xaxis-labels')]")
         self.assertNotIn('GREEN', xlabels.text)
+
+
+class AnalyseSummaryTotalsTest(SeleniumTestCase):
+
+    def test_summary_totals_on_analyse_page(self):
+        self.browser.get(
+            self.live_server_url +
+            '/analyse/#org=CCG&numIds=0212000AA')
+        expected = {
+            'panel-heading': (
+                u'Total prescribing for Rosuvastatin Calcium across all '
+                u'CCGs in NHS England'
+            ),
+            'js-selected-month': u"Sep '16",
+            'js-financial-year-range': u"Apr—Sep '16",
+            'js-year-range': u"Oct '15—Sep '16",
+            'js-cost-month-total': u'29,720',
+            'js-cost-financial-year-total': u'178,726',
+            'js-cost-year-total': u'379,182',
+            'js-items-month-total': u'1,669',
+            'js-items-financial-year-total': u'9,836',
+            'js-items-year-total': u'20,622',
+        }
+        for classname, value in expected.items():
+            selector = (
+                '//*[@id="{id}"]'
+                '//*[contains(concat(" ", @class, " "), " {classname} ")]'.format(
+                    id='js-summary-totals',
+                    classname=classname
+                )
+            )
+            element = self.find_by_xpath(selector)
+            self.assertTrue(
+                element.is_displayed(),
+                '.{} is not visible'.format(classname)
+            )
+            self.assertEqual(element.text.strip(), value)
--- a/openprescribing/media/js/src/analyse-chart.js
+++ b/openprescribing/media/js/src/analyse-chart.js
@@ -43,6 +43,7 @@ var analyseChart = {
     rowCount: ('#data-rows-count'),
     alertForm: ('#alert-form'),
     outliersToggle: ('.outliers-toggle'),
+    summaryTotals: $('#js-summary-totals'),
   },
 
   renderChart: function(globalOptions) {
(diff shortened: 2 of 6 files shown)


evansd:
@brianmackenna Your opinions on this would be useful

brianmackenna:
It is beautiful.

evansd:
Thanks @brianmackenna :smiley: 

Technical aside for Peter and Seb: I can't remember if I've mentioned this before but a habit I picked up somewhere is to never use the same class/id simultaneously for behaviour and styling and to prefix all behavioural classes/ids with `js-`. (By "behavioural" I just mean selectors which JS uses to locate elements.) It makes it easy to tell at a glance that a particular bit of HTML is connected to some JS somewhere, and the prefix makes it easily greppable.

brianmackenna:
So I was reflecting on this and it meets the development need as it came from users at national organisations. However this would also be beneficial for users are at practices/CCGs.

Could we add another total of "for the organisations you have selected the total is x"

If this is time consuming or difficult we should ship as is and add as future development.

evansd:
@brianmackenna That's a good suggestion, and I don't think it would be massively difficult to do, but it's enough extra work that I think we should make it a separate issue so it doesn't block this change going out. Would you mind creating that? Thanks

evansd:
@brianmackenna It turned out to be easier than I thought to do this for organisation totals as well, so I've bundled that in to this pull request.

openprescribing/media/js/src/analyse-chart.js
```diff
@@ -530,5 +529,110 @@ var analyseChart = {
     this.updateCharts();
   },
 
+  displayTotals: function() {
+    var selected;
```
sebbacon: Tiny point but naming this `selectedOrgs` (or indeed `orgSelection`) would aid readability; at first I assumed this was a flag.

openprescribing/media/js/src/analyse-chart.js
```diff
@@ -530,5 +529,110 @@ var analyseChart = {
...
+  // Months are zero-indexed so 3 is April
+  var financialYear = date.getMonth() >= 3 ? date.getFullYear() : date.getFullYear() - 1;
+  var yearStart = new Date(financialYear, 3, 1);
+  return Highcharts.dateFormat('%Y-%m-%d', yearStart);
+}
+
+function formatDateRange(fromDateStr, toDateStr) {
+  var fromDate = new Date(fromDateStr);
+  var toDate = new Date(toDateStr);
+  var formatStr = (fromDate.getFullYear() === toDate.getFullYear()) ? "%b" : "%b '%y";
```
sebbacon: It would be nice to have the financial YTD during the month of April be *Apr 2018* (as opposed to  *Apr - Apr 2018*)

evansd: Tiny but good! I have amended

openprescribing/media/js/src/analyse-chart.js
evansd: Tiny but good! I've changed this

evansd:
> In the absence of tests, before deployment, can we ensure this has been deployed to staging, and tested in Safari, Chrome, IE and Firefox by a non-developer?

I now managed to get the functional tests working. But I think it's worth checking over on staging as well.

openprescribing/frontend/tests/functional/test_charts.py
```diff
@@ -65,3 +65,40 @@ def test_nothing_hidden_by_default(self):
...
+            'js-cost-financial-year-total': u'178,726',
+            'js-cost-year-total': u'379,182',
+            'js-items-month-total': u'1,669',
+            'js-items-financial-year-total': u'9,836',
+            'js-items-year-total': u'20,622',
+        }
+        for classname, value in expected.items():
+            selector = (
+                '//*[@id="{id}"]'
+                '//*[contains(concat(" ", @class, " "), " {classname} ")]'.format(
```
inglesp: Do we have to split long strings like this?  I find it makes the code harder to read.

sebbacon: (I think I prefer it to a long line... though would be interesting to have the debate!)

evansd: I'm with Seb (unsurprisingly, given that I wrote it!). Particularly given that it's broken on a logical boundary. I don't know if xpath has an equivalent to regex verbose mode where whitespace in the expression is ignored but if so I'd be tempted to use that and make the whole thing a triple-quoted string.

evansd:
@inglesp @sebbacon I've just added two small commits that it would be worth you glancing at before we merge.

openprescribing/frontend/tests/functional/test_charts.py
inglesp: OK!  I think my main problem is that I'm not used to it.

inglesp:
Both new commits look sensible.

Issue #1464, opened 2019-02-13
Update the About page


HelenCEBM:
It needs 
- [x] citation updating to 2019
- [x] new funders adding
- [x] new data sources adding, e.g. tariff (@richiecroker), long term trends, others..
- [ ] mention some of the new OP uses that have come up recently

PR #1506, opened 2019-02-25
Update About page

brianmackenna:
Partial updates as requested in #1464 

- citation year
- where we get Rx data from

### DIFF
--- a/openprescribing/templates/about.html
+++ b/openprescribing/templates/about.html
@@ -23,7 +23,7 @@
 
 <h3>Citation</h3>
 
-<p>You are welcome to use data or graphs from this site in your academic output with attribution. Our methods paper will be published shortly, until then please cite "OpenPrescribing.net, EBM DataLab, University of Oxford, 2017" as the source for academic attribution.</p>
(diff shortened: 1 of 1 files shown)


openprescribing/templates/about.html
```diff
@@ -37,7 +37,7 @@ <h3>Code and contributions</h3>
 
 <h3 id="sources">Data sources</h3>
 
-<p><strong>Prescribing data</strong> is from the monthly files published by <a href="https://digital.nhs.uk/practice-level-prescribing-summary">NHS Digital</a>, used under the terms of the Open Government Licence.</p>
+<p><strong>Prescribing data</strong> is from the monthly files published by <a href="https://www.nhsbsa.nhs.uk/information-services-portal-isp">NHS Business Service Authority</a>, used under the terms of the Open Government Licence.</p>
```
inglesp: Should it be "by NHS Business Service Authority" or "by the NHS Business Service Authority"?

brianmackenna: Are you sure you aren't a pharmacist?!

"by the NHS Business Service Authority"

Can you push it back and I'll change it?

inglesp: > Are you sure you aren't a pharmacist?!

High praise!

You should be able to make this change yourself, without me doing anything.  Let me know if you're not sure how to do this.

PR #1510, opened 2019-02-26
Use hashed names to reference static assets

### LINKED ISSUE #1491: Use hashed names to reference our static assets
The specific prompt for this is seeing some JS errors which we thought we had fixed and can't reproduce, and the worry that these are coming from stale versions of the JS cached in people's browsers.

Django provides a mechanism (ManifestCachedStaticFilesStorage) for including the hash of each static file in its name so that every version is unique. We should use this.

evansd:
This patch introduces Django's [ManifestStaticFilesStorage](https://docs.djangoproject.com/en/1.11/ref/contrib/staticfiles/#manifeststaticfilesstorage) backend which will create unique names for every version of our static files so we no longer have issues with stale files being cached.

This involves making sure that every reference to the URL of a static file uses the `{% static %}` tag so that the correct URL gets used in production.

Currently, this doesn't use the backend in test which is not ideal but is what the Django docs recommend. At some point we could start using WhiteNoise which would allow us to have an identical configuration in dev, test and production.

Closes #1491

### DIFF
--- a/openprescribing/frontend/templatetags/template_extras.py
+++ b/openprescribing/frontend/templatetags/template_extras.py
@@ -4,6 +4,7 @@ import math
 from django import template
 from django.utils.safestring import mark_safe
 from django.contrib.humanize.templatetags.humanize import intcomma
+from django.contrib.staticfiles.storage import staticfiles_storage
 from django.utils import timezone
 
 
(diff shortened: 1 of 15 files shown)


evansd:
Currently deployed on [staging](http://staging.openprescribing.net) and looks OK.

evansd:
@inglesp @seb This is finally working now. Sorry, I shouldn't have requested a review until I got the tests passing.

sebbacon:
I should rather say "delighted" to approved this - makes me very happy!

PR #1516, opened 2019-02-27
Handle pack_size with non-ascii character

inglesp:
We've seen a pack_size containing \x0a


evansd:
Would importing `unicode_literals` be dangerous here? From my casual scan of the file I think not, and it would make the Python 3 switch easier.

PR #1518, opened 2019-02-27
Support totals over CCGs with practice level data

evansd:
The Analyse form allows you to select a CCG in the practice field as a
shorthand for "all the practices in this CCG". Previously we didn't
support this when calculating totals (we'd just hide the table) but this
created an annoying inconsistency in the interface.

Note that in the case where we have practice level data but no practices
or CCGs selected we will now throw an error which nothing catches.
However the interface is supposed to prohibit this configuration so if
we do get here then throwing an error (which Sentry will alert us to) is
probably the right thing.

Additionally, if we start including other organisations like STPs then
this logic will need extending. So we loudly fail on unexpected
organisation types rather than silently give wrong results.

### DIFF
--- a/openprescribing/media/js/src/analyse-chart.js
+++ b/openprescribing/media/js/src/analyse-chart.js
@@ -530,17 +530,7 @@ var analyseChart = {
   },
 
   displayTotals: function() {
-    var selectedOrgs;
-    try {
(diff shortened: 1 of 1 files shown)


evansd:
Yeah, I agree it would be nice to extend the test suite but it's just a matter of time and priorities. I'll get the Bugfinder General (aka Helen) to poke at it on staging first.

PR #1529, opened 2019-03-01
Fix n+1 query in measure_by_xxx endpoints

### LINKED ISSUE #1528: Poor peformance on `measure_for_all_ccgs`
When visiting https://openprescribing.net/measure/ppi/, the API endpoint `/api/1.0/measure_by_ccg/?measure=ppi&format=json` is called.

Without caching, this takes in excess of one minute to return.

Informal profiling shows that most of the time is spent in [`_roll_up_measure_values`](https://github.com/ebmdatalab/openprescribing/blob/7c8f4a1fc37709a4345b0e07b16902427a47ceeb/openprescribing/api/views_measures.py#L262), which in this case loops over approx 12,000 values at a time of 5ms per loop iteration.

Things to think about:

* It's odd that 200ish sets of measurevalues being fetched take a minute to roll up; but 40ish standard measures at `measures_for_one_ccg` take much less than 12s to load
* Note that the same problem is exhibited for large CCGs (with many practices). Example: https://openprescribing.net/api/1.0/measure_by_practice/?org=15E&measure=ppi&format=json

inglesp:
Closes #1528.

The extra queries were caused by looking up eg measure_value.pct.name in
_roll_up_measure_values.  The pct/ccg ugliness is not worth solving as
we expect to replace much of this code when the MatrixStore arrives.

### DIFF
--- a/openprescribing/frontend/managers.py
+++ b/openprescribing/frontend/managers.py
@@ -89,6 +89,9 @@ class MeasureValueQuerySet(models.QuerySet):
 
         if parent_org_type == 'ccg':
             parent_org_type = 'pct'
+            qs = qs.select_related('pct')
+        else:
+            qs = qs.select_related(org_type)
 
         return qs.for_orgs(parent_org_type, org_ids) \
             .for_measures(measure_ids) \


evansd:
This fix isn't quite right as it stands: if `org_type`, rather than `parent_org_type`, is `ccg` it tries to do a select_related on that and blows up -- hence the failing test.

But I don't think it's `parent_org_type` that we should be worrying about anyway. It's only `org_type` that's passed to the [_roll_up_measure_values](https://github.com/ebmdatalab/openprescribing/blob/81f8f4961f4c68dae6859b9d879de3c47fc2f861/openprescribing/api/views_measures.py#L262) function, and that's what determines which fields we access.

I'm also not convinced that `selected_related` is the right way of fixing this, because that's going to drag an entire Practice/PCT object along with every MeasureValue, which will mean that we load each org 60 times more than we need to (which includes, in the case of CCGs, their entire geographic boundary). I think we want [prefetch_related](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#prefetch-related) instead, which will result in one additional query but will mean we only load each organisation once.

For this reason, I'm also a bit worried about this line here, but it's hard to know what other bits of code are depending on this.
https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/frontend/managers.py#L28-L37

This bumps in to one of the worst aspects of Django's ORM. I remember someone discussing somewhere whether there should be a way of disabling lazy loading so you didn't get accidentally bitten by this kind of thing. One thing we could start doing is using [assertNumQueries](https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries) in our tests. But it's Yet Another Thing to Do when testing, and is quite a fragile approach.

sebbacon:
Rather than taking a test-oriented approach to detecting the problem, could we take a reporting-oriented approach? 

i.e., like we do with `coverage`, wrap/configure the tests with [`nplusone`](https://github.com/jmcarp/nplusone), and either count the logged errors, or use the `NPLUSONE_RAISE` setting?

evansd:
Ooh, I hadn't seen nplusone before. I was just thinking about how you could create a warning system for this kind of thing. I think it's definitely worth giving that approach a try.

evansd:
Further info, I think the problem came in this commit:
https://github.com/ebmdatalab/openprescribing/commit/892daf00b5812c6b453f68b8e23e42ddd2dfdbdf#diff-5a8d8811aa70512b6cd4a3587653fea6

Previously, the code did an unconditional `select_related('pct')` but now we only do it if `org_ids` is set, which it isn't in the case of the "all CCGs for this measure" page.

evansd:
Alternative PR here:
https://github.com/ebmdatalab/openprescribing/pull/1531

I'd rather do the `(select|prefetch)_related` at the point where we know we need it, rather than burying it in the the queryset method, as it comes with its own performance cost.

PR #1537, opened 2019-03-05
Update requirements.txt to match requirements.in

evansd:
This patch is the result of running:

    pip-compile --output-file requirements.txt requirements.in

I'm not 100% sure why this results in a diff, but I suspect it's
something to do with dependabot updating versions in
requirements.txt and requirements.in but not updating transitive
dependencies.


evansd:
@sebbacon @inglesp Can you both trying running the same command locally and check that it produces the same output?

PR #1562, opened 2019-03-12
Update items, quantity, actual cost definitions in FAQs

evansd:
The aim is for these to be slightly easier to understand by the interested layperson.

Existing text:
openprescribing net_faq_

New text:
localhost_8000_faq_

### DIFF
--- a/openprescribing/templates/faq.html
+++ b/openprescribing/templates/faq.html
@@ -94,11 +94,39 @@
 </p>
 
 <h3 id="prescquantity">What does quantity mean?</h3>
-<p>NHS Digital define quantity in their <a href="http://content.digital.nhs.uk/media/10048/FAQs-Practice-Level-Prescribingpdf/pdf/PLP_FAQs_April_2015.pdf">FAQ</a> (PDF) as:</p>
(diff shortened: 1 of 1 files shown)


HelenCEBM:
Looks good to me! Do we have an explainer on ADQs? Under the issue of adding quantities we could refer people to that if so.

evansd:
Doesn't look like we do have anything on ADQs at the moment, but we should add something once we have decent quality ADQ data in the site (I think there are still some pieces to get in place before we can do that).

evansd:
@richiecroker Would appreciate a quick once over on this just to make sure I haven't accidentally said something subtly wrong.

richiecroker:
Looks good.

A couple of points:

1.  I think we could simply the item definition further by removing "single" and "regular"
2.  Under quantity, we need to have a definition for inhalers: Where the formulation is inhalers the quantity is usually the number of inhalers.

evansd:
@richiecroker Why "_usually_ the number of inhalers"? Do we know what the exceptions are? Or do we just want to say: "... but there are occasionally inconsistencies here so exercise caution when analysing this data"

richiecroker:
inhalers are the slight nightmare.  There was one in particular (@sebbacon will remember it) which was inconsistent in it's coding.  I think I want to say "but there are occasionally inconsistencies here so exercise caution when analysing this data"

PR #1567, opened 2019-03-13
Fix bug in CCG calculations for measures querying practice_statistics

### LINKED ISSUE #1520: Bug in CCG calculations for measures querying practice_statistics
There is a bug in our measure calculation code that affects measures where the value of a numerator or denominator for a practice can be null.

This affects the CCG ratios of measures that query the practice_statistics table, which is around half of our measures.

~specifically~

* ~ktt9_antibiotics~
* ~ktt9_cephalosporins_star~
* ~tamoxifen~

[Edit: I had originally thought that only three measures were affected..]

The problem is that if the denominator for a practice is null, the **numerator** for the corresponding measure value get set to zero (as well as the denominator, which is expected).  This doesn't affect the ratio for the practice measure, since if the denominator is zero we set the ratio to NULL.

However, it does affect the ratio for the CCG measure, because this is computed as the sum of all practice numerators over the sum of all practice denominators.  The sum of all practice denominators is never zero in practice so an incorrect sum of all practice numerators will give an incorrect ratio.

The denominator for a practice can be null when the denominator is selected from the practice_statistics table, because we don't have practice statistics for closed or dormant practices.

For example:

```
SELECT practice, month, SUM(items) AS numerator
WHERE practice = 'Y02514' AND month = TIMESTAMP('2018-06-01') AND bnf_code LIKE '0501%'
GROUP BY practice, month
```

practice | month | numerator
-- | -- | --
Y02514 | 2018-06-01 00:00:00 UTC | 594

But:

```
SELECT practice_id, month, numerator, denominator, calc_value
FROM measures.practice_data_ktt9_antibiotics
WHERE practice_id = 'Y02514' AND month = DATE('2018-06-01')
```

practice_id | month | numerator | denominator | calc_value
-- | -- | -- | -- | --
Y02514 | 2018-06-01 | 0 | 0.0 | null

inglesp:
Fixes #1520.

### DIFF
--- a/openprescribing/frontend/tests/commands/test_import_measures_new.py
+++ b/openprescribing/frontend/tests/commands/test_import_measures_new.py
@@ -146,37 +146,9 @@ class ImportMeasuresTests(TestCase):
         assert seen_practice_with_no_generic_prescribing
         assert seen_practice_with_no_branded_prescribing
 
-        # In production, normalised_prescribing_standard is actually a view,
-        # but for the tests it's much easier to set it up as a normal table.
-        table = Client('hscic').get_or_create_table(
-            'normalised_prescribing_standard', schemas.PRESCRIBING_SCHEMA
-        )
-
         # Upload prescribing_rows to normalised_prescribing_standard, and
         # create a DataFrame for later verification.
-        with tempfile.NamedTemporaryFile() as f:
-            writer = csv.writer(f)
-            for row in prescribing_rows:
-                writer.writerow(row)
-            f.seek(0)
-            table.insert_rows_from_csv(f.name)
-
-            headers = [
-                'sha',
-                'regional_team_id',
-                'stp_id',
-                'ccg_id',
-                'practice_id',
-                'bnf_code',
-                'bnf_name',
-                'items',
-                'net_cost',
-                'actual_cost',
-                'quantity',
-                'month',
-            ]
-            prescriptions = pd.read_csv(f.name, names=headers)
-            prescriptions['month'] = prescriptions['month'].str[:10]
+        prescriptions = self.upload_prescribing_rows(prescribing_rows)
 
         # Do the work.
         call_command('import_measures', measure='desogestrel')
@@ -213,7 +185,7 @@ class ImportMeasuresTests(TestCase):
             stp_id__isnull=False,
             regional_team_id__isnull=False,
         )
-        self.assertEqual(mvs.count(), 625)
+        self.assertEqual(mvs.count(), Practice.objects.count())
         for mv in mvs:
             self.validate_cost_based_measure_value(mv, practices.loc[mv.practice_id])
 
@@ -228,7 +200,7 @@ class ImportMeasuresTests(TestCase):
             stp_id__isnull=False,
             regional_team_id__isnull=False,
         )
-        self.assertEqual(mvs.count(), 125)
+        self.assertEqual(mvs.count(), PCT.objects.count())
         for mv in mvs:
             self.validate_cost_based_measure_value(mv, ccgs.loc[mv.pct_id])
 
@@ -246,7 +218,7 @@ class ImportMeasuresTests(TestCase):
             stp_id__isnull=False,
             regional_team_id__isnull=True,
         )
-        self.assertEqual(mvs.count(), 25)
+        self.assertEqual(mvs.count(), STP.objects.count())
         for mv in mvs:
             self.validate_cost_based_measure_value(mv, stps.loc[mv.stp_id])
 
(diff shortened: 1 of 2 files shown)


openprescribing/frontend/tests/commands/test_import_measures_new.py
```diff
@@ -190,15 +215,246 @@ def test_import_measures(self):
...
+
+                    prescribing_rows.append(row)
+
+        # In production, normalised_prescribing_standard is actually a view,
+        # but for the tests it's much easier to set it up as a normal table.
+        table = Client('hscic').get_or_create_table(
+            'normalised_prescribing_standard', schemas.PRESCRIBING_SCHEMA
+        )
+
+        # Upload prescribing_rows to normalised_prescribing_standard, and
```
sebbacon: This is copy-and-paste from earlier, could be a function?

openprescribing/frontend/tests/commands/test_import_measures_new.py
```diff
@@ -190,15 +215,246 @@ def test_import_measures(self):
...
+            'practice_id',
+            'total_list_size',
+        ]
+        practice_statistics = pd.DataFrame(columns=columns)
+
+        for practice in Practice.objects.all():
+            for month in [7, 8]:
+                timestamp = '2018-0{}-01 00:00:00 UTC'.format(month)
+
+                if month == 8 and practice.code == 'P00000':
```
sebbacon: Comment explaining what's going on here?  Or perhaps an assertion like you do in the other test (`seen_practice_with_no_list_size = True` + `assert seen_practice_with_no_list_size`)?

openprescribing/frontend/tests/commands/test_import_measures_new.py
```diff
@@ -190,15 +215,246 @@ def test_import_measures(self):
...
+                        'practice_id': practice.code,
+                        'ccg_id': practice.ccg_id,
+                        'stp_id': practice.ccg.stp_id,
+                        'regional_team_id': practice.ccg.regional_team_id,
+                        'total_list_size': total_list_size,
+                    },
+                    ignore_index=True,
+                )
+
+        # Upload practice_statistics_rows to BigQuery.
```
sebbacon: Possibly a candidate for including in refactoring I mentioned above?

openprescribing/frontend/tests/commands/test_import_measures_new.py
```diff
@@ -190,15 +215,246 @@ def test_import_measures(self):
...
+        )
+        self.validate_practice_statistics_measure_global(mg, practices, 'practice')
+        mvs = MeasureValue.objects.filter(
+            month=month,
+            practice_id__isnull=False,
+            pct_id__isnull=False,
+            stp_id__isnull=False,
+            regional_team_id__isnull=False,
+        )
+        self.assertEqual(mvs.count(), 625)
```
sebbacon: What does this number relate to - number of practices presumably? Could it be a variable? (e.g. `Practice.objects.count()`)

openprescribing/frontend/tests/commands/test_import_measures_new.py
```diff
@@ -208,15 +464,15 @@ def test_import_measures(self):
         )
         self.assertEqual(mvs.count(), 125)
```
sebbacon: Number of regional teams?

openprescribing/frontend/tests/commands/test_import_measures_new.py
```diff
@@ -226,15 +482,15 @@ def test_import_measures(self):
         )
         self.assertEqual(mvs.count(), 25)
```
sebbacon: Number of STPs?

openprescribing/frontend/tests/commands/test_import_measures_new.py
```diff
@@ -244,9 +500,13 @@ def test_import_measures(self):
         )
         self.assertEqual(mvs.count(), 5)
```
sebbacon: Number of regions?

openprescribing/frontend/tests/commands/test_import_measures_new.py
inglesp: Good point!

openprescribing/frontend/tests/commands/test_import_measures_new.py
inglesp: Good idea.

openprescribing/frontend/tests/commands/test_import_measures_new.py
inglesp: Good idea.

inglesp:
> why is there both a `test_import_measures` and a `test_import_measures_new`?

`test_import_measures_new` tests most of what's in `test_import_measures`, but more comprehensively, so most of `test_import_measures` can be dropped.  But I'm not sure there's much point doing this now, as `import_measures` will be going away soon.

PR #1570, opened 2019-03-13
Automatically update any measures on deploy

### LINKED ISSUE #1517: Automate measures deployment
* Fab command to deploy a measure in staging or production
  * This should also remember to clear cloudflare cache
* Bot command to do same
* Logic in op bot to detect if there are measure changes in a PR, and if so, to trigger a measure build after the next successful deploy

sebbacon:
Closes #1517

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -170,9 +170,7 @@ def run_migrations():
 
 @task
 def build_measures(environment=None, measures=None):
-    env.app = environments[environment]
-    env.environment = environment
-    env.path = "/webapps/%s" % env.app
+    setup_env_from_environment(environment)
 
     with cd(env.path):
         with prefix('source .venv/bin/activate'):
@@ -195,7 +193,6 @@ def build_changed_measures():
         build_measures(environment=env.environment, measures=measures)
 
 
-@task
 def graceful_reload():
     result = sudo_script('graceful_reload.sh %s' % env.app)
     if result.failed:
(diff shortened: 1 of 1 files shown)


fabfile.py
```diff
@@ -168,6 +168,33 @@ def run_migrations():
...
+def build_measures(environment=None, measures=None):
+    env.app = environments[environment]
+    env.environment = environment
+    env.path = "/webapps/%s" % env.app
+
+    with cd(env.path):
+        with prefix('source .venv/bin/activate'):
+            run("cd openprescribing/ && "
+                "python manage.py import_measures --measure {}".format(
+                    measures))
```
inglesp: If we're going to allow this to be called via ebmbot, it'd be nice to be able to rebuild all measures.

fabfile.py
```diff
@@ -168,6 +168,33 @@ def run_migrations():
         warn("Refusing to run migrations in staging environment")
 
 
+@task
+def build_measures(environment=None, measures=None):
+    env.app = environments[environment]
+    env.environment = environment
+    env.path = "/webapps/%s" % env.app
```
inglesp: Could you extract this code to set up the environment as its own function, which you could then also call from `deploy()`?

And while I'm looking at this -- should `graceful_reload` also take an environment?

fabfile.py
sebbacon: I wasn't going to allow it to be called via ebmbot, on the grounds that you shouldn't need to rebuild any measures that haven't changed....?

fabfile.py
sebbacon: 1. Yes
2. No - I see no benefit to it being its own fab command so instead think we should remove the `@task`

fabfile.py
inglesp: I was thinking of the case when `import_measures` ran incorrectly... but this hasn't happened for a long time and in any case we could run this from the console.  So scratch this suggestion.

PR #1574, opened 2019-03-14
Measure validation

sebbacon:
Adds a management command (`./manage.py import_measures --check`) which checks that measure SQL is valid.

Includes a step during fabric build which validates changed measures before building them, with the result that someone deploying a branch to staging will get an exception reported to `#technoise`.

Validation happens using BigQuery's `dryRun` option, which will raise exceptions on malformed SQL without incurring execution costs.

Note that this PR contains accidentally-merged commits from Rich which were then reverted, so commit-by-commit review will be hard; just check *Files changed*, which is not too big.


openprescribing/frontend/management/commands/import_measures.py
```diff
@@ -31,6 +30,8 @@
 from common import utils
 from frontend.models import MeasureGlobal, MeasureValue, Measure, ImportLog
 
+import google
```
inglesp: Can you do `from google.api_core.exceptions import BadRequest` here?

openprescribing/frontend/management/commands/import_measures.py
```diff
@@ -131,16 +163,27 @@ def parse_options(self, options):
...
 def parse_measures():
     """Deserialise JSON measures definition into dict
     """
     measures = OrderedDict()
-    fpath = os.path.dirname(__file__)
-    files = glob.glob(os.path.join(fpath, "./measure_definitions/*.json"))
-    for fname in sorted(files):
+    errors = []
+
+    for fname in get_measure_definition_files():
```
inglesp: These aren't filenames, they're paths, so I think `fname` isn't a great name.

openprescribing/frontend/tests/commands/test_import_measures.py
```diff
@@ -31,15 +36,24 @@ def isclose(a, b, rel_tol=0.001, abs_tol=0.0):
...
     fpath = settings.REPO_ROOT
     fname = os.path.join(
-        fpath, ('openprescribing/frontend/tests/fixtures/'
-                'cerazette_measure.json'))
-    json_data = json.load(open(fname))
-    return {
-        'cerazette': json_data
-    }
+        fpath, ("openprescribing/frontend/tests/fixtures/measure_definitions/"
+                "{}.json".format(name)))
```
inglesp: This is, to me, really oddly formatted.  I would much rather see us break the line limit here, than do this.

openprescribing/frontend/tests/commands/test_import_measures.py
```diff
@@ -55,8 +69,8 @@ def parse_args(*opts_args):
     return cmd.parse_options(options.__dict__)
 
 
-@patch('frontend.management.commands.import_measures.parse_measures',
-       new=MagicMock(return_value=test_measures()))
+@patch('frontend.management.commands.import_measures.get_measure_definition_files',
```
inglesp: May as well use `MODULE` here and in similar places below?

openprescribing/frontend/management/commands/import_measures.py
```diff
@@ -131,16 +163,27 @@ def parse_options(self, options):
         return options
 
 
+def get_measure_definition_files():
```
inglesp: This doesn't return files, it returns paths.

openprescribing/gcutils/bigquery.py
```diff
@@ -37,6 +40,25 @@ class BigQueryExportError(Exception):
     pass
 
 
+@contextmanager
+def exception_sql_printer(sql):
```
inglesp: Good idea!

sebbacon:
You can make errors go to the originating channel (I think), but I wondered if that would be too noisy.... will give it a go.

PR #1590, opened 2019-03-20
Make API work in the presence of basic auth parameters

evansd:
This patch removes the DRF BasicAuthentication class from the
default authenticators config. We're now using basic auth to prevent
staging from being accidentally crawled and DRF was picking up the
credentials and trying to do its own authentication with them which was
breaking.


evansd:
Have confirmed that this fixes the problem on staging.

PR #1593, opened 2019-03-20
WIP: MatrixStore powered API endpoints

### LINKED ISSUE #1519: MatrixStore: second sprint
There are two high level aims here:
1. Integrate the MatrixStore build step into the import pipeline.
2. Switch out the Analyse API backend to be powered by the MatrixStore.

## 1. Integrate the MatrixStore build step into the import pipeline

- [ ] The filename argument to the `matrixstore_build` command should become optional. Omitting the filename should result in the file being created in the "appropriate" place, with that to be decided as per questions below.
- [ ] A step should be added to the import pipeline which runs `matrixstore_build`. This should be after the prescribing and practice statistics data is loaded into BigQuery, but doesn't need to wait for any data to be loaded into Postgres.

### Things to consider:

This step will add several hours to the import time. Do we want to consider consider removing the "import warning" banner, sending the Slack message etc before this step has completed? How awkward is it to do this?

Where should the SQLite file live? Initially I thought `PIPELINE_DATA_BASE_DIR` was the obvious choice, but this is on the root partition which isn't enormous. So perhaps it needs to go on one of the much larger partitions in `/mnt`.

How should the file be named? My preference based on similar past use cases would be something like:
`prescribing_<DATETIME_OF_BUILD>_<LATEST_PRESCRIBING_MONTH>_<HASH_OF_FILE_CONTENTS>.sqlite`
Including the build date gives a useful natural ordering, and having the hash is useful as part of a key for caching purposes.

## 2. Switch out the Analyse API backend to be powered by the MatrixStore

- [ ] Add some state which tells us what the currently live MatrixStore file is. Probably a new database table, but see discussion below
- [ ] Add a management command which updates this state to make a new version of the file live.
- [ ] Add a step to the import pipeline which calls this command after successfully building the new file. (In future we will want to be able to review data before pushing live, but for now it's better that it automatically stays in sync with the data in Postgres.)
 - [ ] Replace the endpoints used by the Analyse page with MatrixStore powered ones. These are
    - [ ] `spending_by_practice` which takes:
      *  an optional list of BNF code prefixes (or things which can be converted to that)
       * an optional list of practice codes (or CCG codes which get translated to practice codes)
       * an optional date
    - [ ] `spending_by_ccg` which takes:
      *  an optional list of BNF code prefixes (or things which can be converted to that)
       * an optional list of CCG codes
- [ ] Use a sample of historical requests from the log to confirm that the old and new endpoints return the same results
- [ ] Using the same sample, compare timings of old and new endpoints to confirm performance improvement
- [ ] Stress test the new endpoints using both nasty queries and high concurrency using e.g. apache bench
- [ ] Confirm that the Trends pages still work correctly as these use the same spending endpoints as the Analyse form

### Things to consider:

Both endpoints can accept an empty list of BNF prefixes and will sum over all prescribing. The interface won't allow you to do this for the numerator, but it will allow it for the denominator so we need to support it. This is currently quite an expensive query to do in the MatrixStore as it involves loading all the data which, from memory, can take in the order of 45 seconds.

One option would be to precompute global totals matrices. This feels a bit icky as part of the point of the MatrixStore was to avoid all the complexity and inflexibility of pre-computation. But summing over the next level down in the BNF hierarchy (chapter, is it?) is sufficiently fast that it may be we only need this one kludge.

We need some state which tells us which file to use for live data. There are various ways to do this, including using symlinks, but I think my preference would be to store this in Postgres. The simplest implementation would be a single row in a dedicated table which stores the filename. There would also be the option for keeping a log of when each new version was deployed.

Wherever this state ends up being we will want some mechanism for updating it. Now that we have the shiny new ebmbot, a reasonable option might be to do this via a management command and then have an ebmbot task that calls the management command. This saves us building a web UI for it.

evansd:
This uses the MatrixStore to power the backend for the following `api/1.0/` endpoints: `spending`, `spending_by_practice`, `spending_by_ccg` and `org_details`.

The only intended user-visible change here is that the restriction on mixing levels of the BNF hierarchy has been removed.

Two unintended but hopefully acceptable changes are:

 * The `org_details` endpoint no longer returns results for organisations with a zero list size. Previously, the endpoint would omit results if there was no corresponding practice statistics row, but would return zero valued rows when it found them. The matrix data structure can't distinguish zero-valued from missing entries and so we omit all zero-valued entries.

 * We now store costs in pence as integers, rather than in pounds as floats. This makes no difference to net costs which only ever had two decimal places in any case. But "actual" costs were stored to a much higher precision -- a precision which was utterly spurious given that these are estimated values. This does mean that when summing over many values a difference accumulates between the results under the old and new systems. But this difference rarely amounts to more than 0.05% of the total, and where it does the absolute difference is always less than 20p.

Closes #1519 and depends on #1651


evansd:
As a follow up task: I think that once this is merged the only thing using any of the `vw__` tables is the NCSO stuff in `frontend.views.spending_utils`. If we can rewrite this to the use the matrixstore then we can get rid of all the code relating to these tables, which is quite a lot.

evansd:
Notes on running this locally:
 * You'll need a matrixstore file. There is one on the OP server at
  `openprescribing-data/matrixstore_2018_11_v2.sqlite`.
 * Put this file (or a symlink to it) locally at: 
  `openprescribing/openprescribing/pipeline/data/matrixstore_live.sqlite`.
 * You'll need org data (all the Practice and CCG models) in Postgres.

The file above was built using the same data imported into staging so the results ought to be the same.

evansd:
The responses from the old and new APIs are identical in many cases, but not in all as my [smoke testing](https://github.com/ebmdatalab/matrixstore-smoketests) has shown.

Most of these discrepancies are due to rounding differences in `actual_cost`. We made the decision to store `actual_cost` as integers to the penny whereas BQ and Postgres store them in pounds with occasionally bizarrely specific decimals. For the most the differences round to less than a penny in any case, but for sums across larger amounts of spending they can occasionally amount to a few pounds. I'm assuming we don't need to worry about this.

The more worrying kind of difference shows up in the number of items. As far as I can tell this only happens in the `spending` endpoint which aggregates across all practices (as opposed to `spending_by_ccg` and `spending_by_practice`) and is presumably due to the types of practices which are or aren't being included, although I haven't got to the bottom of this yet.

openprescribing/matrixstore/db.py
```diff
@@ -0,0 +1,110 @@
...
+from .open import MatrixStore
+
+
+# Create a memoize decorator (i.e. a decorator which caches the return value
+# for a given set of arguments). Here `maxsize=None` means "don't apply any
+# cache eviction, just keep values for ever"
+memoize = lru_cache(maxsize=None)
+
+
+@memoize
```
sebbacon: Why are we accessing the database via a memoized (per-process) file, rather than the Django sqlite backend, which would support concurrency?

sebbacon: I've just realised the cache will be shared, so ignore that. 

Though still interested in why we'd not use Django backend.

openprescribing/matrixstore/open.py
```diff
@@ -0,0 +1,81 @@
+import os.path
+import sqlite3
+
+from .matrix_ops import zero_like
+from .serializer import serialize, deserialize
+
+
+class MatrixStore(object):
```
sebbacon: As we add new commands to SQL here, it would be good to also add a brief example to the MatrixStore README

openprescribing/matrixstore/db.py
evansd: There are few non-standard things we want to do with our SQLite connection including:
 * opening the connection with special flags because we're treating it as read-only which allows SQLite to make various optimisations;
 * setting up our custom SQL functions like `matrix_sum`;
 * deserializing the binary matrix blobs

I'm sure it possible to do all these things within Django by creating a custom database backend and defining custom field types etc, but I suspect it would be a lot more complicated than the code we currently have in `matrixstore.open`.

On top of that, I think the kinds of query we'll want to run are sufficiently non-standard that we wouldn't want to use the ORM for them anyway. And, if anything, having the ORM involved may risk a false familiarity for new team mates which is actively unhelpful.

openprescribing/matrixstore/open.py
evansd: Yes, definitely.

PR #1600, opened 2019-03-25
Create aafpercent.json

richiecroker:
recreating for new pull request


sebbacon:
I don't think initials are a good choice for the authored/checked fields - too ambiguous. Email addresses would be better.

richiecroker:
email addresses added

PR #1609, opened 2019-03-27
 Add call_management_command fab task

inglesp:


inglesp:
> Did you ditch the idea of a whitelist?

Yes.  The only way for Slack users to trigger commands is via new ebmbot handlers, so I didn't think one was necessary.

PR #1626, opened 2019-04-03
Fix DT fetcher properly

inglesp:


openprescribing/dmd/management/commands/fetch_and_import_drug_tariff.py
```diff
@@ -47,13 +47,13 @@ def handle(self, *args, **kwargs):
                 urllib.unquote(os.path.splitext(
                     os.path.basename(a.attrs['href']))[0]))
             month_name, year = words[-2:]
+
+            # We have seen the last token in `words` be "19_0".  The year is
+            # reported to us via Slack, so if we pull out some nonsense here we
+            # *should* notice.
+            year = re.match('\d+', year).group()
```
sebbacon: How about asserting it's greater than, say, the year last month, and less than next year?  Just in case we don't notice...

inglesp: Nice idea, but quite fiddly as we don't know, at this point, what the year last month was.

PR #1637, opened 2019-04-10
Refactor API tests

evansd:
This PR refactors the tests for the API endpoints in preparation for using them to test the MatrixStore.

The main change is to remove the fixture values for the `vw__` tables and replace them with queries which populate those tables from the prescribing fixtures. Previously, the data was inconsistent between the views and the other tables which made testing difficult. We also fix an inconsistency with the ImportLog dates.

As a bonus, a bit of fiddling with the `db_timeout` decorator allows us to use the standard `TestCase` in place of `TransactionTestCase` which brings the run time for the API tests from 6 minutes down to 4 seconds.


inglesp:
This makes me very happy, and I think both the extra consistency and the test speedup are worthy of some very cheap prosecco for the dev team.

The SQL in api_test_data.sql isn't quite correct though (but I'm not sure whether it matters) because it doesn't take into account a practice's current CCG.  The BQ SQL selects from `normalised_prescribing_standard`, which is a view that includes a join on the practice table, so that its `pct` column always refers to the prescription's practice's current CCG.

You have two options here:

* fix the fixtures so that practices don't change CCGs (as you've done)
* fix the SQL so that the `vw__*_by_ccg` tables only include a practice's current CCG

I'd prefer fixing the SQL, but would completely understand if you felt that it wasn't worth it!

(For reference, we test that the `vw__*_by_ccg` calculations handle a practice moving CCG using fixtures generated in https://github.com/ebmdatalab/openprescribing/commit/bb3251d568956cb18ef47deb67f70cc9e44bf271#diff-2cceee09b4ab8c8570b64222f4384d9d.)

evansd:
Ah, I see. I think I prefer fixing the SQL too. I don't think it will be loads of work (it's just an extra join, right?) and it means we don't have to worry about consistency issues in future. I'll have a look at this tomorrow.

evansd:
@inglesp Have now updated the SQL so it does a join to get the current CCG membership for each practice

sebbacon:
Splendid :+1:

PR #1651, opened 2019-04-19
Build MatrixStore during import pipeline

evansd:
This PR addresses the first half of #1519, namely integrating the MatrixStore build process into the data import pipeline. It also introduces the `MATRIX_SUM` custom SQL function as this is useful for working out the precomputed totals.

At this stage, nothing should affect the user-visible site: we're simply creating a file which nothing reads.


openprescribing/matrixstore/tests/test_read_existing_file.py
```diff
@@ -0,0 +1,105 @@
...
+            'sparse_float': self.make_matrix(sparse=True, integer=False),
+            'dense_int': self.make_matrix(sparse=False, integer=True),
+            'dense_float': self.make_matrix(sparse=False, integer=False),
+        }
+        for key, value, expected_value in self.get_values(data):
+            self.assertIsInstance(value, type(expected_value))
+            if hasattr(value, 'todense'):
+                value = value.todense()
+                expected_value = expected_value.todense()
+            self.assertEqual(value.tolist(), expected_value.tolist())
```
inglesp: This is a good test to have.  I think it'd be easier to understand if `create_fixture()` was called separately.

Also, could the fixture contain metadata about the environment it was created in?   (eg Python version, OS version, library versions)

openprescribing/matrixstore/build/generate_filename.py
```diff
@@ -0,0 +1,60 @@
...
+    which can be used as a cache key and also for de-duplication (so it's easy
+    to see if rebuilding a file has resulted in any change to the data).
+    """
+    last_modified = datetime.datetime.utcfromtimestamp(
+        os.path.getmtime(sqlite_path)
+    )
+    date_min, date_max = get_date_range_from_file(sqlite_path)
+    hash_str = hash_file(sqlite_path)
+    return 'matrixstore_{date_max}_{modified}_{hash}.sqlite'.format(
+        date_min=date_min.strftime('%Y-%m'),
```
inglesp: `date_min` is not used in the filename.

openprescribing/matrixstore/build/generate_filename.py
```diff
@@ -0,0 +1,60 @@
...
+    dates = [date for (date,) in connection.execute('SELECT date FROM date')]
+    connection.close()
+    return date_from_str(min(dates)), date_from_str(max(dates))
+
+
+def date_from_str(date_str):
+    """
+    Return `datetime.date` from a string of the from 'YYYY-MM-DD'
+    """
+    return datetime.date(*map(int, date_str.split('-')))
```
inglesp: Why not use `strptime` here?

openprescribing/matrixstore/README.md
```diff
@@ -115,6 +115,51 @@ matrix = deserialize(value)
...
+```
+
+Note that the `items` column is automatically deserialized into a matrix
+object.
+
+If you know you only want a single row, you can use the `query_one`
+function:
+```python
+items, quantity = matrixstore.query_one(
+    'SELECT items, quantity FROM presentation WHERE bnf_code =?',
```
inglesp: Replace `=?` with `= ?`?

openprescribing/matrixstore/open.py
```diff
@@ -0,0 +1,53 @@
...
+            # the connection across threads, should we want to
+            check_same_thread=False
+        )
+        return cls(connection)
+
+    def query(self, sql, params=()):
+        for row in self.connection.cursor().execute(sql, params):
+            yield convert_row_types(row)
+
+    def query_one(self, *args, **kwargs):
```
inglesp: Using `*args` and `**kwargs` seems unnecessary here, and not doing so would help with `help()`.

openprescribing/matrixstore/tests/test_open.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def test_query(self):
+        excluded_code = self.factory.presentations[0]['bnf_code']
+        results = self.matrixstore.query(
+            'SELECT bnf_code, items FROM presentation WHERE bnf_code != ?',
+            [excluded_code]
+        )
+        for bnf_code, items_matrix in results:
+            self.assertNotEqual(bnf_code, excluded_code)
+            self.assertIsInstance(items_matrix[0, 0], numbers.Number)
```
inglesp: This test wouldn't fail if `.query()` returned no rows.  Could it test that the expected number of rows are returned?

openprescribing/matrixstore/open.py
```diff
@@ -0,0 +1,53 @@
+import os.path
+import sqlite3
+
+from .serializer import deserialize
+
+
+class MatrixStore(object):
```
inglesp: Is `open` the right name for this module?

openprescribing/matrixstore/build/precalculate_totals.py
```diff
@@ -0,0 +1,73 @@
...
+logger = logging.getLogger(__name__)
+
+
+def precalculate_totals(sqlite_path):
+    if not os.path.exists(sqlite_path):
+        raise RuntimeError('No SQLite file at: {}'.format(sqlite_path))
+    connection = sqlite3.connect(sqlite_path)
+    # Disable the sqlite module's magical transaction handling features because
+    # we want to use our own transactions below
+    connection.isolation_level = None
```
inglesp: For hygiene, should we reset this afterwards?

openprescribing/matrixstore/README.md
```diff
@@ -115,17 +115,91 @@ matrix = deserialize(value)
...
+
+And query it like this:
+```python
+results = matrixstore.query(
+    'SELECT bnf_code, items FROM presentation WHERE bnf_code LIKE ?',
+    ['10%']
+)
+for items, quantity in results:
+    print(bnf_code, items)
+```
```
inglesp: When we have code samples in the README, could we include the output too?

openprescribing/openprescribing/settings/production.py
```diff
@@ -121,3 +121,5 @@
 
 # Path to import log for pipeline data
 PIPELINE_IMPORT_LOG_PATH = '/home/hello/openprescribing-data/log.json'
+
```
sebbacon: Could do with a short comment explaining it's a working directory for the import process, requires several GB of space

openprescribing/matrixstore/README.md
```diff
@@ -135,3 +135,24 @@ time we build a new file.
...
+
+When we upgrade to Python 3, various bits of this code can be simplified
+or improved:
+ * The various references to `sqlite3.Binary` can be removed. For
+   whatever reason, the sqlite3 module in Python 3 can handle the
+   buffer/bytes objects as they are.
+ * We can remove the `to_pybytes` call in `serialize` for the same
+   reason.
+ * We can use the `uri` option to `sqlite3.connect` to open the file in
+   immutable mode, which should make things a bit faster:
```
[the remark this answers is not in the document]
evansd: I think the difference only shows up when there are lots of concurrent queries, so we'd need to do some benchmarked load testing to quantify this. But your asking the question has reminded me of the other reason that `immutable` is nice: it means we can potentially accept arbitrary SQL through the web interface as it becomes impossible to modify any data through the connection. I'll amend the comment accordingly.

openprescribing/matrixstore/build/import_practice_stats.py
```diff
@@ -25,12 +25,19 @@ def import_practice_stats(sqlite_path):
     if not os.path.exists(sqlite_path):
         raise RuntimeError('No SQLite file at: {}'.format(sqlite_path))
     connection = sqlite3.connect(sqlite_path)
+    dates = [date for (date,) in connection.execute('SELECT date FROM date')]
+    practice_statistics = get_practice_statistics_for_dates(dates)
+    write_practice_stats(connection, practice_statistics)
```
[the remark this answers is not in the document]
evansd: Yes, I 100% agree. In general I try to avoid abbreviations even in function/variable names as I think it increases cognitive load when reading the code out of context. But I think sometimes `practice_statistics` just got a bit too unwieldy.

openprescribing/matrixstore/open.py
evansd: No, I don't think it is. But then at the time I couldn't think of the right name and I didn't want to get stuck so I just went with the first thing I thought of. Better suggestions are very welcome. `matrixstore is the natural one, but then `from matrixstore.matixstore import MatrixStore` isn't great. Maybe `connection` or `database_wrapper`?

openprescribing/matrixstore/build/precalculate_totals.py
evansd: We close the connection immediately afterwards so it seems a bit redundant to me. But we can do if you think it make things clearer.

openprescribing/matrixstore/open.py
sebbacon: How about `wrapper` or similar, but also import it in `__init__` so you can `from matrixstore import MatrixStore`?

I'm not normally a fan of fiddling namespaces in `__init__` but as the primary entry point for matrixstore operations it may make sense

openprescribing/matrixstore/tests/test_read_existing_file.py
```diff
@@ -0,0 +1,105 @@
...
+from matrixstore.matrix_ops import sparse_matrix, finalise_matrix
+from matrixstore.serializer import (
+    deserialize, serialize, serialize_compressed
+)
+
+
+class TestReadExistingFile(SimpleTestCase):
+    """
+    This tests that we can still correctly read serialized matrices (sparse,
+    dense, compressed and uncompressed) from an SQLite fixture created by a
```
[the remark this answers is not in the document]
evansd: :) I think I oscillate a bit, depending on who I talked to last but "es-cue-ell" dominates slightly for me. Apparently, the author uses that pronunciation but has explicitly declared that both options are acceptable.

openprescribing/matrixstore/open.py
inglesp: +1 for `connection`.

And as the `MatrixStore` class is the primary way we'll interact with the database, `from matrixstore import MatrixStore` feels sensible.

openprescribing/matrixstore/build/precalculate_totals.py
inglesp: Agree it's redundant now, but if we add code after the commit and before closing the connection, it's possible (although unlikely) we'd miss that the isolation level wasn't what we expected.

openprescribing/matrixstore/open.py
evansd: I don't actually anticipate we'll be importing this directly very much as the primary interface point for the rest of the code will be the `matrixstore.db` module which comes in the next PR. (In the same way that we don't import Django's connection wrappers but just use the managed connections.) I wanted to avoid `matrixstore.matrixstore` for the sake of confusion and inelegance more than the sake of length. So I'd rather avoid messing about with `__init__.py` unless we really need to.

I'm happy with either `wrapper` or `connection`. Seb, as technical director and keeper of the bike shed paint brushes, do you want to make the casting vote?

sebbacon: `connection`

openprescribing/matrixstore/tests/test_read_existing_file.py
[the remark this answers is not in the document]
evansd: I'm no doubt inconsistent in my usage in the comments so maybe after the next PR is merged I'll do a big find-and-replace commit which fixes all of these

openprescribing/matrixstore/README.md
evansd: This is a good suggestion, but I can't really face doing this right now.

openprescribing/matrixstore/management/commands/matrixstore_set_live.py
```diff
@@ -0,0 +1,73 @@
...
+        else:
+            target_file = get_most_recent_file(date)
+        self.stdout.write(
+            "Updating live symlink to: {}".format(target_file)
+        )
+        temp_file = get_temp_filename(symlink)
+        os.symlink(target_file, temp_file)
+        os.rename(temp_file, symlink)
+        self.stdout.write(
+            "NOTE: You will need to restart the application in order for this "
```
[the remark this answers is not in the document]
evansd: I don't know what happens generally to stdout in the pipeline, but this message was specifically for people running the command manually, who'll be most in need of the reminder. I don't think it would serve much use as a line in the pipeline logs.

openprescribing/matrixstore/README.md
```diff
@@ -135,3 +210,53 @@ time we build a new file.
...
+   self.connection = sqlite3.connect(
+       'file://{}?immutable=1&mode=ro'.format(encoded_path),
+       uri=True,
+       check_same_thread=False
+   )
+   ```
+   This should improve performance a bit when we have lots of
+   simultaneous requests, but more importantly it opens up the possiblity
+   of allowing users to make arbitrary SQL queries through the web
+   interface as it guarantee that the connection is read-only.
```
madwort: typo: `guarantees`

evansd: Thanks Tom!

evansd:
@seb @inglesp I'm not going to be around much tomorrow, but it would be great if you could take another look at this and let me know if you're happy

PR #1659, opened 2019-04-25
Update import_dmd2

inglesp:
`import_dmd2` now:

* imports data from the BNF code mapping spreadsheet
* infers BNF codes for VMPs that are missing from the spreadsheet, where possible
* sets `Presentation.dmd_name`, where possible
* produces reports of oddities in the data that we could send to the BSA


openprescribing/dmd2/management/commands/import_dmd2.py
```diff
@@ -311,3 +313,166 @@ def set_vmp_bnf_codes(self):
...
+    in the middle:
+
+    * Imodium 2mg capsules
+    * Imodium Original 2mg capsules
+
+    And there are several cases where names of two products with the same BNF
+    code have different versions of the same ratio:
+
+    * Taurolidine 5g/250ml intraperitoneal solution bottles
+    * Taurolidine 2g/100ml intraperitoneal solution bottles
```
sebbacon: This is probably one we should fix. Currently it would return "Taurolidine", right? Which is just wrong. 

As I understand it, there are two VMPPs where the pack size is a 250ml or 100ml bottle. 

We should probably normalise to some standard ratio. 

In this case the MDR name is `Taurolidine_IntraperitonealSoln100mg/5ml`

Rich has suggested going for %w/v which means g of solute/100 mL of solution so this would be `Taurolidine intraperitoneal solution 2% w/v` - but see discussion [here]( https://ebmdatalab.slack.com/archives/C31D62X5X/p1556529092090200) (specifically, getting Brian to concur!)

inglesp: > Currently it would return "Taurolidine", right?

No, because "Taurolidine" is too short.  So it won't infer a name at all.

openprescribing/dmd2/management/commands/import_dmd2.py
```diff
@@ -394,13 +414,72 @@ def set_dmd_names(self):
...
+            ON vmp.vpid = vmpp.vpid
+        WHERE vmp.bnf_code IS NOT NULL
+          AND vmp.bnf_code != vmpp.bnf_code
+        '''
+
+        with connection.cursor() as cursor:
+            cursor.execute(sql)
+            self.logs['vmpps-with-different-bnf-code-to-vmp'] = cursor.fetchall()
+
+        sql = sql.replace('v', 'a')
```
evansd: :smile:

openprescribing/dmd2/tests/fixtures/bnf_code_mapping/gen_test_snomed_mapping.py
```diff
@@ -0,0 +1,78 @@
+# This script takes a BNF code mapping spreadsheet, and generates a new
```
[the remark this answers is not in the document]
inglesp: Nice idea, but wouldn't save as much time as I'll lose by fiddling about trying to implement it.

PR #1667, opened 2019-05-01
Use dmd2 data for extended information on PPU page

inglesp:


openprescribing/frontend/models.py
```diff
@@ -448,6 +458,112 @@ def current_version(self):
...
+        if self.dmd_obj_type == VMP:
+            vmp_ids = self.dmd_obj_ids
+        elif self.dmd_obj_type in [AMP, VMPP]:
+            vmp_ids = set(obj.vmp_id for obj in self.dmd_objs)
+        elif self.dmd_obj_type == AMPP:
+            vmp_ids = set(ampp.amp.vmp_id for ampp in self.dmd_objs)
+        else:
+            return []
+
+        return VirtualProductPresStatus.objects.filter(vmp__id__in=vmp_ids).distinct()
```
evansd: Does this lookup pattern (used a few places in this PR) end up doing an unnecessary join? Would this work equivalently:
```
return VirtualProductPresStatus.objects.filter(vmp__in=vmp_ids).distinct()
```
I don't suppose it matters much here in any case as it's not like the performance difference would be noticeable.

inglesp: They produce the same SQL:

```
>>> print(VirtualProductPresStatus.objects.filter(vmp__id__in=vmp_ids).distinct().query)
SELECT DISTINCT "dmd2_virtualproductpresstatus"."cd", "dmd2_virtualproductpresstatus"."descr" FROM "dmd2_virtualproductpresstatus" INNER JOIN "dmd2_vmp" ON ("dmd2_virtualproductpresstatus"."cd" = "dmd2_vmp"."pres_statcd") WHERE "dmd2_vmp"."vpid" IN (1, 2, 3)
>>> print(VirtualProductPresStatus.objects.filter(vmp__in=vmp_ids).distinct().query)
SELECT DISTINCT "dmd2_virtualproductpresstatus"."cd", "dmd2_virtualproductpresstatus"."descr" FROM "dmd2_virtualproductpresstatus" INNER JOIN "dmd2_vmp" ON ("dmd2_virtualproductpresstatus"."cd" = "dmd2_vmp"."pres_statcd") WHERE "dmd2_vmp"."vpid" IN (1, 2, 3)
```

Have I missed something?

evansd: Ah, no. I have. Sorry, I misunderstood the relationship structure here.

openprescribing/frontend/models.py
```diff
@@ -448,6 +458,112 @@ def current_version(self):
...
+        if not self.dmd_objs:
+            return None
+        else:
+            return type(self.dmd_objs[0])
+
+    def tariff_categories(self):
+        '''Return all tariff categories for this presentation.
+
+        Tariff categories are set at VMPP level, so this doesn't make sense for
+        presentations that correspond to AMPs.
```
sebbacon: I think it's just too early in the morning for me, but why not? 

* VMPP: "500mg paracetamol tablets 28 pack" 
* VMP: "500mg paracetamol tablets"
* AMPP: "500mg paracetamol tables 28 pack (Boots)"
* AMP: "500mg paracetamol tablets (Boots)"

Per a conversation you had with Rich a while ago, didn't we decide it's incorrect to consider AMPs to be in the DT in any case?

But if they are, why aren't AMPPs too?

My sense is that DT would only apply to VMPs and VMPPs?

If my morning brain can't cope with it, it's probably a sign that greater brains with less understanding of dm+d might also get stuck here.

openprescribing/frontend/models.py
```diff
@@ -448,6 +458,112 @@ def current_version(self):
...
+
+    def tariff_categories_descr(self):
+        '''Return a description of the presentation's tariff category.'''
+        return ', '.join(tc.descr for tc in self.tariff_categories())
+
+    def availability_restrictions(self):
+        '''Return all availability restrictions for this presentation.
+
+        Availabiliy restrictions are set at AMP level, so this doesn't make
+        sense for presentations that correspond to VMPPs.
```
sebbacon: Same question as previously. If "500mg paracetamol tablets (Boots)" has an availability restriction, then I presume AMPPs  ( "500mg paracetamol tablets 28 pack (Boots)") would too, but why would "500mg paracetamol tablets" AND NOT "500mg paracetamol tablets (Boots)"?

openprescribing/frontend/models.py
```diff
@@ -448,6 +458,112 @@ def current_version(self):
                 next_version = version.replaced_by
         return version
 
+    @cached_property
+    def dmd_objs(self):
+        '''Return highest level dm+d objects with this presentation's BNF code.
```
sebbacon: I think this could do with a bit more explanation, or linking somewhere with one.

e.g. (but this is hard to write well!)  "because BNF codes are most often mapped by NHS BSA to a VMP plus several of its family members, we are usually unable to distinguish between prescribing for at a VMP level or a VMPP level in the data. Therefore we return the highest level dm+d object, which is actually the most accurate level about which we can make assertions"

openprescribing/frontend/models.py
```diff
@@ -448,6 +458,112 @@ def current_version(self):
...
+
+        filter_params = {filter_key: self.dmd_obj_ids}
+        return DtPaymentCategory.objects.filter(**filter_params).distinct()
+
+    def tariff_categories_descr(self):
+        '''Return a description of the presentation's tariff category.'''
+        return ', '.join(tc.descr for tc in self.tariff_categories())
+
+    def availability_restrictions(self):
+        '''Return all availability restrictions for this presentation.
```
sebbacon: "all availability restrictions for all members of this presentation's dm+d family"?

openprescribing/frontend/models.py
```diff
@@ -448,6 +458,112 @@ def current_version(self):
...
+            return []
+
+        filter_params = {filter_key: self.dmd_obj_ids}
+        return AvailabilityRestriction.objects.filter(**filter_params).distinct()
+
+    def availability_restrictions_descr(self):
+        '''Return a description of the presentation's availabilty restriction.
+
+        If any AMPs have "None" as their availability restriction, we
+        consider that the presentation itself has no availability restriction.
```
sebbacon: In addition, I'd argue if *all* have "None" it would be nice UI to explicitly state "None"

openprescribing/frontend/models.py
inglesp: The difference is that an AMPP belongs to a VMPP, so we can unambiguously match an AMPP against a DT category.

In contrast, an AMP belongs to a VMP, which has one or more VMPPs, which may be in different DT categories.

Does this make sense?

openprescribing/frontend/models.py
inglesp: Same answer, I think.

inglesp:
I've got the wrong approach here.

Suppose we want to find a presentation's availability restrictions, which are set at AMP level.

Given a presentation, my approach in this PR has been to find the highest level dm+d objects associated with that presentation, and then:

* if those objects are VMPs, find the associated AMPs and return their availability restrictions
* if those objects are AMPs, return their availability restrictions
* if those objects are VMPPs, return nothing (because we can't map to AMPs)
* if those objects are AMPPs, find the associated AMPs and return their availability restrictions

But this is wrong (and fiddly to implement, and difficult to explain).

Take the following presentations:

* Sucralfate_Susp 1g/5ml S/F; 0103030S0AAABAB; generic
* Antepsin_Susp 1g/5ml; 0103030S0BBABAB; branded

These both belong to [this VMP family](https://openprescribing.net/dmd/vmp/317264006/relationships/).

There are two AMPs:

* [Sucralfate 1g/5ml oral suspension sugar free (Special Order)](https://openprescribing.net/dmd/amp/30082211000001102/); 0103030S0AAABAB; generic
  * this has availability restriction "Special"
* [Antepsin 1g/5ml oral suspension (Chugai Pharma UK Ltd)](https://openprescribing.net/dmd/amp/906411000001100/); 0103030S0BBABAB; branded
  * this has availability restriction "Not available"

With the approach in this PR, we would find that the generic presentation has both "Special" and "Not available" as its availability restrictions.  This is clearly wrong.

~~~

I'm going to stop thinking about this for a couple of days.

sebbacon:
This seems to me like the kind of thing that comes to you in the shower, so I concur with leaving it for a  bit!

inglesp:
Closing in favour of #1685.

PR #1668, opened 2019-05-01
Prepare for black

inglesp:
Prepare for using black by removing various constructions that are syntax errors in Python 3.

### DIFF
--- a/openprescribing/common/generate_measure_sql.py
+++ b/openprescribing/common/generate_measure_sql.py
@@ -1,3 +1,5 @@
+from __future__ import print_function
+
 import csv
 
 reader = csv.DictReader(open('statins.csv', 'rU'))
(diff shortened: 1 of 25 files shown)


evansd:
Should we replace codeclimate with a travis job that runs `black --check` (probably as an allowed failure)?

openprescribing/prints.txt
```diff
@@ -0,0 +1,22 @@
+./common/generate_measure_sql.py
```
evansd: Did you intend to commit this file?

inglesp: I think you know the answer to that!

sebbacon:
In the past I've found codeclimate's code complexity warnings to be useful pointers to code smells - but I think we've (at least in effect?) got that all turned off anyway, because they're only useful sometimes.  So I think I'd be happy with replacing at least the python parts of codeclimate with a black test

PR #1670, opened 2019-05-02
Changes to the dm+d relationships page

inglesp:
Before:

image

After:

image


inglesp:
Merging now, as I'd like to make use of the extra links now.

As discussed on Slack, this is for a UI that we don't yet expose, so further review can come later.

Issue #1645, opened 2019-04-16
Colour changes on OP screen - Black and Purple


brianmackenna:
We received this screenshot below from a user. After receiving his browser version info (IE and Chrome) We advised him to 
a) hard refresh
b) check configuration settings (i.e. is it an accessibility setting)
c) contact his IT

The user advised it was working fine on all his other devices and his IT thought it was a local issue but they were not sure which. (I have sent a follow up)

Nothing immediate to be done but I think it happened before and @helen asked about it on slack since this report so guess there was another report to her personally.

![image](https://user-images.githubusercontent.com/40460354/56229879-da81f900-6072-11e9-926e-599388b5e964.png)

sebbacon:
It's pretty clearly the stylesheet not loading - possibly just the bootstrap one. If it's in Chrome and IE then it sounds like a network configuration issue, possibly.

The bootstrap stylesheet is loaded from https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.3.5/css/bootstrap.min.css, but I can't think how that would be related.

We could ask the user if they can load that page when they click on it.

Any thoughts @evansd?

brianmackenna:
User feedback
>Just checked it on the same laptop but at home and it works like normal (no purple thing)?! Must be something to do with set-up at work.

sebbacon:
Great - can we ask them to visit that link from work? viz https://cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.3.5/css/bootstrap.min.css  

and report back what they see. it should be a bunch of code.

brianmackenna:
Update
> The link is blocked by our firewall !!! MY head of dept has asked our IT dept to resolve the issue. She (rightly) thinks your site is too important to be restricted in anyway.  I will let you know if that sorts it out as I guess that is a solution of sorts even if it not something that can be resolved ‘your side’

sebbacon:
I think we can count this as fixed!

I would be intrigued to know what the IT dept justification is for blocking files from that link (it's widely used across the internet and will break quite a few sites) but I doubt we'll get much of an answer.

evansd:
I actually think we should self-host all our dependencies anyway. Relying on third-party hosted stuff just creates potential for mysterious problems (like this one) and prevents doing development offline. Given that we're using Cloudflare there's no performance advantage. (The other possible advantage -- that first time visitors might already have some of those dependencies in their browser cache -- is pretty negligible I think.)

[...]

PR #1680, opened 2019-05-07
Stop using third-party hosted assets, where possible

evansd:
Doing so leads to issues like #1645 (i.e. mysterious breakages for some
users) and given that we host everything via Cloudflare's CDN there's no
performance advantage.

We make an exception for the Mapbox assets, because we're dependent on
their tile server anyway for the maps to work.

We also don't bother vendoring those files only loaded for IE8 as the
site is pretty much unusable on IE8 anyway.


sebbacon:
Great drive-by fixage!

One comment / query: can we obviate the risk of referring to a file that hasn't been vendored?

I'm hand-waving as it's part of the django machinery I don't know much about, but it looks like ManifestStaticFilesStorage might help here?

evansd:
Fortunately that's handled for us automatically: when DEBUG is false the `static` template tag will blow up loudly if you reference a file that's not in the manifest.

Just re-running the tests which terminated due to Travis flakiness.

sebbacon:
It it worth running a functional test or two with `DEBUG=False` as part of our suite?

evansd:
Well the issue here (which came up when I first introduced ManifestStaticFilesStorage in #1510) is that we're not using the same storage backend in test. This is Django's recommended way of working around the fact that `collectstatic` isn't run by default during test setup, and that Django won't serve your static files for you if DEBUG is False. But I don't think it's a great solution.

It's possible to run with the same config in production and test if we start using WhiteNoise, but I didn't want to foist my pet project on you without being sure there was a need first.

sebbacon:
OK!

evansd:
Happy to merge as is? I can create a separate ticket for updating our test set up.

sebbacon:
yes!

PR #1685, opened 2019-05-09
Use dmd2 data for extended information on PPU page

inglesp:
This replaces #1667, which had incorrect logic.


sebbacon:
That's both obviously correct and easier to read than #1667. Didn't occur to be at all that this would be the right way of doing it when reviewing the old PR. It's a funny example of how it's obvious things can be with the benefit of hindsight and a good night's sleep :)

PR #1709, opened 2019-05-23
Add caching for expensive All England numbers

evansd:
At present the All England page contains no user-specific content and so
gets cached by Cloudflare. When we add an alert signup form it will stop
being cacheable, but it takes some expensive db queries to generate and
so it will become quite slow to load.

This patch adds application-level caching for a few of these expensive
queries which can be safely cached. Data is a cached in Redis with the
connection configured via a REDIS_URL environment variable. If no such
variable is present then the application will still work, it just won't
do any caching.

The application also looks for a SOURCE_COMMIT_ID environment variable
which is uses as part of the cache keys. This has the effect that every
deployment automatically invalidates the cache. Again, if this variable
isn't present then the application falls back to doing no caching.

### DIFF
--- a/openprescribing/common/utils.py
+++ b/openprescribing/common/utils.py
@@ -63,6 +63,27 @@ def get_env_setting(setting, default=None):
             raise ImproperlyConfigured(error_msg)
 
 
+def get_env_setting_bool(setting, default=None):
+    """ Get the environment setting as a boolean
+
+    Return the default, or raise an exception if none supplied
+    """
+    value = get_env_setting(setting, default=default)
+    if value is default:
+        return value
+    normalised = value.lower().strip()
+    if normalised == 'true':
+        return True
+    elif normalised == 'false':
+        return False
+    else:
+        raise ImproperlyConfigured(
+            'Value for env variable {} is not a valid boolean: {}'.format(
+                setting, value
+            )
+        )
+
+
 def under_test():
     return db.connections.databases['default']['NAME'].startswith("test_")
 
--- a/openprescribing/frontend/views/views.py
+++ b/openprescribing/frontend/views/views.py
@@ -16,6 +16,7 @@ from django.contrib.auth import SESSION_KEY
 from django.contrib.auth import authenticate
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.auth.models import User
+from django.core.cache import cache
 from django.core.urlresolvers import get_resolver
 from django.db import connection
 from django.db.models import Avg, Count, Sum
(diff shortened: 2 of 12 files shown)


openprescribing/openprescribing/settings/base.py
```diff
@@ -383,3 +383,33 @@
...
+            "OPTIONS": {
+                # This C-based parser is much faster than the default pure
+                # Python one
+                "PARSER_CLASS": "redis.connection.HiredisParser",
+            }
+        }
+    }
+else:
+    # The dummy cache backend implements all the usual methods but never
+    # actually does any caching so it's perfect for developemt when you always
```
sebbacon: tpyo

openprescribing/frontend/views/views.py
```diff
@@ -275,15 +276,48 @@ def regional_team_home_page(request, regional_team_code):
...
+    Wrapper which caches the result of calling `function` with the supplied
+    arguments.
+
+    Note that all arguments must be serializable as strings. The commit sha of
+    the code is used as part of the cache key so any new deployment will
+    automatically invalidate the cache.
+    """
+    # If we don't have a git sha for the currently executing code which we can
+    # use to cache against then we do the safe thing which is not to cache at
+    # all
```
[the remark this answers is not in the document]
evansd: When running in development, would be one case. In general I wanted to make sure we never ended up caching against an empty sha. On reflection maybe a more robust approach would be to blow up immediately if REDIS_URL is supplied but not SOURCE_COMMIT_ID

evansd:
@sebbacon  Are you happy with the updates here?
@inglesp Did my reply answer your question sufficiently?

inglesp:
> @inglesp Did my reply answer your question sufficiently?

I think a belt-and-braces check would be worth adding if you can face it.

evansd:
@inglesp I pushed another update. I think I'm happier with this solution now as it's explicit and blows up on startup if anything's misconfigured. Let me know what you think.

openprescribing/common/utils.py
```diff
@@ -63,6 +63,27 @@ def get_env_setting(setting, default=None):
             raise ImproperlyConfigured(error_msg)
 
 
+def get_env_setting_bool(setting, default=None):
```
inglesp: Does a default of `None` make sense for a boolean setting?

evansd: In this context `None` means "no default supplied so blow up if the var isn't defined", which I just copied from the existing API

inglesp: Oh yeah, of course.

PR #1717, opened 2019-05-30
Move non dmd models to frontend

inglesp:
This moves `TariffPrice` and `NCSOConcession` to the `frontend` app.

Once this is merged and the migration has been run, we can merge the next PR (incoming) which switches over to use these new models.


openprescribing/frontend/models.py
```diff
@@ -987,3 +987,37 @@ class PPUSaving(models.Model):
...
+    # 3: Category C
+    # 11: Category M
+    tariff_category = models.ForeignKey('dmd2.DtPaymentCategory')
+    price_pence = models.IntegerField()
+
+    class Meta:
+        unique_together = ('date', 'vmpp')
+
+
+class NCSOConcession(models.Model):
```
evansd: I don't want to create unnecessary delays here but I wonder if, given that you're refactoring this stuff anyway, it would make sense to split unreconciled NCSO entries into a separate model and remove the `null=True` from `vmpp` (and possibly remove the `drug` field as well). It feels like the fact that we can't automatically reconcile NCSO entries is an implementation detail that ought not to leak into the rest of the system.

I don't have a good sense of how complex of a change this is though. And I guess there's no reason we can't add the non-NULL constraint and remove the `drug` field through a migration later, so maybe it's better tackled after you've merged the new DMD stuff.

evansd: On reflection, definitely ignore this for now. But I think it's a change worth considering once the dust has settled.

inglesp: Interesting, and not something I'd considered.  How/where does this leak?

But agreed, I'm going to ignore this for now.

evansd: The NULL itself is the leak, I think. It doesn't make conceptual sense that a concession would not have an associated VMPP, until you realise that the table is doing double duty as a store of concession data and a queue of things needing human moderation.

evansd: For instance, it prevents us using, say, `max(id)` of that table as a cache key for concession data.

inglesp: OK, makes sense.

PR #1721, opened 2019-06-03
Create nimodipine.json

HelenCEBM:


HelenCEBM:
Why it matters: The only indication in the BNF for the prescription of nimodipine is in the prevention of ischaemic neurological defects following aneurysmal subarachnoid haemorrhage. The course length is 21 days but most patients will complete the course while admitted or be discharged with the remainder. Extension of this course in primary care may be due to error or miscommunication.

PR #1723, opened 2019-06-04
Only enable caching for web process

evansd:
Currently, any attempt to run the application other than via
gunicorn_start blows up because ENABLE_CACHING is set in the config file
but SOURCE_COMMIT_ID is not set (it is determined dynamically via the
gunicorn_start script).

The quick fix here is to move ENABLE_CACHING out of the config file and
into gunicorn_start. The longer term fix is to have a common entry-point
script (not gunicorn specfic) which does all the necessary environment
setup and which everything can use.


inglesp:
In parallel to this, are you going to remove `ENABLE_CACHING` from the environment file?

evansd:
I am indeed

inglesp:
Then I approve.

PR #1729, opened 2019-06-04
Show measure tag counts

### LINKED ISSUE #1445: Add number of measures in each category
E.g. like this
`Opioids (4)`

Might be nice to have but only do if trivial

inglesp:
Fixes #1445.

### DIFF
--- a/openprescribing/api/views_measures.py
+++ b/openprescribing/api/views_measures.py
@@ -6,11 +6,11 @@ from rest_framework.decorators import api_view
 from rest_framework.exceptions import APIException
 from rest_framework.response import Response
 
+from frontend.measure_tags import MEASURE_TAGS
 from frontend.models import ImportLog
 from frontend.models import Measure
 from frontend.models import MeasureGlobal
 from frontend.models import MeasureValue
-from frontend.models import MEASURE_TAGS
 
 import view_utils as utils
 
--- a/openprescribing/common/measure_tags.json
+++ b/openprescribing/common/measure_tags.json
@@ -50,16 +50,16 @@
   "lowpriority": {
     "name": "NHS England Low Priority",
     "description": [
-      "This is a list of products ",
-      "considered a low priority for NHS Funding ",
+      "This is a list of products",
+      "considered a low priority for NHS Funding",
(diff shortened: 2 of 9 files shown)


inglesp:
And screenshot:

image

inglesp:
The app can and does start -- I was able to get the screenshot! -- and the migrations do run in the tests, when specific tests are run directly.  (But I was slightly surprised that it all seemed to work...)  I haven't dug into exactly why the tests fail when specific tests are not invoked but it's not important.

I think your proposed fix is the right one.  Do you think it would be worth caching the return value of `get_measure_tags`?  And if so, is `frontend.views.views._cache` the right thing to use?  And if so, should that be moved to its own module?

evansd:
Ah yes, sorry. I didn't mean to imply that the app wouldn't start at all, just that it wouldn't start if the tables didn't exist which would prevent the migrations running when setting up a new database.

In fact, even though I was right about that, it was more by luck than judgement: it's just that some of our migrations happen to import `frontend.models` whereas I had wrongly thought that the models always got imported on startup.

I think it is worth caching, but I think just storing it in process memory would be simpler than involving Redis. (I think of it as taking a thing we used to do on startup and doing it lazily, rather than adding caching to something that we'd otherwise expect to be dynamic.)

One option would be to use the [lru_cache decorator](https://github.com/ebmdatalab/openprescribing/blob/780982312508ffbb70a6112f28bceb5136de2fb9/openprescribing/matrixstore/db.py#L21). It would also be possible to make the `MEASURE_TAGS` dict a `SimpleLazyObject` so you wouldn't need to change any of the code which accesses it, but that might be a bit too magical.

inglesp:
Don't think `SimpleLazyObject` does what we want, as it doesn't cache.  Have rolled my own -- not too much work, but let me know if you think it's overkill!

evansd:
Apologies if I'm misunderstanding, but why doesn't SimpleLazyObject do what we want? It only calls the supplied function once when it's first accessed:
```
>>> from django.utils.functional import SimpleLazyObject
>>> def init():
...     print('called')
...     return {'foo': 'bar', 'baz': 'boo'}
... 
>>> lazy = SimpleLazyObject(init)
>>> lazy['foo']
called
'bar'
>>> lazy['foo']
'bar'
>>> lazy['baz']
'boo
```

inglesp:
Oh how embarrassing.  What I think I was remembering was:

```
def lazy(func, *resultclasses):
    """
    Turn any callable into a lazy evaluated callable. result classes or types
    is required -- at least one is needed so that the automatic forcing of
    the lazy evaluation code is triggered. Results are not memoized; the
    function is evaluated on every access.
    """
```

But, honestly, gah.

PR #1730, opened 2019-06-04
Exclude neutral measures from dashboard

### LINKED ISSUE #1177: Exclude neutral measures from new dashboards
These measures shouldn't be highlighted as having greatest potential for improvement on the new dashboards, e.g. https://openprescribing.net/ccg/06T/
image

inglesp:
A neutral measure is one where there is disagreement about whether a
higher or lower value is better.  That is, it is one where low_is_good
is null.

Fixes #1177.

### DIFF
--- a/openprescribing/frontend/views/views.py
+++ b/openprescribing/frontend/views/views.py
@@ -1092,6 +1092,7 @@ def _home_page_context_for_entity(request, entity):
         MeasureValue.objects
         .filter(**mv_filter)
         .exclude(measure_id='lpzomnibus')
+        .exclude(measure__low_is_good__isnull=True)
         .values('measure_id')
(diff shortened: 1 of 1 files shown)


openprescribing/frontend/views/views.py
```diff
@@ -1092,6 +1092,7 @@ def _home_page_context_for_entity(request, entity):
         MeasureValue.objects
         .filter(**mv_filter)
         .exclude(measure_id='lpzomnibus')
+        .exclude(low_is_good__isnull=True)
```
evansd:
```python
.exclude(measure__low_is_good__isnull=True)
```

inglesp: Thanks for doing my job for me!

Issue #1553, opened 2019-03-08
All England Alerts


brianmackenna:
Linked to #1017  and #1458 

Could we have a "regular alert" on prescribing measures, cost saving opportunities etc. and other things, particularly (Low Priority Prescribing) for All England page?

I think this would be very helpful for people at national NHS organisations but may also be useful for interested members of the public, pharma, journalists etc.

brianmackenna:
These will be different from CCG/practice alerts as they are measuring change against peers. It should reasonably mirror the All England page. Features could include

- curated list of measures  national priority measures
- greatest spread/variation in CCGs
- top 5/10 spend
- where spend at a chemical level or chapter level has changed by x% compared to previous period
- total NHS spend YTD and versus this time last year

evansd:
https://docs.google.com/document/d/1qHu02waSkGQBK-BYxLjYMTsGaI5YD8epU-Htkyc5niA/edit

HelenCEBM:
Can we close this?

evansd:
Yep! We'll make a new issue for feedback/improvements

PR #1731, opened 2019-06-04
Email alerts for All England

### LINKED ISSUE #1503: Broken link in NCSO alert email
In my Islington CCG alert the link to the "[view our dashboard for NHS Islington CCG](http://email.openprescribing.net/c/eJx1kMFugzAMhp-muUSgEKDAISekbrt0h552Qk5w23TBqUjQtLdfYJO6SyVLtmPn92ePqjRV0xXMwHQHe6Eh-GU2qMgEnxlPBkOwnkIGDucYmFVSFJ2QUhaNbKoil0UpRd62VdvuKuFgvuAXapmjnkaI4EDnhJFdVdmN3blrsdZVPZadaar9fsS6AAECa1Ezp3ZSJmM4gXWDHVN-2OLkJ0_x6r75hsFXhkzIbO0_PEFNlZLNSs8WKJ_AfCIRJES6ho0oqodKSlaV4Z9KevqbOSz3tAk-TkQwoXoGxMKib2ii-kiH5L9fOWi_RH7sT--8f4zgZz_z4-uJvwVn6RI98b5_-QHQX4TN)" is broken. My browser gives  http://%22%22/

evansd:
This PR adds a sign-up form on the All England dashboard, plus a new management command `send_all_england_alerts` which does just that.

The emails themselves will no doubt be subject to many changes (they're currently just a copy of what's already on the dashboard) but they're not quite "user facing" until the point we actually decide to send them out so it might be acceptable to merge and start collecting signups before iterating on the email content.

Relates to #1553 and closes #1503 

Example alert email:
localhost_8080_

### DIFF
--- a/openprescribing/frontend/management/commands/send_monthly_alerts.py
+++ b/openprescribing/frontend/management/commands/send_monthly_alerts.py
@@ -69,7 +69,11 @@ class Command(BaseCommand):
         """
         query = (
             Q(approved=True, user__is_active=True) &
-            ~Q(user__emailmessage__tags__contains=['measures', now_month]))
+            ~Q(user__emailmessage__tags__contains=['measures', now_month]) &
+            # Only include bookmarks for either a practice or pct: when both
+            # are NULL this indicates an All England bookmark
+            (Q(practice__isnull=False) | Q(pct__isnull=False))
+        )
         if options['recipient_email'] and (
                 options['ccg'] or options['practice']):
             dummy_user = User(email=options['recipient_email'], id='dummyid')
--- a/openprescribing/frontend/views/bookmark_utils.py
+++ b/openprescribing/frontend/views/bookmark_utils.py
@@ -871,6 +871,10 @@ def make_all_england_email(bookmark, tag=None):
     msg.subject = 'Your monthly update on prescribing across NHS England'
 
     date = ImportLog.objects.latest_in_category('ppu').current_at
+
+    # This allows us to switch between calculating savings at the practice or
+    # CCG level. We use CCG at present for performance reasons but we may want
+    # to switch in future.
     entity_type = 'CCG'
     ppu_savings = _cache(_all_england_ppu_savings, entity_type, date)
     measure_savings = _cache(_all_england_measure_savings, entity_type, date)
--- a/openprescribing/frontend/views/bookmark_utils.py
+++ b/openprescribing/frontend/views/bookmark_utils.py
@@ -31,6 +31,11 @@ from frontend.models import NCSOConcessionBookmark
 from frontend.views.spending_utils import (
     ncso_spending_for_entity, ncso_spending_breakdown_for_entity,
 )
+from frontend.views.views import (
+    _cache, _first_or_none, _all_england_low_priority_total,
+    _all_england_low_priority_savings, _all_england_measure_savings,
+    _all_england_ppu_savings
+)
 
 GRAB_CMD = ('/usr/local/bin/phantomjs ' +
             settings.APPS_ROOT +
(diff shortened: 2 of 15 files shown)


openprescribing/frontend/management/commands/send_monthly_alerts.py
```diff
@@ -69,7 +69,11 @@ def get_org_bookmarks(self, now_month, **options):
         """
         query = (
             Q(approved=True, user__is_active=True) &
-            ~Q(user__emailmessage__tags__contains=['measures', now_month]))
+            ~Q(user__emailmessage__tags__contains=['measures', now_month]) &
+            # Only include bookmarks for either a practice or pct: when both
+            # are NULL this indicates an All England bookmark
```
inglesp: Why do we want to send them separately?

openprescribing/frontend/views/bookmark_utils.py
```diff
@@ -859,3 +864,45 @@ def unescape_href(text):
...
+    )
+
+    unsubscribe_path = reverse(
+        'bookmark-login',
+        kwargs={'key': bookmark.user.profile.key}
+    )
+    unsubscribe_link = settings.GRAB_HOST + unsubscribe_path
+
+    context = {
+        'entity_type': entity_type,
```
inglesp: As this is always `"CCG"`, does it need to be passed to the template?

HelenCEBM:
Looks great! 
I'd just suggest:
- where we say *on various measures* we should point out that they cover safety, efficacy and cost-effectivenss, 
- and then after *if all CCGs had prescribed as least as well as the median*, add "on the cost-saving measures": 

image

brianmackenna:
This is great and another new feature for a new user group. 

1.   Suggest Title Tweak: _OpenPrescribing All England Prescribing Dashboard: Monthly Update_
Before settling on this tweak I wrote 3 versions that had cost/financial in the title. My first reading of the update was that it is exclusively financial. Having read again and written down titles I don't think it is a good idea to use cost/financial in the title, mainly as things like Low Priority aren't just about cost. 

2. Tweak the feedback invite
_This is the first iteration of our new experimental service and we are keen for feedback. Please respond to this e-mail with thoughts and ideas for features you would like to see._

3. Re:launch discuss at DataLab day tomorrow? I'd be keen to ship and start getting feedback. I think  the type of things we would like to do will take a bit of time. Presume when matrix fully implemented, new analyse form etc. might be a better time to return to it.

4. Shall we set a date in the diary of when to return to this?

openprescribing/frontend/management/commands/send_monthly_alerts.py
evansd: I think it's certainly helpful while we're still working on these to be able to send our standard alerts without sending the national ones. But even longer term it feels "cleaner" (in some unspecified sense) to keep the code sending the different alert types separate.

openprescribing/frontend/views/bookmark_utils.py
evansd: I've added a comment to try to explain this:
```
    # This allows us to switch between calculating savings at the practice or
    # CCG level. We use CCG at present for performance reasons but we may want
    # to switch in future.
```

evansd:
@HelenCEBM Thanks. I've also applied those changes to the All England dashboard itself, from where the original text was lifted.

@brianmackenna Thanks. I've incorporated both those changes. Would definitely make sense to discuss launch date tomorrow and set a date to revisit with further improvements. You had loads of great ideas in  the Google Doc.

sebbacon:
Agreed yesterday to :shipit:

And revisit in 2 or 3 months

PR #1733, opened 2019-06-05
Change MIN_GHOST_GENERIC_DELTA to match measure definition

### LINKED ISSUE #1646: Ghost branded generics number disagreement
The figures for GBG excess costs for the measure and on the CCG front page should match, and don't:
image

image

Seb suggested "might be related to non-standard practices", but it seems to be discrepant at practice level too, e.g. https://openprescribing.net/practice/D83030/ and https://openprescribing.net/measure/ghost_generic_measure/practice/D83030/

inglesp:
Fixes #1646


inglesp:
Seb's :+1:ed the proposal on the original issue, so I'll merge this now.  (Thanks for the review!)

PR #1745, opened 2019-06-11
Add new org types to Analyse form

evansd:
This adds STP and Regional Team as options to the Analyse form. It should be relatively easy to add PCN as an option once we have those in the database.

Note that this doesn't work with the map view because we don't have geographic data for the other org types. For now we just display a warning about this on the map tab.


brianmackenna:
This is a nice neat feature. Minor suggestion

- [ ] Tweak map warning (additional bit in italics)
Please note: map view is currently only supported for practice or CCG level data. _You can select multiple CCGs so you may wish to draw your STP by selecting you member CCGs._

evansd:
Just changed this back to WIP so please don't review yet. Some of my fixture changes have broken unrelated tests.

evansd:
@sebbacon @inglesp I think this is finally ready for review

openprescribing/media/js/src/analyse-map.js
```diff
@@ -206,8 +206,8 @@ return x[1];
 
   getBoundsUrl: function(options) {
     var boundsUrl = config.apiHost + '/api/1.0/org_location/?format=json&';
-    if (options.org === 'CCG') {
-      boundsUrl += 'org_type=ccg';
+    if (options.org !== 'practice') {
+      boundsUrl += 'org_type=' + options.org.toLowerCase();
```
inglesp: Very minor: why not switch the branches so you have `if (options.org === 'practice') { ... } else { ... }`?

openprescribing/media/js/src/chart_formatters.js
```diff
@@ -37,22 +37,39 @@ var formatters = {
...
+    switch(orgType) {
+      case 'practice':
+        return 'practice';
+      case 'ccg':
+      case 'CCG':
+        return 'CCG';
+      case 'stp':
+        return 'STP';
+      case 'regional_team':
+        return 'regional team';
```
inglesp: We call it "Regional Team" in the measure pages.

openprescribing/templates/analyse.html
```diff
@@ -16,7 +16,7 @@
 
 <h1>Search GP prescribing data</h1>
 
-<p>Search 700 million rows of prescribing data, and get prescribing information by CCG or practice. You can search for any numerator over any denominator.</p>
+<p>Search 700 million rows of prescribing data, and get prescribing information by practice, CCG, STP or region. You can search for any numerator over any denominator.</p>
```
inglesp: "region" or "regional team" or "Regional Team"?

openprescribing/api/views_spending.py
```diff
@@ -513,35 +513,57 @@ def tariff(request, format=None):
...
+        # Due to the number of practices we only return data for all practices
+        # if a single date is specified
+        if not date and not org_ids:
+            return Response(
+                'Error: You must supply either a list of practice IDs or a date '
+                'parameter, e.g. date=2015-04-01',
+                status=400
+            )
+
+    # We don't (yet?) have a "proper" code field for STPs so we're using they
```
sebbacon: s/they/their/

openprescribing/api/views_spending.py
```diff
@@ -513,35 +513,57 @@ def tariff(request, format=None):
...
+    # ONS code
+    code_field = 'code' if org_type != 'stp' else 'ons_code'
+
+    if org_type == 'practice':
+        orgs = Practice.objects.all()
+    elif org_type == 'ccg':
+        orgs = PCT.objects.filter(org_type='CCG')
```
sebbacon: I note we are only ever interested in PCTs with `org_type` of CCG. I wonder if we should do a drive-by refactor here - perhaps skip importing non-CCGs in the first place? Not super important, just a thought.

openprescribing/api/views_spending.py
```diff
@@ -513,35 +513,57 @@ def tariff(request, format=None):
...
+    else:
+        return Response('Error: unrecognised org_type parameter', status=400)
+
+    # Filter and sort
+    if org_ids:
+        orgs = orgs.filter(**{code_field + '__in': org_ids})
+    orgs = orgs.order_by(code_field)
+
+    # For most orgs we just want the code and name but (again) practices get
+    # slightly different treatment
```
sebbacon: Can we explain why?

openprescribing/media/js/src/analyse-chart.js
```diff
@@ -543,10 +544,7 @@ var analyseChart = {
     for (var i = 0; i < data.length; i++) {
       entry = data[i];
       // We include an org both if its ID is selected, or if it belongs to a
-      // CCG which is itself selected.  This logic will need extending if we
-      // add other org types to the Analyse form, so we deliberately throw an
-      // 'Unhandled orgType' error elsewhere if we're given an org type we
-      // don't recognise.
+      // CCG which is itself selected
```
sebbacon: s/CCG/grouping (or org)/

brianmackenna:
At some point could we think about adding the functionality that you can select the groupings above CCG/STP in the hierarchy on the organisation highlighting button please?

For example at a practice level you can enter a CCG when you have selected to highlight practices (see [here](https://openprescribing.net/analyse/#org=practice&orgIds=08H&numIds=0212000B0&denomIds=2.12&selectedTab=summary)) 

I frequently see all CCGs in London compared on charts so it would be good to be able to select London Region at the CCG highlighting level.

sebbacon:
You know what to do @brianmackenna!

I think this would be a very worthwhile feature but we should also have a discussion about changing the UI - the current way is not intuitive and it would get worse with this addition I think.

openprescribing/media/js/src/chart_formatters.js
evansd: I think there's a distinction between title and non-title case here which previously only showed up with "practice/Practice" as it doesn't apply to CCG. I think we want to say e..g "compare with other regional teams" rather then "compare with other Regional Teams" though we use "Regional Team" as an axis label

openprescribing/templates/analyse.html
evansd: "regional team" would be the most consistent, although I felt like the list was already getting a bit long winded so I went for something shorter. Can change if it bothers you though.

openprescribing/media/js/src/analyse-map.js
evansd: Yes, fair point I was just being lazy and making the minimal change.

openprescribing/api/views_spending.py
evansd: That feels like slightly too significant a change to make as drive by. Are you _sure_ there's nowhere we care about other sorts of PCT? I'd say better tackled as part of the org refactor.

openprescribing/api/views_spending.py
evansd: I can't say anything more informative than "for compatibility with the existing API"

openprescribing/media/js/src/analyse-chart.js
evansd: No, this is specific to the practice/CCG relationship and doesn't apply more generally

evansd:
@sebbacon @inglesp Thanks both! A few new commits to review when you have a moment

PR #1758, opened 2019-06-18
Improve dm+d search

inglesp:
* Allow searching by object type
* Allow excluding invalid/unavailble/missing BNF code
* Only show 10 results per object type if multiple object types returned
* Show invalid/unavailable/missing BNF code in results
* Add minimum length for search string

### DIFF
--- a/openprescribing/dmd2/forms.py
+++ b/openprescribing/dmd2/forms.py
@@ -1,0 +1,56 @@
+from django import forms
+
+from crispy_forms.helper import FormHelper
+from crispy_forms.layout import Layout, Field, ButtonHolder, Submit
+from crispy_forms.bootstrap import InlineCheckboxes
+
+
+obj_types_choices = [
+    ('vtm', 'VTMs'),
+    ('vmp', 'VMPs'),
+    ('amp', 'AMPs'),
+    ('vmpp', 'VMPPs'),
+    ('ampp', 'AMPPs'),
+]
+
+include_choices = [
+    ('unavailable', 'Unavailable items'),
+    ('invalid', 'Invalid items'),
+    ('no_bnf_code', 'Items with no BNF code'),
+]
+
+class SearchForm(forms.Form):
+    q = forms.CharField(
+        label='Query or SNOMED code',
+        min_length=3,
+    )
+    obj_types = forms.MultipleChoiceField(
+        label='Search...',
+        choices=obj_types_choices,
+        required=False,
+        initial=[tpl[0] for tpl in obj_types_choices],
+    )
+    include = forms.MultipleChoiceField(
+        label='Include...',
+        choices=include_choices,
+        required=False,
+    )
+
+    # This is only used in tests
+    max_results_per_obj_type = forms.IntegerField(
+        required=False,
+        widget=forms.HiddenInput(),
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(SearchForm, self).__init__(*args, **kwargs)
+        self.helper = FormHelper()
+        self.helper.form_method = 'GET'
+        self.helper.layout = Layout(
+            Field('q'),
+            InlineCheckboxes('obj_types'),
+            InlineCheckboxes('include'),
+            ButtonHolder(
+                Submit('submit', 'Search'),
+            )
+        )
--- a/openprescribing/dmd2/forms.py
+++ b/openprescribing/dmd2/forms.py
@@ -34,6 +34,7 @@ class SearchForm(forms.Form):
         label='Include...',
         choices=include_choices,
         required=False,
+        help_text='Unavailable items are not available to be prescribed and/or have been discontinued.',
     )
 
     # This is only used in tests
--- a/openprescribing/dmd2/models.py
+++ b/openprescribing/dmd2/models.py
@@ -46,6 +46,12 @@ class VTM(models.Model):
     def title(self):
         return self.nm
 
+    def status(self):
+        if self.invalid:
+            return 'invalid'
+        else:
+            return None
+
 
 class VMP(models.Model):
     class Meta:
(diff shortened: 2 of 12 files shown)


inglesp:
Screenshot:

image

openprescribing/dmd2/forms.py
```diff
@@ -0,0 +1,56 @@
...
+
+obj_types_choices = [
+    ('vtm', 'VTMs'),
+    ('vmp', 'VMPs'),
+    ('amp', 'AMPs'),
+    ('vmpp', 'VMPPs'),
+    ('ampp', 'AMPPs'),
+]
+
+include_choices = [
```
sebbacon: It's a shame to have to repeat opaque strings like `invalid` throughout the code, but it's tricky to turn these strings into constants elegantly without repetition.  Could this be a case for... metaclasses?

I'm thinking:

```python
class ChoicesType(type):
    def __getattr__(cls, key):
        if key in [x[0] for x in cls.choices]:
            return key
        else:
            raise AttributeError(key)


class DmdLevelChoice(metaclass=ChoicesType):
    choices = [
        ("vtm", "VTMs"),
        ("vmp", "VMPs"),
        ("amp", "AMPs"),
        ("vmpp", "VMPPs"),
        ("ampp", "AMPPs"),
    ]


>>>  print(DmdLevelChoice.vtm)
vtm
>> print(LevelChoice.choices)
[('vtm', 'VTMs'), ('vmp', 'VMPs'), ('amp', 'AMPs'), ('vmpp', 'VMPPs'), ('ampp', 'AMPPs')]
```

openprescribing/dmd2/forms.py
```diff
@@ -0,0 +1,56 @@
...
+        label='Search...',
+        choices=obj_types_choices,
+        required=False,
+        initial=[tpl[0] for tpl in obj_types_choices],
+    )
+    include = forms.MultipleChoiceField(
+        label='Include...',
+        choices=include_choices,
+        required=False,
+    )
```
sebbacon: I feel these fields need a bit of explanation, though I can see that's hard to do automagically with crispy forms.  Perhaps with a side bar or an asterisked bit of text?

In particular, we need to explain that "not available" [excludes](https://github.com/ebmdatalab/openprescribing/blob/7f30479e92a9b8039b941c81d85c583444a2db9f/openprescribing/dmd2/managers.py) items marked in dm+d as "Actual Products Not Available" and "invalid" means... um... I'm not sure...

Looking at the [spec](https://www.nhsbsa.nhs.uk/sites/default/files/2017-02/Technical_Specification_of_data_files_R2_v3.1_May_2015.pdf) I think `invalid` is reserved for things where there have been mistakes, or similar. Perhaps we should never show these things. For example, p43 of the [implementation guide](https://www.nhsbsa.nhs.uk/sites/default/files/2017-02/dmd_Implemention_Guide_%28Primary_Care%29_v1.0.pdf) says you should not include invalid items in the DMDProducts table

Or perhaps we just need an explanation "`invalid` is anything marked as such in the dm+d, such as errors that have been corrected"

openprescribing/dmd2/models.py
```diff
@@ -46,6 +46,12 @@ def __str__(self):
     def title(self):
         return self.nm
 
+    def status(self):
```
sebbacon: Can we have a docstring here (and all the other `status` implementations here) explaining that this should return a comma-separated list of ways this item could be filtered out?

Feels like there's some way of refactoring this nicely but I can't think what that is.  One status method called by all the models with switches in it? (yuck) Some way of defining them on the queryset managers?

openprescribing/dmd2/views.py
```diff
@@ -229,21 +235,45 @@ def bnf_code_relationships_view(request, bnf_code):
...
-    ctx = {"q": q}
+        if form.is_valid():
+            search_params = form.cleaned_data
+            max_results_per_obj_type = search_params.pop('max_results_per_obj_type') or 10
+            results = search(**search_params)
+            _annotate_search_results(results, search_params, max_results_per_obj_type)
 
-    if q:
-        results = search(q)
+            if len(results) == 1:
```
sebbacon: Can we have a comment here?  We could avoid a few moments of brow-furrowing in the future :)

openprescribing/dmd2/views.py
```diff
@@ -229,21 +235,45 @@ def bnf_code_relationships_view(request, bnf_code):
...
     return render(request, "dmd/search.html", ctx)
+
+
+def _annotate_search_results(results, search_params, max_results_per_obj_type):
+    for result in results:
+        result['obj_type_human_plural'] = result['cls']._meta.verbose_name_plural
+        result['num_hits'] = len(result['objs'])
+
+        if len(results) > 1:
+            if len(result['objs']) > max_results_per_obj_type:
```
sebbacon: This could do with a brief comment explaining it truncates the results and adds a list to the full set

openprescribing/dmd2/search.py
```diff
@@ -1,37 +1,54 @@
...
+            continue
+
+        qs = cls.objects
+        if 'invalid' not in include:
+            qs = qs.valid()
+        if 'unavailable' not in include:
+            qs = qs.available()
+        if 'no_bnf_code' not in include:
+            qs = qs.with_bnf_code()
+        qs = qs.search(q)
```
sebbacon: It's pleasing how it all comes together here!

openprescribing/dmd2/forms.py
inglesp: > "invalid" means... um... I'm not sure...

The tech spec just says "If set to 1 indicates this is an invalid entry in file."  It should be clear to a switched-on user that this is filtering by a value in our database, so I don't actually think it needs an explanation -- we don't provide explanations of other fields.

As for unavailable items, what about "Unavailable items are not available to be prescribed and/or have been discontinued", which covers the different ways that an object can be unavailable.

openprescribing/dmd2/forms.py
inglesp: Nice idea, but I don't think it's worth the extra mental overhead.

openprescribing/dmd2/models.py
inglesp: I can't think of a nice way of refactoring this, so I'll leave it as is.

inglesp:
@sebbacon how's this for help text?

image

openprescribing/dmd2/forms.py
sebbacon: I'm still not convinced about `invalid`, but I guess we can wait to see if we get any queries and deal with it then...

PR #1784, opened 2019-06-26
Fix triple counting of All England measure numbers

evansd:
Some of the All England totals are generated by aggregating CCG level data from
the MeasureValues table. It's not possible to select CCG level data by
selecting only rows where `pct_id` is not NULL because practice level
data also has the `pct_id` set. So we have to instead select where
`practice_id` is NULL.

Previously, including the extra condition that `pct_id` be not NULL was
redundant as this was true for all rows. However with the inclusion of
STP and Regional Team data this stopped being true and our query was now
returning CCG, STP and Regional Team data which resulting in a triple
counting of some of the All England numbers.

### DIFF
--- a/openprescribing/frontend/views/views.py
+++ b/openprescribing/frontend/views/views.py
@@ -1328,7 +1328,13 @@ def all_england_ppu_savings(entity_type, date):
 def all_england_measure_savings(entity_type, date):
     return (
         MeasureValue.objects
-        .filter(month=date, practice_id__isnull=(entity_type == 'CCG'))
+        .filter(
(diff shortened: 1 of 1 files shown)


inglesp:
I should have thought about this more when I added STPs and Regional Teams to the measures.

evansd:
Oh yes, the manager would be a great place for this. Something like `filter_by_org_type`?

sebbacon:
[Some notes for the rest of the team](https://gist.github.com/sebbacon/adb42259888bfb12d8a631aa91459392)

Issue #528, opened 2017-07-31
Practices not geographically in their CCGs


sebbacon:
It's hard to find a clear definition in legislation of what exactly CCG boundaries are about. The clearest definition is [from the BMA](https://www.bma.org.uk/advice/employment/commissioning/commissioning-gps-guidance), which says:

> A CCG's geographical footprint must include all practices in an area and not contain 'gaps' within the boundaries. Once CCG boundaries are defined and finalised, the CCG or CCGs a practice is eligible to be a member of will be dictated by geography.
>
> While it would be feasible for a practice on the boundary of a CCG to change membership to a bordering CCG (thus retaining coherent geographical boundaries), practices in the centre of a CCG's locality would not have this option

However, we've found a large number of practices (particularly around Birmingham and Bradford) which are geographically outside the CCG of which they're a member -- all those marked red on this map:

![image](https://user-images.githubusercontent.com/211271/28818600-770a6674-76a3-11e7-8382-9ade143fd3d2.png)

For example, these are both assigned to 04X but are within the boundaries of 13P:

* https://openprescribing.net/practice/M85730/
* https://openprescribing.net/practice/M85097/

lmfrench:
I've had a look at M85730 & M85097, and both used to be part of 13P - M85730 moved to 04X April 2016, and M85097 in June 2015

If you have a list of the practices not in the right boundary I can check whether they used to be

sebbacon:
So it would seem that  a practice can choose to belong to any CCG within whose boundaries it practices.

A practice has both *inner* and *outer* boundaries. The *inner* boundary covers addresses of patients who must be accepted in a practice. The *outer* boundary is an expanded area within which registered patients can move without being compelled to re-register elsewhere. (To complicate matters, any patient has the right to apply to any practice).

The (approximate, uncleaned) inner boundaries are available [here](https://data.england.nhs.uk/dataset/gp-practice-submitted-inner-catchment-area-kml-file). In urban areas these boundaries overlap considerably; and the practice boundaries are not coterminous with CCG boundaries. Here we can see that M85097 covers CrossCity (`13P`) and Solihull (`05P`) and just touches South and Central (`04X`). My interpretation of the definitions I've read lead me to believe this gives the practice the right to be a member of any of these three CCGs:

![image](https://user-images.githubusercontent.com/211271/28817510-58ffb68c-76a0-11e7-84aa-f0e97695629b.png)

[...]

PR #1786, opened 2019-06-26
Infer practice boundaries

evansd:
Boundaries are inferred from practice locations using a Voronoi
partition (i.e. assign each point in the country to its closest practice).

These boundaries have no real meaning but they're easy enough to explain and
they allow us to render plausible-looking maps for arbitrary collections of
practices.

The boundaries are clipped at the national border to stop them extending into
the sea -- or Wales -- and generally looking ridiculous.

The current output can be inspected here:
https://gist.github.com/evansd/954f3181115358c1cdbdc094553adf76


openprescribing/frontend/management/commands/infer_practice_boundaries.py
```diff
@@ -0,0 +1,70 @@
...
+        # Clip the region to the CCG boundary so that practice boundaries never
+        # extend outside their CCG
+        boundary = region.intersection(ccg.boundary)
+        # Find the practices contained within the region (there may be multiple
+        # as multiple practice codes sometimes share the same location)
+        for practice in practices:
+            # We check containment using the region rather than the clipped
+            # boundary as we sometimes have practices located just outside
+            # their CCG boundary (presumably due to lack of precision in the
+            # geo data)
```
sebbacon: You can say that again: https://github.com/ebmdatalab/openprescribing/issues/528

openprescribing/frontend/models.py
```diff
@@ -205,6 +205,7 @@ class Practice(models.Model):
     address5 = models.CharField(max_length=200, null=True, blank=True)
     postcode = models.CharField(max_length=9, null=True, blank=True)
     location = models.PointField(null=True, blank=True)
+    boundary = models.GeometryField(null=True, blank=True, srid=4326)
```
sebbacon: For consistency, should we be specifying srid for the precending PointField too?

openprescribing/frontend/management/commands/infer_practice_boundaries.py
```diff
@@ -0,0 +1,130 @@
...
+    they're located, with returned regions clipped to `clip_boundary`
+    """
+    cursor.execute(
+        'CREATE TEMPORARY TABLE regions (original GEOMETRY, clipped GEOMETRY)'
+    )
+    for region in regions:
+        clipped = region.intersection(clip_boundary)
+        # This is a workaround for the error "Relate Operation called with a
+        # LWGEOMCOLLECTION type" which happens when clipped boundaries end up
+        # including things which aren't polygons (i.e. points or lines) and we
```
sebbacon: Out of interest, how or when would clipping a voronoi partition result in a line?

PR #1787, opened 2019-06-27
Set ForeignKeys to VMPP to on_delete=DO_NOTHING

inglesp:
When importing dm+d data, we want to first delete all existing dmd
models, and then recreate them.  This is prevented on_delete=PROTECT,
which prevents the deletion in Django.  on_delete=DO_NOTHING instead
lets the database enforce the foreign key constraint, which is what we
want.


evansd:
Ah, so once again it transpires that I've misunderstood a fundamental aspect of how Django interacts with the database. This fix looks exactly right. I just wonder if `DO_NOTHING` is what we want as the default everywhere.

evansd:
Oh one more thing: with migrations, should we get in the habit of running `sqlmigrate` and including its output with the PR? Might help as a sanity check.

inglesp:
Here you go:

```
$ ./manage.py sqlmigrate frontend 0050
BEGIN;
--
-- Alter field vmpp on ncsoconcession
--
SET CONSTRAINTS "frontend_ncsoconcession_vmpp_id_56812377_fk_dmd2_vmpp_vppid" IMMEDIATE; ALTER TABLE "frontend_ncsoconcession" DROP CONSTRAINT "frontend_ncsoconcession_vmpp_id_56812377_fk_dmd2_vmpp_vppid";
ALTER TABLE "frontend_ncsoconcession" ADD CONSTRAINT "frontend_ncsoconcession_vmpp_id_56812377_fk_dmd2_vmpp_vppid" FOREIGN KEY ("vmpp_id") REFERENCES "dmd2_vmpp" ("vppid") DEFERRABLE INITIALLY DEFERRED;
--
-- Alter field tariff_category on tariffprice
--
SET CONSTRAINTS "frontend_tariffprice_tariff_category_id_f24ef49e_fk_dmd2_dtpa" IMMEDIATE; ALTER TABLE "frontend_tariffprice" DROP CONSTRAINT "frontend_tariffprice_tariff_category_id_f24ef49e_fk_dmd2_dtpa";
ALTER TABLE "frontend_tariffprice" ADD CONSTRAINT "frontend_tariffprice_tariff_category_id_f24ef49e_fk_dmd2_dtpa" FOREIGN KEY ("tariff_category_id") REFERENCES "dmd2_dtpaymentcategory" ("cd") DEFERRABLE INITIALLY DEFERRED;
--
-- Alter field vmpp on tariffprice
--
SET CONSTRAINTS "frontend_tariffprice_vmpp_id_25937ecd_fk_dmd2_vmpp_vppid" IMMEDIATE; ALTER TABLE "frontend_tariffprice" DROP CONSTRAINT "frontend_tariffprice_vmpp_id_25937ecd_fk_dmd2_vmpp_vppid";
ALTER TABLE "frontend_tariffprice" ADD CONSTRAINT "frontend_tariffprice_vmpp_id_25937ecd_fk_dmd2_vmpp_vppid" FOREIGN KEY ("vmpp_id") REFERENCES "dmd2_vmpp" ("vppid") DEFERRABLE INITIALLY DEFERRED;
COMMIT;
```

inglesp:
> I just wonder if `DO_NOTHING` is what we want as the default everywhere.

`PROTECT` has the benefit that an error gets raised immediately.

openprescribing/dmd2/tests/test_import_dmd2.py
```diff
@@ -147,6 +147,21 @@ def test_another_import(self):
...
+        vmpp = VMPP.objects.get(id=22479511000001101)
+        concession = NCSOConcession.objects.create(
+            date='2019-06-01',
+            vmpp=vmpp,
+            drug=vmpp.nm,
+            pack_size=vmpp.qtyval,
+            price_pence=123
+        )
+
+        vmpp.delete()
```
sebbacon: Comment here to explain why we're doing it? And/or assertion that concession still exists?

inglesp: Oops, that shouldn't be there.

PR #1793, opened 2019-06-27
Update import_stps to import from new spreadsheet

### LINKED ISSUE #1539: Get STP data from ODS
It's linked to from https://digital.nhs.uk/services/organisation-data-service/data-downloads/other-nhs-organisations.

It helpfully contains STP to Regional Team mappings, which would allow a page like "show measure for all STPs in this Regional Team".

inglesp:
Fixes #1539.  Was a little easier than I thought!


openprescribing/frontend/models.py
```diff
@@ -91,7 +91,11 @@ def cased_name(self):
 
 class STP(models.Model):
     ons_code = models.CharField(max_length=9, primary_key=True)
+    code = models.CharField(max_length=3, null=True)  # TODO make this the PK
```
evansd: Should we have `unique=True` here in preparation for making it the primary key?

PR #1796, opened 2019-06-28
Don't error if the STP name is None

sebbacon:
As a later comment says, the name only appears in the first row for
each set of STP members


sebbacon:
Note I hotfixed this on the production server as I was part way through deploying new maps

inglesp:
Dammit, thanks.

PR #1799, opened 2019-06-28
Quick fix for broken STP home page

evansd:


evansd:
Merging this now as no one else is around and we've just sent newsletter and are getting errors.

PR #1800, opened 2019-06-28
Seb stp hotfix

sebbacon:


sebbacon:
Self-merging for emergency fixage

PR #1801, opened 2019-06-28
[WIP]: change STP primary key from `ons_code` to `code`

sebbacon:
This is possible now we import official ODS codes


sebbacon:
Current status: was in the process of generating new fixtures, when I realised that the migration was going to be tricky, because changing a primary key means breaking all the foreign key constraints (I presume some `SET CONSTRAINTS ALL DEFERRED` trickery will solve this, but it was getting late on  a Friday evening)

evansd:
Closing for now. We'll revisit when as and when we need to do something similar for PCNs

PR #1802, opened 2019-07-01
Exclude items with zero quantity when calculating median prices

### LINKED ISSUE #1764: Importing ghost_generic_measure failed
When importing data for 2019_04, `import_measures` failed for `ghost_generic_measure` with `DataError: invalid input syntax for type json`.

This is entirely unrelated to the [previous failure](https://github.com/ebmdatalab/openprescribing/issues/571#issuecomment-503509570), even though the error is the same.

The problem is that for 2019_02, 2019_03, 2019_04, the 10th percentile for STPs and regional teams is `-Infinity`, and this cannot be serialized to JSON.

<details>
<summary>SQL</summary>

```
SELECT month, stp_10th, regtm_10th
FROM measures.global_data_ghost_generic_measure
WHERE IS_INF(stp_10th) OR IS_INF(regtm_10th)
```
</details>

There are 20 rows in `practice_data_ghost_generic_measure` with a `numerator` of `-Infinity`, and these correspond to the three problematic months.

<details>
<summary>SQL</summary>

```
SELECT *
FROM measures.practice_data_ghost_generic_measure
WHERE IS_INF(numerator)
ORDER BY month
```
</details>

While the practice ratio calculations can handle an infinite ratio (by converting to NaN), calculations for other organisation types do not handle this.  (This is presumably because for all other measures, an infinite ratio is caused by a zero denominator, and a CCG/STP/regional team's denominator is never zero because at least one practice will have a non-zero denominator.)

The numerator for this measure is [calculated](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/frontend/management/commands/measure_definitions/ghost_generic_measure.json) by summing the `possible_savings` column in `measures.vw__ghost_generic_measure`.

There are 30 rows in `measures.vw__ghost_generic_measure` where `possible_savings` is `-Infinity`.

<details>
<summary>SQL</summary>

```
SELECT *
FROM measures.vw__ghost_generic_measure
WHERE IS_INF(possible_savings)
```
</details>

And of those in the past five years, all are for presentation `1308010Z0AAABAB` (Ingenol Mebutate_Gel 500mcg/g).

<details>
<summary>SQL</summary>

```
SELECT bnf_code, COUNT(bnf_code)
FROM measures.vw__ghost_generic_measure
WHERE IS_INF(possible_savings) AND month > '2014-04-01'
GROUP BY bnf_code
```
</details>

The `possible_savings` column is [calculated](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/frontend/management/commands/measure_sql/vw__ghost_generic_measure.sql#L51) by subtracting the median PPU for a presentation in a given month from the cost of prescribing that presentation in a practice in that month (ish).

The median PPU of a presentation in a given month is [calculated](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/frontend/management/commands/measure_sql/vw__median_price_per_unit.sql) as the median value of `IEEE_DIVIDE(net_cost, quantity)`.

There are lots of prescribing records for `1308010Z0AAABAB` with zero quantity, which explains why the median of `IEEE_DIVIDE(net_cost,quantity)` is infinite.

<details>
<summary>SQL</summary>

```
SELECT bnf_code, COUNT(bnf_code)
WHERE quantity = 0 AND month > '2014-04-01'
GROUP BY bnf_code
ORDER BY COUNT(bnf_code) DESC
```
</details>

But this doesn't just affect the three months we have problems with -- 75 months are affected.

<details>
<summary>SQL</summary>

```
WITH ppu AS (
  SELECT
    month,
    IEEE_DIVIDE(net_cost, quantity) AS price_per_unit
  FROM
    hscic.normalised_prescribing_standard
  WHERE bnf_code = '1308010Z0AAABAB'
  ORDER BY month
 )
 
SELECT
  DISTINCT month,
  percentile_cont(price_per_unit, 0.5) OVER (PARTITION BY month) AS median_price_per_unit
FROM
  ppu
ORDER BY
  month
```
</details>

**Question:** in the measure data, why are only three months affected?

**Answer:**  because `1308010Z0AAABAB` has only been in the drug tariff since 2019_02.

<details>
<summary>SQL</summary>

```
SELECT date
FROM dmd.vmpp
INNER JOIN dmd.tariffprice tp ON vmpp.id = tp.vmpp
WHERE vmpp.bnf_code = '1308010Z0AAABAB'
```
</details>

**Question:** why have we only seen this now?

**Answer:** because this is the first import where we are using the new BNF code/SNOMED code mapping, and `1308010Z0AAABAB` is only present in the new mapping.  As such, we were not previously able to connect prescribing of `1308010Z0AAABAB` with pricing information in the drug tariff.

sebbacon:
Fixes #1764

### DIFF
--- a/openprescribing/frontend/management/commands/measure_sql/vw__ghost_generic_measure.sql
+++ b/openprescribing/frontend/management/commands/measure_sql/vw__ghost_generic_measure.sql
@@ -57,6 +57,10 @@ ON
   rx.month = dt.date
   AND rx.bnf_code = dt.bnf_code
 WHERE
-  -- lantanaprost quantities are broken in data
-  rx.bnf_code <> '1106000L0AAAAAA'
-  -- trivial savings / costs are discounted in the measure definition
+-- These can be prescribed fractionally, but BSA round quantity down,
+-- making quantity unreliable. See #1764
+  rx.bnf_code <> '1106000L0AAAAAA' -- latanoprost
+AND
+  rx.bnf_code <> '1308010Z0AAABAB' -- Ingenol Mebutate_Gel
+
+-- trivial savings / costs are discounted in the measure definition


inglesp:
:+1: thanks

PR #1810, opened 2019-07-02
Add PCN model and support in Analyse form

evansd:
This adds a basic model for Primary Care Networks (collections of practices, where a practice can belong to at most one PCN) and supports queries using this model in the API endpoints and Analyse form.

This PR depends on #1786 as we need practice boundaries in order to render maps for PCNs.


openprescribing/frontend/tests/fixtures/orgs.json
```diff
@@ -85,5 +85,19 @@
...
+    "name": "Transformational Sustainability"
+  }
+},
+{
+  "model": "frontend.pcn",
+  "pk": "PCN0002",
+  "fields": {
+    "name": "Sustainable Transformation"
   }
+}
```
sebbacon: You're really getting the hang of NHS lingo these days!

openprescribing/frontend/views/views.py
```diff
@@ -371,7 +371,11 @@ def analyse(request):
         # page load (see `alertForm` in `chart.js`)
         form = SearchBookmarkForm(
             initial={'email': getattr(request.user, 'email', '')})
-    return render(request, 'analyse.html', {'form': form})
+    context = {
+        'form': form,
+        'pcns_enabled': Practice.objects.filter(pcn_id__isnull=False).exists()
```
sebbacon: Or `PCN.objects.active()`?

evansd: That generates worse SQL and, given that this will be deleted anyway once we've imported PCN data, I'd rather leave it as is.

```sql
SELECT (1) AS "a" FROM "frontend_practice" WHERE "frontend_practice"."pcn_id" IS NOT NULL LIMIT 1
-- vs --
SELECT (1) AS "a" FROM "frontend_pcn" WHERE NOT ("frontend_pcn"."ons_code" IN (SELECT U0."ons_code" AS Col1 FROM "frontend_pcn" U0 LEFT OUTER JOIN "frontend_practice" U1 ON (U0."ons_code" = U1."pcn_id") WHERE U1."code" IS NULL)) LIMIT 1
```

PR #1813, opened 2019-07-03
Include PCNs in measures

evansd:
This adds PCN as a column on MeasureValue and updates the measures import to calculate PCN values.

It also extends the `practice` BigQuery table and `normalised_prescribing_standard` BigQuery view to include a `pcn` column. Further investigation reveals that this probably wasn't necessary if all we want to do is update the measures calculations, but it's probably no bad thing anyway. (Note that the production versions of these tables/views have already been modified.)

We tend to order columns relating to organisations by the organisation size. For this purpose I decided that PCNs were smaller than CCGs so as to have a consistent ordering everywhere. I've no idea if this is true.


evansd:
Tests are currently failing and I haven't yet managed to work out what's wrong.

PR #1822, opened 2019-07-09
Remove reference to email confirmation link

### LINKED ISSUE #1815: Remove confusing line about "confirming email address" in signup email
We don't require confirmation anymore and it's confusing to suggest that we do.

evansd:
Since #1397 was merged we no longer require confirmation of email
addresses, but we still send verification emails for the tedious reasons
outlined in that PR. However these emails still contain a line about
confirming your address by clicking the link which is confusing, so this
PR removes it.

Closes #1815


evansd:
@HelenCEBM FYI

PR #1824, opened 2019-07-09
Add one-measure-for-all-England view

sebbacon:


openprescribing/templates/measure_for_one_entity.html
```diff
@@ -12,20 +12,22 @@
 {% block content %}
   <h1>
     {{ measure.name }}
-    <br><small>{{ entity.name }}</small>
+    {% if entity %}<br><small>{{ entity.name }}</small>{% endif %}
```
inglesp: Rather than have two `{% if entity %}...{% endif %}` blocks, have one.  You'll duplicate part of the `<h1>` element, but the template might be clearer.

openprescribing/frontend/views/views.py
```diff
@@ -427,6 +427,30 @@ def measure_for_one_entity(request, measure, entity_code, entity_type):
...
+        "orgName": "All {}s in England".format(entity_type),
+        "aggregate": True,
+        "rollUpBy": "measure_id",
+    }
+
+    measure_options = _build_measure_options(measure_options)
+    entity_type_human = _entity_type_human(entity_type)
+    context = {
+        "entity_type": entity_type,
+        "entity_type_human": entity_type_human,
```
inglesp: Don't need to put `entity_type_human` in the context.

openprescribing/frontend/tests/functional/test_measures.py
```diff
@@ -75,6 +75,12 @@ def test_all_england(self):
             'Compare all CCGs in England on this measure',
             '/measure/core_0/'
         )
+        self._verify_link(
+            panel_element,
+            '.inner li:nth-child(2)',
+            'Break the overall score down into individual presentations',
+            '/measure/core_0/all-england/'
+        )
```
inglesp: We should have a test case for the new page.  The code that generates these new pages is quite hard to reason about (which links appear on which pages?) so these tests act as a kind of spec.  It's likely we'll need to touch this code as part of the org refactor, so a new test case would help prevent regressions there.

openprescribing/templates/measure_for_one_entity.html
```diff
@@ -10,22 +10,35 @@
...
-  <p>This measure shows how this organisation compares with its peers across NHS England. This is indicative only, and should be approached with caution. <a href='{% url 'faq' %}#measureinterpret'>Read more about measures</a>.</p>
+    <p>This measure shows how this organisation compares with its
+    peers across NHS England. This is indicative only, and should be
+    approached with caution.
+
+    <a href='{% url 'faq' %}#measureinterpret'>Read more about measures</a>.</p>
+
+  {% else %}
+
+  <h1>{{ measure.name }}</h1>
```
evansd: The `else` block and the main body of the conditional above, could both do with indenting I think.

sebbacon:
Thanks @evansd.

The use case is embedding a chart that live-updates in an online journal.

I had thought about generating it directly from code as part of the pipeline but I was leery of introducing new moving parts. An extra conditional in a template, even if we never use it again, seemed the least moving parts.

evansd:
The questions I was thinking of were:
 * What's the expected lifetime of the journal article?
 * Does the chart need to have all the usual on-hover tooltips, or can it be a static image?

If the lifetime isn't strictly limited then I worry that we're taking on an obligation to maintain that chart at roughly it's current dimensions in perpetuity.

I wonder if another option is a standalone bit of code running under cron that grabs a screenshot of the chart and sticks it in a GCS bucket and we serve it directly from there. The openprescribing codebase doesn't need to know anything about that. And when the cron job eventually breaks or we decide to delete it, then the chart just stays on its latest version forever.

openprescribing/api/views_measures.py
```diff
@@ -79,39 +79,51 @@ def measure_global(request, format=None):
...
         else:
-            assert False, 'Unexpected org: {}'.format(org)
+            assert False, 'Unexpected org: {}'.format(org_id)
+    return (org_id, org_type)
 
+
+@api_view(['GET'])
+def measure_numerators_by_org(request, format=None):
+    measure = request.query_params.get('measure', None)
+    org_id, org_type = _get_org_id_and_type_from_request(request)
```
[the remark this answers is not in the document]
sebbacon: Hm. What do you think about `org_kind`?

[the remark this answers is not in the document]
sebbacon: Just to note we already use the term `entity_type` and `entity_field` elsewhere.  Perhaps standardising on the latter.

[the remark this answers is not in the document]
sebbacon: Oh, I just realised I misunderstood your point! Ignore

sebbacon:
Fair point about maintaining the chart. In two minds about keeping the embed code in place to make screenshotting easier, as won't have to think about it for a couple of weeks...

PR #1825, opened 2019-07-09
Add PCN homepages

evansd:


evansd:
@sebbacon @inglesp 
You'll see I had some fun with the functional tests yesterday (see [batch of commits](https://github.com/ebmdatalab/openprescribing/pull/1825/commits) on Jul 15). There turned out to be a subtle bug in the test for cost savings which didn't manifest with the original fixture but did with the newly generated fixture (none of which is a fault with `generate_measure_fixtures` which remains awesome). Tracking this down was hampered by the fact that the functional tests are both slow and flaky, so I ended up losing the whole day (and some) to this.

However, I've addressed some of the slowness and flakiness and fixed the bug and added a full set of PCN tests so on the whole I think it was worth it.

openprescribing/frontend/tests/functional/test_measures.py
```diff
@@ -1243,3 +1222,26 @@ def _get_extreme_measure(mvs):
...
 
     return Measure.objects.get(id=measure_id)
+
+
+def _get_cost_savings(measure_values, rollup_by=None, target_percentile=50):
+    """
+    This duplicates the cost-saving logic in `measure_utils.js` (specifically
+    the logic found in the functions `_getSavingAndPercentilePerItem` and
+    `getPerformanceSummary`). This means that we roll-up the values by measure
+    or by org and then only include those who mean percentile has been greater
```
inglesp: s/who/whose/

openprescribing/frontend/tests/functional/test_measures.py
```diff
@@ -1243,3 +1222,26 @@ def _get_extreme_measure(mvs):
...
+    than the target percentile over the period.
+    """
```
inglesp: This comment could do with being replicated in `measure_utils.js`!

openprescribing/frontend/tests/functional/test_measures.py
```diff
@@ -23,6 +24,26 @@ class MeasuresTests(SeleniumTestCase):
 
     fixtures = ['functional-measures']
 
+    # These methods override the default behaviour of loading and flushing
+    # fixtures for each test method and instead load them once for the test
+    # class. This is safe as none of the tests involve mutating data, and
+    # increases the run speed by a factor of nearly 4 (311s to 78s).
```
inglesp: Wow.

openprescribing/templates/all_pcns.html
```diff
@@ -6,12 +6,25 @@
 
 {% block content %}
 
-<h1>All PCNs</h1>
+<h1>Find a PCN</h1>
 
+<p>Search for a Primary Care Network by name or ONS code</p>
```
inglesp: Are they ONS codes or ODS codes?

evansd: I _think_ they're ONS codes. My understanding is that we prefer ODS codes and use the field `code` for those, but where they are not currently available (as with STPs and most likely PCNs) we use ONS codes with the field `ons_code`.

sebbacon: We don't know in this case. There are neither ODS nor ONS codes available, currently. Perhaps just "...by name or code"?!

PR #1829, opened 2019-07-10
Use new service account on Travis

inglesp:
We wanted to make sure we were using a minimal set of permissions.

This service account has role "Pipeline Runner".


evansd:
Is it worth naming the file something like `google-auth.ebmdatalabtest-1ad80daf5cc1.json.enc ` so it's obvious what it is from the name? (I suggested `google-auth` rather than `google-credentials` for brevity, but maybe the full thing would be better.)

inglesp:
> Is it worth naming the file something like google-auth.ebmdatalabtest-1ad80daf5cc1.json.enc so it's obvious what it is from the name?

Yes, why not.

PR #1830, opened 2019-07-10
Fix dmd field descriptions still marked `TODO`

sebbacon:
Using terminology copied from the DM+D tech spec
https://www.nhsbsa.nhs.uk/sites/default/files/2017-02/Technical_Specification_of_data_files_R2_v3.1_May_2015.pdf

### DIFF
--- a/openprescribing/dmd2/models.py
+++ b/openprescribing/dmd2/models.py
@@ -478,7 +478,7 @@ class AMP(models.Model):
 
 class ApIng(models.Model):
     class Meta:
-        verbose_name = "TODO"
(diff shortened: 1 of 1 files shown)


openprescribing/dmd2/models.py
```diff
@@ -478,7 +478,7 @@ def status(self):
 
 class ApIng(models.Model):
     class Meta:
-        verbose_name = "TODO"
+        verbose_name = "Actual Product Excipients"
```
inglesp: Suggest just "Excipients", since it'll only be shown under AMPs.

PR #1834, opened 2019-07-11
Update README and comments about Google credentials

inglesp:


environment-sample
```diff
@@ -47,3 +43,7 @@ CF_API_EMAIL=cf_api_email
 NEWRELIC_API_KEY=newrelic_api_key
 
 NEW_RELIC_LICENSE_KEY=new_relic_license_key
+
+# The path to a file containing your credentials for accessing Google Cloud
```
sebbacon: I think this needs more explanation, e.g. 

The path to a file containing the credentials for a service account that has permissions to access to the required GCP services (currently X and Y). Leave blank in a development sandbox and use OAuth instead (via `glcoud auth application-default log`)

?

inglesp: Don't think so -- I've moved it to the `Only required on the server` section, and have added [comments to the code](https://github.com/ebmdatalab/openprescribing/blob/5bad75f1d47ed57aebad55a97990a99c6a0d9f8f/openprescribing/gcutils/bigquery.py#L66-L69) to explain what to do if `DefaultCredentialsError` is raised, which I think is enough.

PR #1836, opened 2019-07-11
Improve MatrixStore performance

evansd:
This patch improves the performance of summing a selection of matrices by a factor of around 3 to 4 through a combination of:
 * using in-place matrix addition rather than constructing a new matrix each time;
 * skipping some unnecessary checks whether deserialising matrices.

The majority of the time is now spent in the zstandard decompression function so changing the compression settings, or using a different library like LZ4 which prioritises decompression speed, is an obvious next place to look.

### DIFF
--- a/openprescribing/matrixstore/sql_functions.py
+++ b/openprescribing/matrixstore/sql_functions.py
@@ -1,5 +1,7 @@
 import sqlite3
 
+from scipy.sparse import csc_matrix, _sparsetools
+
 from .matrix_ops import zeros_like
 from .serializer import serialize, deserialize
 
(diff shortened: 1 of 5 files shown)


openprescribing/matrixstore/sql_functions.py
```diff
@@ -13,8 +15,13 @@ def step(self, value):
...
         matrix = deserialize(value)
         if self.accumulator is None:
-            self.accumulator = zeros_like(matrix)
-        self.accumulator += matrix
+            # We need this to be in Fortran (i.e. column-major) order for the
+            # fast addition path below to work
+            self.accumulator = zeros_like(matrix, order='F')
+        try:
+            fast_in_place_add(self.accumulator, matrix)
+        except FastInPlaceAddError:
```
inglesp: Would we expect to hit this?

evansd: Yes, but only rarely. Where a matrix is sufficiently populated (I think 50% is our current cutoff) we store it as a dense matrix and so this technique wouldn't apply. This is the case with the practice statistics matrices and possibly also with some very commonly prescribed presentations although I haven't checked the precise numbers on those.

PR #1877, opened 2019-07-16
Allow filtering by org type in admin bookmarks list

evansd:
This makes it easy to see who's subscribed to All England alerts.

### DIFF
--- a/openprescribing/frontend/admin.py
+++ b/openprescribing/frontend/admin.py
@@ -79,10 +79,13 @@ class OrgBookmarkResource(resources.ModelResource):
         model = OrgBookmark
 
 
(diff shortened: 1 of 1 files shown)


openprescribing/frontend/admin.py
```diff
@@ -79,11 +79,34 @@ class Meta:
         model = OrgBookmark
 
 
+class OrgBookmarkTypeFilter(admin.SimpleListFilter):
+    title = 'organisation type'
+    parameter_name = 'org_bookmark_type'
+
```
sebbacon: I had to go check Django documentation to know how `lookups` and `queryset` are related. Maybe a short comment to explain these are used together by the base class?

Issue #1807, opened 2019-07-02
Spike: report changed numbers following new code deploys


sebbacon:
To avoid gnarly hidden bugs causing unexpected surprises in the data, do a spike:

* manage.py command which can be run via slack bot and is run following deploys
* scrapes a set of urls finding numbers
* compares them with the same numbers from last time it ran
* reports on percentage change, with a link to the associated git commit 

Design goals:
* Not more than 2 or 3 days work to complete
* Doesn't introduce a big maintenance burden

PR #1878, opened 2019-07-16
Add command to check numbers on the site

inglesp:
Addresses #1807.

No tests since this is not production code.


openprescribing/pipeline/management/commands/check_numbers.py
```diff
@@ -0,0 +1,246 @@
...
+    # Wait until all AJAX requests are complete.
+    while browser.execute_script('return jQuery.active > 0'):
+        sleep(0.1)
+
+    source = browser.page_source
+
+    with open(os.path.join(log_path, name + '.html'), 'w') as f:
+        f.write(source.encode('utf8'))
+
+    doc = BeautifulSoup(source, 'html.parser')
```
evansd: Given that everything else is so nicely factored out, it might make sense to split this into "fetch and render HTML" and "parse numbers from HTML" functions.

openprescribing/frontend/management/commands/scrape_numbers.py
```diff
@@ -0,0 +1,243 @@
+# coding=utf8
```
evansd: I'm guessing this was an old name for this file and was committed by accident

openprescribing/frontend/management/commands/scrape_numbers.py
```diff
@@ -0,0 +1,243 @@
...
+
+        name = k
+        pattern = v[0][0][0]
+        keys = v[0][0][1]
+
+        # Ignore records starting with prefixes we're not interested in.
+        if any(pattern.startswith(prefix) for prefix in prefixes_to_ignore):
+            continue
+
+        # Ignore any URLs that
```
sebbacon: Typo / incomplete sentence. And it would be useful because the purpose isn't obvious without extra code-reading elsewhere, and domain knowledge about how/why all-england is special.

openprescribing/frontend/management/commands/scrape_numbers.py
```diff
@@ -0,0 +1,243 @@
...
+    doc = BeautifulSoup(source, 'html.parser')
+    body = doc.find('body')
+
+    # The SVG charts contain all sorts of things that look like numbers (eg
+    # rgb(255,255,255)) so we drop all <svg> elements.
+    for tag in body.find_all('svg'):
+        tag.decompose()
+
+    rx = re.compile(u'''
+        £[\d,\.]+           # Anything that looks like a cost
```
sebbacon: I found one instance of `&pound;` in the javascript code, too.

openprescribing/pipeline/management/commands/check_numbers.py
```diff
@@ -0,0 +1,246 @@
...
+            notify_slack('Not checking numbers: import in progress')
+            return
+
+        previous_log_path = get_previous_log_path()
+
+        timestamp = datetime.now().strftime('%Y%m%d%H%M%S')
+        log_path = os.path.join(settings.CHECK_NUMBERS_BASE_PATH, timestamp)
+        mkdir_p(log_path)
+
+        with webdriver.Firefox() as browser:
```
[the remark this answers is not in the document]
sebbacon: I also have a memory that the chromedriver is much easier to install headless, in case that helps...

openprescribing/pipeline/runner.py
```diff
@@ -442,6 +442,13 @@ def run_all(year, month, under_test=False):
             continue
         run_task(task, year, month, last_imported=last_imported)
 
+    # See check_numbers.py.
+    check_numbers_glob = os.path.join(
+            settings.CHECK_NUMBERS_BASE_PATH, '*', 'numbers.json'
+    )
+    for path in glob.glob(check_numbers_glob):
+        os.unlink(path)
```
sebbacon: Could we use `remove` rather than the unix idiom? More like English.

openprescribing/frontend/management/commands/scrape_numbers.py
```diff
@@ -0,0 +1,243 @@
...
+        ''',
+        re.VERBOSE
+    )
+    return rx.findall(body.text)
+
+
+def compare_numbers(previous_numbers, numbers):
+    '''Compare dictionaries of numbers, returning list of any differences.
+    '''
+
```
sebbacon: Just a thought: using a unified diff format for comparing the lists of numbers (using `difflib`) might make it easier to see what's going on.  (Or even for comparing the whole logs, if they format well as JSON).

evansd: Oh yes, now I think about it I used that for diffing JSON during my matrixstore smoketesting:
```python
    ...
    diff = difflib.unified_diff(as_text_lines(item_a), as_text_lines(item_b), n=8)
    return '\n'.join(diff)

def as_text_lines(data):
    return json.dumps(data, indent=2, sort_keys=True).splitlines()
```

openprescribing/pipeline/management/commands/check_numbers.py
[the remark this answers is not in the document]
inglesp: I'm thrashing around here, so would appreciate some help when we're next around at the same time!

openprescribing/pipeline/management/commands/check_numbers.py
```diff
@@ -0,0 +1,246 @@
+# coding=utf8
+
+'''
+Checks whether any numbers displayed on a selection of URLs on the sight have
```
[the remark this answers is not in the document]
inglesp: Weird.  Also weird is that I'd force pushed my fix to this before I opened the PR, so I've no idea why you've seen it here!

openprescribing/pipeline/management/commands/check_numbers.py
```diff
@@ -0,0 +1,246 @@
...
+from django.core.urlresolvers import get_resolver
+
+from bs4 import BeautifulSoup
+from selenium import webdriver
+
+# from openprescribing.slack import notify_slack
+from openprescribing.utils import mkdir_p
+from pipeline.runner import in_progress as import_in_progress
+
+def notify_slack(msg):
```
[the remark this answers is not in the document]
inglesp: Again, something I'd fixed in a force push...  I'd used it for local testing.

openprescribing/frontend/management/commands/scrape_numbers.py
inglesp: It's because the file contains a non-ascii `£`.  Without this line, Python raises `SyntaxError: Non-ASCII character '\xc2' in file...`.

openprescribing/frontend/management/commands/scrape_numbers.py
inglesp: Nice idea, but I don't think it'd be particularly helpful, as we don't have information about where any of the numbers are from.  I anticipate that if we find that any numbers have changed over the deploy, we'll need to visually compare the old and new versions of the pages (which we've saved) to understand what the difference is.

openprescribing/frontend/management/commands/scrape_numbers.py
inglesp: Bleugh.

openprescribing/pipeline/management/commands/check_numbers.py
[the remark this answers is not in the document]
inglesp: Have made a couple of tweaks to the code to get Firefox to run headlessly.  It requires the geckodriver binary to be on the path.  It's downloadable from https://github.com/mozilla/geckodriver/releases/tag/v0.24.0.  Where should I put it?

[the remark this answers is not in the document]
evansd: I'd go for `/usr/local/bin`. Or at least symlink to it from there, even if you want to put the binary itself somewhere else.

PR #1881, opened 2019-07-17
Remove import-export library

inglesp:
It wasn't being used and added several seconds to inglesp's startup time
locally.


sebbacon:
I don't mind removing it, but just to note the use-case: I added it in because people kept asking how many CCGs were signed up to alerts and I wanted to make it self-service.

sebbacon:
> several seconds

Mildly interested in why.... any idea?

inglesp:
>>  several seconds

> Mildly interested in why.... any idea?

It does a load of autodiscovery stuff.  (I investigated a few months ago but can't remember the details.)

PR #1883, opened 2019-07-17
Measures navigation improvements

### LINKED ISSUE #1006: Low Priority navigation and other issues
I think the Low-priority and LP omnibus measures for practices and CCGs could be easier to navigate by adding more links. This is an issue because people may not realise they can access all the possible comparisons/breakdowns that we have available, or have to click through many views to get what they are after. Sometimes using the back button solves the issues but if links are shared it can be difficult to access some of the other views. 

### Issue 1
From a CCG's homepage you can click "View the 17 measures" to see a breakdown of the measures for the whole CCG. Or, if you click on the measure itself you see it broken down across the practices in the CCG. It's not obvious that these links will lead you to different pages. (Not sure if there's an easy solution for this except stating the second option in the text also).

### Issue 2 
It's difficult to navigate all the pages because a lot of them aren't linked to each other, for example haing clicked on either of the links above, it's difficult to get to [this](https://openprescribing.net/measure/lpzomnibus/) page where you can compare the omnibus measure across all CCGs. 

### Summary of page navigation as it stands

| page | navigates to | also navigates to | | problem |
| -- | -- | -- | -- | -- |
| CCG homepage | CCG-level breakdown into individual LP measures | practice-level omnibus | | not obvious that the two links take you to different pages image
| CCG-level breakdown into individual LP measures | any breakdown for the individual measures image | standard measures | | no direct link back to omnibus, for practices or CCGs
| practice-level omnibus (for any CCG) | CCG-level breakdown into individual LP measures | | | no direct link back to omnibus for whole CCG, or practice-level breakdown into individual LP measures image
| practice-level breakdown into individual measures | CCG-level breakdown into individual LP measures | *presentation*-level breakdown | | no link to practice-level omnibus
| measure (all CCGs) page | practice-level omnibus (for any CCG) | | | no link to CCG-level breakdown into individual LP measures

evansd:
Improves some of the interlinking between the various measure/organisation pages.

Closes #1006

### DIFF
--- a/openprescribing/frontend/views/views.py
+++ b/openprescribing/frontend/views/views.py
@@ -1272,13 +1272,17 @@ def _build_measure_options(options):
         if options.get('aggregate'):
             options['chartTitleUrlTemplate'] = _url_template('measure_for_all_ccgs')
         elif options['orgType'] == 'regional_team':
-            options['chartTitleUrlTemplate'] = _url_template('measure_for_ccgs_in_regional_team')
+            options['chartTitleUrlTemplate'] = _url_template(
+                'measure_for_one_regional_team'
+            )
         elif options['orgType'] == 'stp':
-            options['chartTitleUrlTemplate'] = _url_template('measure_for_ccgs_in_stp')
+            options['chartTitleUrlTemplate'] = _url_template('measure_for_one_stp')
         elif options['orgType'] == 'pcn':
-            options['chartTitleUrlTemplate'] = _url_template('measure_for_practices_in_pcn')
+            options['chartTitleUrlTemplate'] = _url_template('measure_for_one_pcn')
+        elif options['orgType'] == 'ccg':
+            options['chartTitleUrlTemplate'] = _url_template('measure_for_one_ccg')
         else:
-            options['chartTitleUrlTemplate'] = _url_template('measure_for_practices_in_ccg')
+            options['chartTitleUrlTemplate'] = _url_template('measure_for_one_practice')
     else:
         view_name = 'measures_for_one_{}'.format(options['orgType'])
         options['chartTitleUrlTemplate'] = _url_template(view_name)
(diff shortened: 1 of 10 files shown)


openprescribing/templates/measure_for_children_in_entity.html
```diff
@@ -17,6 +17,11 @@ <h1>{{measure.name}} by {{ child_entity_type_human }}s in {{ parent.name }}</h1>
     results for the individual {{ child_entity_type_human }}, plus deciles
     across all {{ child_entity_type_human }}s in the NHS in England.
   </p>
+  <p>
+    <a href="{{ parent_entity_measure_url }}" class="btn btn-default">
+      View measure as combined total for this {{ parent_entity_type_human }} &rarr;
```
inglesp: Not sure I would know what "View measure as combined total for this CCG" means.  What about "Break the measure down into individual presentations for this CCG"?  This is closer to terminology we use elsewhere.

openprescribing/frontend/views/views.py
```diff
@@ -580,8 +580,13 @@ def _measure_for_children_in_entity(
             'measure_for_one_' + parent_entity_type,
             kwargs={'measure': measure.id, 'entity_code': parent_entity_code}
         ),
+        'other_measures_url': reverse(
```
inglesp: What about `all_measures_url`?

openprescribing/templates/measure_for_all_entities.html
```diff
@@ -21,6 +21,11 @@ <h1>{{measure.name}} by all {{ entity_type_human }}s</h1>
     plus deciles across all {{ entity_type_human }}s in the
     NHS in England.
   </p>
+  <p>
+    <a href="{% url 'all_england' %}#{{ measure.id }}" class="btn btn-default">
+      View measure for NHS England combined &rarr;
```
inglesp: "combined" doesn't sound quite right to me.  Not quite sure what to suggest though.

openprescribing/frontend/views/views.py
evansd: Yes! Thanks

evansd:
Thanks @inglesp! I think what I'll do is merge this as it is and then create a separate issue for ~~bikeshedding~~ thoughtful user-focused discussion over the wording as Helen's already made some suggestions and I'm sure others will want to chime in to.

Issue #1835, opened 2019-07-11
Improve ability to cross-browse with DMD


sebbacon:
There is scope to embed some time trends on VMP/AMP pages, which would be quite nice, but before that, some basic cross-linkage to help people navigate and to help spiders index our site:

At all levels:
 * Show BNF Code (and "no BNF code available") if none
 * Link to analyse form with BNF Code
 * Indicate BNF section and BNF Chemical with link to trends   ([e.g.](https://openprescribing.net/chemical/0101010C0/))
 * Link to price-per-unit breakdown page at all-England level ([e.g.](https://openprescribing.net/all-england/0601060D0AAA0A0/price-per-unit/?date=2019-04-01))

At VMPP / AMPP level:
 * Link to tariff history if in the tariff ([e.g.](https://openprescribing.net/tariff/?codes=0905022C0AAAFAF))
 * Indicate if there are price concessions in the current month 

At VTM level:
 * Improve our mapping process so we assign BNF chemical codes to VTMs (this one is a bit more work than other points in this issue)
 * Use this to link to "chemical trends" section

On PPU presentation-breakdown page ([e.g.](https://openprescribing.net/all-england/0601060D0AAA0A0/price-per-unit/?date=2019-04-01)):
  * Link to the VMP in the DM+D

On BNF section and chemical trends:
* Link to a list of all matching DMD entities?

On front page of DM+D browser:
* Add a pageable complete list so it can be crawled 

On main menu:
* Add link to DMD browser

PR #1885, opened 2019-07-18
Add links to/from d+md pages and rest of site

inglesp:
Addresses #1835.


openprescribing/frontend/tests/test_models.py
```diff
@@ -249,3 +252,42 @@ def test_regional_team_with_no_pcts_excluded(self):
...
+            possible_savings=10,
+        )
+
+        GenericCodeMapping.objects.create(
+            from_code='0101010I0AAAEAE',
+            to_code='0101010I0AAABAB'
+        )
+
+        # Mag Ox_Cap 100mg
+        self.assertTrue(PPUSaving.objects.exists_for_bnf_code('0101010I0AAABAB'))
```
sebbacon: Can we add a test for the code path where the mapping ends in a percent?

openprescribing/dmd2/forms.py
```diff
@@ -21,7 +21,7 @@
 
 class SearchForm(forms.Form):
     q = forms.CharField(
-        label='Query or SNOMED code',
+        label='Query, SNOMED code, or BNF code/prefix',
```
sebbacon: The BNF prefix will only work from the chemical level, i.e. 9 chars and longer, right?  We need to somehow make this clear - in the label, and perhaps in validation too?

inglesp: > The BNF prefix will only work from the chemical level, i.e. 9 chars and longer, right? 

No -- we just do `Q(bnf_code__startswith=q.upper())` in `DMDObjectQuerySet.search()`.

Currently we ensure that `q` is at least 3 characters long, but maybe it'd be sensible to make it longer.

openprescribing/frontend/tests/test_models.py
inglesp: Good spot.

inglesp: No need to do this now.

PR #1895, opened 2019-07-22
Use LZ4 compression rather than Zstandard

evansd:
LZ4 (http://www.lz4.org/) is a compression standard optimised for
extremely fast decompression. In particular it is more than twice as
fast as Zstandard (http://www.zstd.net/) which is what we currently use.
As decompression can represent around 50% of the call time when summing
large numbers of matrices it is worth swapping to something faster.

The trade-off, of course, is that the compression ratio is not as good.
Below is a table giving the compressed sizes and compression times for a
full set of prescribing data using Zstandard at level 16 (our current
compression scheme) and various levels of LZ4. (Note that decompression
time is basically constant across compression levels.)

| Compression | Size  | Time |
| ----------- | ----: | ---: |
| None        |  14GB |  N/A |
| Zstd (16)   | 3.4GB |  45m |
| LZ4 (16)    | 4.4GB |  37m |
| LZ4 (10)    | 4.4GB |  15m |
| LZ4 (8)     | 4.5GB |  12m |
| LZ4 (6)     | 4.6GB |   7m |

As the table shows, there's no point going above level 10 (no real gain
in compression ratio) and little point in going below 10 (which is
already significantly faster than the status quo).

The increased file size is mildly annoying. However the increase is not
so large that it makes any categorical difference to how we manage the
data: it's still easily small enough to fit in RAM, and to download and
work with locally. And the performance benefits are clearly worth it.

For the moment, we retain the ability to read Zstandard compressed
blocks so that we can safely deploy this patch before we build a new
matrixstore file using the new compression scheme. As soon as we've done
that and confirmed that there are no issues we can remove all Zstandard
related code.


openprescribing/matrixstore/serializer.py
```diff
@@ -61,17 +61,21 @@ def serialize(obj):
...
 
 
 def deserialize(data):
     """
     Deserialize binary data using our custom PyArrow context, automatically
     detecting compressed data and decompressing if necessary
     """
-    if memoryview(data)[:4] == ZSTD_MAGIC_NUMBER:
-        data = decompressor.decompress(data)
+    magic_number = memoryview(data)[:4]
```
sebbacon: TIL `memoryview`

PR #1900, opened 2019-07-24
Ensure insert_rows_from_pg always uses correct schema 

### LINKED ISSUE #1889: No measures in database after `import_measures`
During the 2019_05 import, `import_measures` ran, but there are no `MeasureValue` or `MeasureGlobal` objects in the database.

This is because the `setting` column on `hscic.practices` is empty.  But why?!

inglesp:
This is an improvement over what we have now, and should mean that we avoid problems like #1889 in future.  Longer term, we should aim for BQ and PG tables to match exactly.

Closes #1889.

### DIFF
--- a/openprescribing/frontend/management/commands/generate_measure_fixtures.py
+++ b/openprescribing/frontend/management/commands/generate_measure_fixtures.py
@@ -261,7 +261,7 @@ class Command(BaseCommand):
             for row in prescribing_rows:
                 writer.writerow(row)
             f.seek(0)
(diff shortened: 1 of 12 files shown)


inglesp:
> Great. One question: should perhaps `schema` be a required argument to `insert_rows_from_csv`, too?

I thought about that, but it's used quite a lot in test setup, and I was feeling lazy.  Rectified in 8f0b4388f1be0ae31daf1bf227a6b2bd095a0f92.

PR #1923, opened 2019-07-30
Remove unused "materialized views" and supporting code

evansd:
These views are in fact ordinary Postgres tables created directly
through SQL (i.e. not using Django models or migrations) and populated
by querying BigQuery. Now that the MatrixStore powers the Analyse page
and the Price Concessions pages, these tables are no longer required.

The tables that have been removed are:

    vw__ccgstatistics
    vw__chemical_summary_by_ccg
    vw__chemical_summary_by_practice
    vw__practice_summary
    vw__presentation_summary
    vw__presentation_summary_by_ccg

Once we've confirmed that there are no issues with this patch, we should
delete these tables from Postgres.

Note that there are a few other tables which follow the same naming
convention but are slightly different:

 1. The `vw__medians_for_tariff` table is a genuine Postgres materialized
    view and is used by the Ghost Branded Generics code.

 2. The `vw__median_price_per_unit` and `vw__ghost_generic_measure`
    tables are BigQuery views which are referenced in the SQL for some
    measures.  These only exist in BigQuery and are not imported into
    Postgres.

Depends on #1913


evansd:
It was extremely satisfying!

On the end-to-end tests, I think the answer is probably "yes". But this makes me realise that I don't actually know what the end-to-end tests do, and how they differ from the usual tests. I'm sure I can discover myself with a bit of reading, so don't feel obliged to explain it all to me. But I thought I should flag up my ignorance :)

PR #1927, opened 2019-07-30
Send All England alerts as part of monthly alerts

### LINKED ISSUE #1880: Send All England alerts automatically as part of `send_alerts`

evansd:
There's a bit of complexity here because the two commands don't respect
the same set of options so we only attempt to send the All England
alerts if there are no additional options provided.

Closes #1880

### DIFF
--- a/openprescribing/frontend/management/commands/send_monthly_alerts.py
+++ b/openprescribing/frontend/management/commands/send_monthly_alerts.py
@@ -14,6 +14,7 @@ from frontend.models import User
 
 from common.alert_utils import EmailErrorDeferrer
 from frontend.views import bookmark_utils
+from .send_all_england_alerts import send_alerts as send_all_england_alerts
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


openprescribing/frontend/management/commands/send_monthly_alerts.py
```diff
@@ -165,6 +166,27 @@ def send_search_bookmark_email(self, search_bookmark, now_month):
         except bookmark_utils.BadAlertImageError as e:
             logger.exception(e)
 
+    def send_all_england_alerts(self, options):
+        # The `send_all_england_alerts` command doesn't respect the same set of
+        # options that this command does, so we only invoke it for certain
+        # limited sets of options
+        set_options = {k: v for (k, v) in options.items() if v is not None}
```
inglesp: Sometimes, I miss Ruby.

evansd: Yes, I think it was the ergonomics of operations like this which originally drove Gary Bernhardt from Python to Ruby.

PR #1933, opened 2019-07-31
Fix dm+d links from tariff page

inglesp:
We were showing links to every single VMP


inglesp:
Merging myself, as this is quite badly broken!

PR #1941, opened 2019-08-02
Don't include presentations without prescribing in MatrixStore

evansd:
There doesn't seem to be any significant performance difference here,
but without this we have to add an irritating extra `IS NOT NULL` clause
to the end of every MatrixStore query.


evansd:
Apologies for the onslaught of small PRs. I'm trying to get all the changes to the on-disk MatrixStore format done before the next build.

PR #1943, opened 2019-08-05
Fix NCSO functions for orgs with no prescribing data

evansd:
At present, trying to load the homepage for a practice which has long
since closed triggers an error.


evansd:
Merging myself as there's no one around to review

PR #1944, opened 2019-08-05
Only check for alert signup for supported org types

evansd:
Only a subset of organisation types support signing up for NCSO Price
Concession alerts. The code to check whether the current users has
already subscribed to alerts will throw an error if run against
unsupported org types, so we make sure not to do that.


evansd:
Yes, I agree it could all do with an overhaul. But, to be fair to the code, it blew up nice and early with a clear error message which is big improvement on subtly failing.

PR #1945, opened 2019-08-05
Add "Search dm+d" link on object page

inglesp:


openprescribing/dmd2/templates/dmd/dmd_obj.html
```diff
@@ -4,8 +4,9 @@ <h1>{{ title }}</h1>
 
 {% if obj_type != 'VTM' %}
   <p>
+    <a href="">Search dm+d</a>
```
evansd: This doesn't look right. Am I missing something?

inglesp:
> This doesn't look right.

It's not.

> Am I missing something?

You're not.

Sorry.  Could you re-review please?

PR #1947, opened 2019-08-05
Attempt to better explain ndarray vs matrix complexity

evansd:
This PR adds some comments to try to explain the complexity we face around the `matrix` and `ndarray` type in numpy. It also cleans up some of the code that deals with this issue, which was written before I actually understood what was going on.

### DIFF
--- a/openprescribing/matrixstore/row_grouper.py
+++ b/openprescribing/matrixstore/row_grouper.py
@@ -81,15 +81,13 @@ class RowGrouper(object):
         rows = len(self._group_selectors)
         columns = matrix.shape[1]
         grouped_output = numpy.empty((rows, columns), dtype=matrix.dtype)
-        # There's some complexity here depending on whether `matrix` is an
-        # ndarray (which it usually will be as the result of `MATRIX_SUM()` is
-        # always an ndarry) or the older-style matrix class (which it will be
-        # if we call this function directly with a sparse matrix). If our input
-        # is a matrix, then we can only write sum results into a matrix so we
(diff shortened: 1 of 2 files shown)


openprescribing/matrixstore/row_grouper.py
```diff
@@ -121,6 +118,20 @@ def sum_one_group(self, matrix, group_id):
 def has_type_matrix(value):
     """
     Return whether `value` is a numpy array or a numpy matrix
+
```
inglesp: I've found, twice, when reading this code, that I parse "has_type_matrix" as "has a type matrix", and then I wonder what a "type matrix" is.

What about renaming to "is_matrix"?

openprescribing/matrixstore/row_grouper.py
```diff
@@ -81,14 +81,12 @@ def sum(self, matrix):
...
-        # ndarray (which it usually will be as the result of `MATRIX_SUM()` is
-        # always an ndarry) or the older-style matrix class (which it will be
-        # if we call this function directly with a sparse matrix). If our input
-        # is a matrix, then we can only write sum results into a matrix so we
-        # use the `asmatrix` method below to create a compatible view onto the
-        # underlying output array.  In either case, our return value is always
-        # an ndarray.
+        # This is awkward. We always want to return an `ndarray` even if the
+        # input type is `matrix`. But where the input is a `matrix` the `out`
+        # argument to `sum` below must be a `matrix` also. So we need a view on
```
inglesp: For clarity, change "`sum`" to "`numpy.sum`"?

PR #1958, opened 2019-08-07
Support being run under both supervisor and systemd

evansd:
This PR modifies gunicorn-related scripts so that they can run under
either supervisor (which they do at present) or systemd. This adds
complexity temporarily, but makes the process of updating our server
configuration much easier.

The primary motivation for this is to make use of systemd's socket
management which allows us to do application restarts which are both
graceful (i.e. involve no dropped connections or error responses) and
full (i.e. involve starting a completely new process instead of trying
to load new code into an existing process). Gunicorn's graceful restarts
are not full, and this resulted in many frustrating issues.


evansd:
Below is the recipe (already tested but reverted for now) for running staging under systemd. The same applies _mutatis mutandis_ for production.

We need two, root-owned, systemd config files:

`/etc/systemd/system/app.openprescribing_staging.web.service`
```ini
[Service]
User=www-data
ExecStart=/webapps/openprescribing_staging/bin/gunicorn_start staging
SyslogIdentifier=app.openprescribing_staging.web
Restart=always
RestartSec=4

[Install]
WantedBy=multi-user.target
```

`/etc/systemd/system/app.openprescribing_staging.web.socket`
```ini
[Socket]
ListenStream=/webapps/openprescribing_staging/run/gunicorn.sock

[Install]
WantedBy=sockets.target
```

First we stop and remove the supervisor service:
```sh
sudo supervisorctl stop openprescribing_staging
cd /etc/supervisor/conf.d/
sudo mv openprescribing_staging.conf openprescribing_staging.conf.disabled
sudo supervisorctl update
```

Then we enable and start the systemd service:
```sh
sudo systemctl enable --now app.openprescribing_staging.web.socket
sudo systemctl enable --now app.openprescribing_staging.web.service
``` 

Done.

sebbacon:
Good stuff.

Some observations

1) When we switch this over, ideally we'd include the config in fabric / `deploy/` somehow, per our other projects. Or at a minimum check in the file and reference it in the README.
2) For consistency, I'd like us to use the same system control template across all our projects. I guess pending a decision about using GCS or not. But important not to lose sight of it.
3) Do we definitely want this behaviour?

```ini
Restart=always
RestartSec=4
```

If we wanted to match the current behaviour of supervisord`, I think we'd want something more like this:

```ini
[Service]
Restart=on-failure
[Unit]
StartLimitInterval=150
StartLimitBurst=3
```

However, that's the *default* behaviour of `supervisord`, so I doubt that was a deliberate choice, and we could decide to change it.

The only circumstances where we might expect a failure which could be automatically recovered from are resource depletions (sockets, memory). I think having a maximum number of restarts (or possibly even none) is a sensible thing to do.

4) I'd like to discuss the graceful part with you to understand it better - let's catch up shortly!

PR #1961, opened 2019-08-08
Remove unused imports

inglesp:
Found via `flake8 --select=F401`

### DIFF
--- a/openprescribing/dmd2/management/commands/generate_dmd_fixture.py
+++ b/openprescribing/dmd2/management/commands/generate_dmd_fixture.py
@@ -10,7 +10,7 @@ have foreign keys to, such as UnitOfMeasure or VTM.
 from django.core import serializers
 from django.core.management import BaseCommand
 from django.db.models.fields.related import ForeignKey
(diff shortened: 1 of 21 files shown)


evansd:
The `ready()` method is where you would attach signal handlers, if you were going to. But we don't, so the whole method is unnecessary.

Technically, we could drop this whole file and remove the `default_app_config` definition from [`frontend/__init__.py`](https://github.com/ebmdatalab/openprescribing/blob/32763048e957f7e4e373a8c49d68720ce32f82f2/openprescribing/frontend/__init__.py) and that would work.

But I think the official way of doing things now would be to remove the `default_app_config` variable and then change the reference in `INSTALLED_APPS` from `frontend` to `frontend.apps.FrontendConfig`.

The idea is, I think, to make the app registration process a bit more explicit and to support having multiple, differently configured instances of the same app. The old way still works via various fallback mechanisms, but I don't know how long they plan to support that.

inglesp:
Closing in favour of #1969.

PR #1969, opened 2019-08-09
Make flake8 happy

inglesp:


evansd:
The imports in `frontend.apps.FrontendConfig.ready()` turned out to have a purpose after all, which is to attach the signal handlers and register the admin components. I've reinstated them with an explanatory comment, plus a comment to disable that specific flake8 warning.

Also, the travis lint job needs to install `flake8` as well as `black`. As black is Python 3 only we have the awkward situation (hopefully temporary) where our linter requirements are separate from the rest of our requirements.

evansd:
I have a reasonably strong personal preference to only use flake8's semantic checks and to disable anything purely syntactic. Black takes care of all syntactic matters relating to the code itself. And I have very little enthusiasm for flake8 telling me how many spaces I ought to put in my comments.

Whether we do this via whitelisting the semantic stuff or blacklisting the syntactic I don't really mind.

sebbacon:
Love this :heart_eyes:

PR #1971, opened 2019-08-10
Fix size of PPU bubbles in cases where all drugs are prescribed in similar amounts

### LINKED ISSUE #1970: PPU blog seems out of proportion
Issue #460 mentions that there is a minimum blob size.

image
image

The above example shows whether there are two presentations prescribed.  Proportions are similar (188 vs 154), but blog sizes are vastly different.  

This led to members of the MO team assuming (by eyeballing the charts) that there wasn't much prescribed, which is inaccurate.

What is the current metholdology for blog size?

inglesp:
Fixes #1970.

From https://openprescribing.net/practice/L83016/0402010ABAAAXAX/price_per_unit/?date=2019-05-01#local-panel... before:

image

After:

image

### DIFF
--- a/openprescribing/media/js/src/bubble.js
+++ b/openprescribing/media/js/src/bubble.js
@@ -27,11 +27,6 @@ domready(function() {
       subtitle: {
         text: null,
       },
-      plotOptions: {
-        bubble: {
-          minSize: 3,
-        },
-      },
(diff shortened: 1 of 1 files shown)


evansd:
Merging this in Peter's absence as Rich is keen for the fix

PR #1977, opened 2019-08-12
Remove supervisor related code

evansd:
Now that we're happily running under systemd this is not required.

### DIFF
--- a/deploy/fab_scripts/graceful_reload.sh
+++ b/deploy/fab_scripts/graceful_reload.sh
@@ -17,27 +17,6 @@ case "$script_dir" in
   ;;
 esac
 
-# Service names follow the pattern:
-#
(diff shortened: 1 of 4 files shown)


deploy/fab_scripts/graceful_reload.sh
```diff
@@ -40,22 +40,4 @@ esac
...
-  pid_file="$script_dir/../../run/gunicorn.pid"
-  PID=$(cat "$pid_file")
-  if [[ -n "$PID" ]]; then
-      kill -HUP $PID;
-  else
-      echo "Error: server $1 not running, so could not reload";
-      exit 1;
-  fi
-fi
+systemctl restart "$service_name"
```
inglesp: Is this file the right place for the `Service names follow the pattern:...` comment?

PR #1979, opened 2019-08-12
DO NOT MERGE: Code mockup of how a MatrixStore API could look

evansd:
This is a mockup of how interacting with the MatrixStore might look using an imagined `LaballedArray` class which is a wrapper around a `numpy.ndarray` that associates string labels with row and column indices.

Done as a PR to get the review and comments tooling. Obviously not for merging.

### DIFF
--- a/openprescribing/api/views_spending.py
+++ b/openprescribing/api/views_spending.py
@@ -429,9 +429,8 @@ def _get_total_prescribing_entries(bnf_code_prefixes):
     across all practices for all presentations matching the supplied BNF code
     prefixes
     """
-    db = get_db()
     items_matrix, quantity_matrix, actual_cost_matrix = _get_prescribing_for_codes(
-        db, bnf_code_prefixes
+        bnf_code_prefixes
     )
     # If no data at all was found, return early which results in an empty
     # iterator
@@ -442,20 +441,17 @@ def _get_total_prescribing_entries(bnf_code_prefixes):
     # powered by the `vw__presentation_summary` table which summed over all
     # practice types)
     group_all = get_row_grouper("all_practices")
-    items_matrix = group_all.sum(items_matrix)
-    quantity_matrix = group_all.sum(quantity_matrix)
-    actual_cost_matrix = group_all.sum(actual_cost_matrix)
+    items_matrix = group_all.sum_one_group(items_matrix)
+    quantity_matrix = group_all.sum_one_group(quantity_matrix)
+    actual_cost_matrix = group_all.sum_one_group(actual_cost_matrix)
     # Yield entries for each date (unlike _get_prescribing_entries below we
     # return a value for each date even if it's zero as this is what the
     # original API did)
-    for date, col_offset in sorted(db.date_offsets.items()):
-        # The grouped matrices only ever have one row (which represents the
-        # total over all practices) so we always want row 0 in our index
-        index = (0, col_offset)
+    for date in items_matrix.labels:
         yield {
-            "items": items_matrix[index],
-            "quantity": quantity_matrix[index],
-            "actual_cost": round(actual_cost_matrix[index], 2),
+            "items": items_matrix[date],
+            "quantity": quantity_matrix[date],
+            "actual_cost": round(actual_cost_matrix[date], 2),
             "date": date,
         }
 
@@ -584,9 +580,8 @@ def _get_prescribing_entries(bnf_code_prefixes, orgs, org_type, date=None):
     If a date is supplied then data for just that date is returned, otherwise
     all available dates are returned.
     """
-    db = get_db()
     items_matrix, quantity_matrix, actual_cost_matrix = _get_prescribing_for_codes(
-        db, bnf_code_prefixes
+        bnf_code_prefixes
     )
     # If no data at all was found, return early which results in an empty
     # iterator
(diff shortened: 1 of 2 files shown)


openprescribing/frontend/views/spending_utils.py
```diff
@@ -65,27 +53,25 @@ def ncso_spending_for_entity(entity, entity_type, num_months, current_month=None
...
-    tariff_costs = costs.tariff_costs[:, 0]
-    extra_costs = costs.extra_costs[:, 0]
-    quantities = costs.quantities[:, 0]
-    names = _get_names_for_bnf_codes(costs.bnf_code_offsets.keys())
+    quantities, tariff_costs, extra_costs = _get_concession_cost_matrices(
+        month, month, org_type, org_id
+    )
+    # We sum across columns (i.e. dates) to get the totals over time for each
+    # BNF code.
+    quantities = quantities.sum_columns()
```
inglesp: Would it be possible/sensible to make this something like `quantities.sum_over_dates()` (might require metaprogramming) or `quantities.sum('dates')`?

openprescribing/frontend/views/spending_utils.py
```diff
@@ -116,113 +102,51 @@ def _get_names_for_bnf_codes(bnf_codes):
 
 def _get_concession_cost_matrices(min_date, max_date, org_type, org_id):
     """
-    Given a date range (inclusive) and an organisation return a set of matrices
-    detailing all spending affected by price concessions during the period.
-
-    Returns a ConcessionCostMatrices object with the following attributes:
+    Given a date range (inclusive) and an organisation return three
+    LaballedArrays (of the same shape) detailing all spending affected by price
```
inglesp: `LabelledArray` suggests something 1-dimensional -- why not `LabelledMatrix`?

openprescribing/frontend/views/spending_utils.py
```diff
@@ -116,113 +102,51 @@ def _get_names_for_bnf_codes(bnf_codes):
...
-        date_offsets=prices.date_offsets,
-        quantities=quantities,
-        tariff_costs=prices.tariff_prices * quantities,
-        extra_costs=prices.price_increases * quantities,
-    )
-
-
-def _get_prescribed_quantity_matrix(bnf_code_offsets, date_offsets, org_type, org_id):
+    tariff_prices, price_increases = _get_concession_price_matrices(min_date, max_date)
+    bnf_codes = tariff_prices.row_labels
```
inglesp: Depending on how much the user of the API should know about the implementation of the `LabelledArray`, could this be `tariff_prices.labels('bnf_codes')`?

openprescribing/api/views_spending.py
```diff
@@ -442,20 +441,17 @@ def _get_total_prescribing_entries(bnf_code_prefixes):
...
-    for date, col_offset in sorted(db.date_offsets.items()):
-        # The grouped matrices only ever have one row (which represents the
-        # total over all practices) so we always want row 0 in our index
-        index = (0, col_offset)
+    for date in items_matrix.labels:
         yield {
-            "items": items_matrix[index],
-            "quantity": quantity_matrix[index],
-            "actual_cost": round(actual_cost_matrix[index], 2),
+            "items": items_matrix[date],
```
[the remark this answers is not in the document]
evansd: Actually, my intention was that `__getitem__` would look up values. So for a n-dimensional array it would take a n-tuple and, as a convenience, for a 1-dimensional array it would accept a single index value. In other words, it would mimic numpy's `__getitem__` minus all the slicing stuff.

I've realised that the example above is unnecessarily confusing because I've reused variable names:
```python
items_matrix = group_all.sum_one_group(items_matrix)
```
After the `sum_one_group` call, `items_matrix` is now a 1-dimensional array indexed by date.

openprescribing/frontend/views/spending_utils.py
evansd: Ah yes, good idea! I think I prefer the non-metaprogramming version, although the resulting API isn't quite as pretty.

openprescribing/frontend/views/spending_utils.py
evansd: "Matrix" isn't quite right either, as these can be 1-dimensional after summing rows/columns. I was following numpy's "ndarray" terminology, but I'd be open to other suggestions.

openprescribing/frontend/views/spending_utils.py
evansd: Yes. As with the `sum_columns` discussion above, I think we should give axes names and then dispense with any references to rows or columns.

openprescribing/frontend/views/spending_utils.py
evansd: Summarising our conversation for the record: we've decided that "matrix" is probably the least confusing name, even if it is a bit wrong sometimes

openprescribing/frontend/views/spending_utils.py
sebbacon: Or per my suggestion above

```python
quantities.aggregate(by="date", numpy.sum)
```

openprescribing/frontend/views/spending_utils.py
```diff
@@ -65,27 +53,25 @@ def ncso_spending_for_entity(entity, entity_type, num_months, current_month=None
...
-    names = _get_names_for_bnf_codes(costs.bnf_code_offsets.keys())
+    quantities, tariff_costs, extra_costs = _get_concession_cost_matrices(
+        month, month, org_type, org_id
+    )
+    # We sum across columns (i.e. dates) to get the totals over time for each
+    # BNF code.
+    quantities = quantities.sum_columns()
+    tariff_costs = tariff_costs.sum_columns()
+    extra_costs = extra_costs.sum_columns()
+    bnf_codes = quantities.labels
```
sebbacon: `quantities.row_labels`?

openprescribing/frontend/views/spending_utils.py
```diff
@@ -65,27 +53,25 @@ def ncso_spending_for_entity(entity, entity_type, num_months, current_month=None
...
-    for bnf_code, offset in costs.bnf_code_offsets.items():
+    for bnf_code in bnf_codes:
         results.append(
             (
                 bnf_code,
                 names[bnf_code],
-                int(quantities[offset]),
-                float(tariff_costs[offset]),
-                float(extra_costs[offset]),
+                int(quantities[bnf_code]),
```
sebbacon: Or per prev suggestion 

```python
int(quantities.getValue(bnf_code="01040D0BCBC"))
```

PR #1987, opened 2019-08-14
Rewrite numerator breakdown API to use MatrixStore

evansd:
This involves two behavioural changes to the API, both -- hopefully -- improvements:
  * the `presentation_name` field is now taken from DM&D rather than the oddly truncated BNF name;
 * the All England breakdown now only includes practices which belong to a CCG which better matches the measure calculation and what we do elsewhere on the site.

The code itself suffers from the same clunkiness and excessive indirection as we've already identified in other MatrixStore code, and can be cleaned up when the new abstraction is finished.

### DIFF
--- a/openprescribing/api/views_measures.py
+++ b/openprescribing/api/views_measures.py
@@ -112,8 +112,10 @@ def measure_numerators_by_org(request, format=None):
     org_type, org_id = _get_org_type_and_id_from_request(request)
     group_by_org = get_row_grouper(org_type)
 
-    # Closure over `group_by_org` and `org_id` which takes a prescribing matrix
-    # and returns the total value for that organisation over the last N months
+    # Nested function which takes a prescribing matrix and returns the total
+    # value for the current organisation over the last 3 months (where the
+    # current organisation is defined by the `group_by_org` and `org_id`
+    # variables)
     def get_total(matrix):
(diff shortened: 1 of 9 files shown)


openprescribing/api/views_measures.py
```diff
@@ -79,141 +79,132 @@ def measure_global(request, format=None):
...
-            three_months_ago=three_months_ago,
-            extra_join=extra_join,
-            order_col=order_col,
+    measure_id = request.query_params.get("measure", None)
+    measure = Measure.objects.get(pk=measure_id)
+    org_type, org_id = _get_org_type_and_id_from_request(request)
+    group_by_org = get_row_grouper(org_type)
+
+    # Closure over `group_by_org` and `org_id` which takes a prescribing matrix
+    # and returns the total value for that organisation over the last N months
```
inglesp: N = 3.

sebbacon: Pedant point - not actually a closure, just a nested function referencing a nonlocal variable?

inglesp: Ooh.  What's the difference?

evansd: If we returned a reference to this function outside the current scope it would become a closure. Pedantry should never be apologised for.

Issue #1930, opened 2019-07-31
Exclude closed/dormant practices when counting number of practices in CCG


HelenCEBM:
Our statement relating to practices "currently" in a CCG on CCG dashboards/measure pages isn't really true as we include closed ones. It would also be more useful if we indicate how many *open* practices there are. 
![image](https://user-images.githubusercontent.com/24392156/62200507-7e22bb00-b37d-11e9-876c-838bdde64250.png)

It's easy to count the active ones yourself in the above example but larger CCGs e.g. Kernow we state 72 practices but it takes a while to count up the closed/dormant ones.

I think we should change it to `There are 16 practices currently active in this CCG (plus 2 closed/dormant practices)`

Probably not necessary but we could change the order in which practices are listed so that the closed/dormant ones appear together at the bottom of the list.

HelenCEBM:
This is perfect on the dashboard but not quite working on the measures pages, e.g. 
![image](https://user-images.githubusercontent.com/24392156/65325014-aec8e900-dba5-11e9-89b3-f580ad0b1fef.png)

PR #2004, opened 2019-08-22
Exclude closed/dormant practices from count of practices

inglesp:


openprescribing/frontend/views/views.py
```diff
@@ -206,8 +206,15 @@ def all_pcns(request):
 def pcn_home_page(request, pcn_code):
     pcn = get_object_or_404(PCN, ons_code=pcn_code)
     practices = Practice.objects.filter(pcn=pcn, setting=4).order_by("name")
+    num_open_practices = len([p for p in practices if p.status_code == "A"])
+    num_non_open_practices = len([p for p in practices if p.status_code != "A"])
```
evansd: You could avoid repeating the `status_code` with:
```python
num_non_open_practices = len(practices) - num_open_practices
```

openprescribing/templates/_practice_list.html
```diff
@@ -0,0 +1,16 @@
+<p>
+  There are {{ num_open_practices }} practices currently active in this CCG
```
evansd: I think you need to replace `CCG` with `{% entity_type_human %}`, or whatever the variable is called, so this works with PCNs.

inglesp: Ach, I had that in the first commit.  Well spotted.

inglesp:
Fixes #1930.

PR #2007, opened 2019-08-22
added hyperlinks to paper

richiecroker:


inglesp:
@richiecroker do you want this reviewed and merged?

richiecroker:
@inglesp not sure why I didn't respond to this, yes please!

evansd:
Ah, I see. Yeah, you want a `<br>` for that as the text appears in different contexts.

richiecroker:
ok, what's current status of this, i'll make a change.

evansd:
Although actually, on reflection, even that wouldn't be great. The layout's only been designed with a single paragraph in mind and so anything we try to do will look broken in at least one of the contexts.

richiecroker:
we have others that we've used multiple paragraphs in, e.g. https://openprescribing.net/measure/seven_day_prescribing/

Too much text there to put in one paragraph.

evansd:
OK. It looks broken here: https://openprescribing.net/measure/seven_day_prescribing/all-england/

richiecroker:
is that possibly due to the fact that we've not put </p> in consistently.  I'm sure I spoke to someone about this when we did it, but maybe not.

richiecroker:
Just realised there's no <p> on the first paragraph, maybe that's why it's broken there.  Shall I add one in and see what happens?

evansd:
Sadly that will just make it even more broken :( There's literally no way to do what you want to do correctly given how things are set up. But if you're happy for it to look a bit weird in some contexts then what you originally had with the single `<p>` tag is probably the least worst option. Sorry for leading you round the houses!

richiecroker:
My view is that it's better to have readable content in most contexts, even if a bit broken in some, rather than having one massive block of text which is somewhat more indigestible.  I'll try and raise it tomorrow to see what others think.

richiecroker:
Also amused by me typing `<p>` in here and it just adding a paragraph!

evansd:
I've made an issue to fix this properly. Just sticking that `<p>` back in is fine for now (I'll do it as I need to deploy that change manually anyway).

PR #2024, opened 2019-08-27
Create WIP_enviromental_inhalers.json

brianmackenna:
This is a work in progress. Descriptions and "why it matters" needs expanding.

This is a prototype for staging as I have never created measures using a view, I'm not certain it is fully correct.


openprescribing/frontend/management/commands/measure_definitions/WIP_enviromental_inhalers
```diff
@@ -0,0 +1,38 @@
...
+  "tags": [
+    "respiratory"
+  ],
+  "numerator_columns": [
+    "SUM(items) AS numerator, "
+  ],
+  "numerator_from": "{hscic}.brian.enviromental_inhalers",
+  "numerator_where": [
+   "(form_route='pressurizedinhalation.inhalation')",
+    "1 = 1"
```
sebbacon: You wouldn't need the 1 = 1 here.

brianmackenna:
Is there a particular reason why `vw__measures` includes 2 underscores?

openprescribing/frontend/management/commands/measure_definitions/WIP_enviromental_inhalers
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "Enviromental Impact of Inhalers",
```
inglesp: Enviro_n_mental

inglesp:
The file needs to be called `something.json`.  (Before it's deployed, you'll want to rename it anyway, as the filename becomes part of the URL of the measure.)

openprescribing/frontend/management/commands/measure_definitions/WIP_enviromental_inhalers
brianmackenna: thanks and updated. I wonder how many years I have been spelling that wrong for.....

inglesp:
@brianmackenna I'm pretty sure we don't need to use a view at all here -- are you happy for me to rewrite the measure definition?

brianmackenna:
Please do - Once done if we could briefly chat on a call as well? Obviously I'm a bit unclear about views and the measure generation more generally.

brianmackenna:
@inglesp  has rewritten a new PR #2038  with a new view so I will close this one.

PR #2029, opened 2019-08-28
Extract ENV from environment in gunicorn_start

### LINKED ISSUE #2026: No metrics captured in NewRelic
There's nothing at https://rpm.newrelic.com/accounts/1550585/applications/45170403!

inglesp:
It was passed as an argument by systemd, but the version of
app.openprescribing.web.service was incorrectly passing "staging" when
it should've passed "production".

Fixes #2026.


inglesp:
Ansible tests are failing with `$ENV must be 'production' or 'staging', got ''`

This makes me think that the environment is being set up for Ansible differently.

@sebbacon can we discuss?

sebbacon:
See https://github.com/ebmdatalab/openprescribing/pull/1977/commits/245ec881f2f46c5d6f4a7b60b49ca2297dfd5d98

inglesp:
The tests are now failing because the ansible tests run with `local` settings, but `gunicorn_start` expects either `production` or `staging`.

Should we create a new `ansible` settings?  Or should we use `production` or `staging`?

PR #2031, opened 2019-08-28
Improve feedback to users when dm+d item has no BNF code

sebbacon:


sebbacon:
Will do, just wanted to see what you thought in principle

PR #2050, opened 2019-09-11
Fix definition of "mean PPU" in price-per-unit calculation

evansd:
The price-per-unit savings tool involves calculating the average
price-per-unit paid by an organisation for a given set of presentations
in a given month. Previously we were taking the average of ratios,
rather than the average ratio i.e. rather than doing `SUM(cost) /
SUM(quantity)` we were doing `AVG(cost / quantity)`.

This had the consequence that small prescriptions with an unusually high
or low price-per-unit would have a disproportionately large effect on
the overall PPU. As there is generally a floor on PPU but not a ceiling
(it's hard to pay less than a certain amount, but it's easy to pay very
high prices) this resulted in our mean PPU being higher than the true
value. This is turn meant that we were understating the savings
available through switching.

The difference turns out to be small but non-trivial. For instance, in
June 2019 the old method gave national savings of £22.9m and the
corrected method gives £24.3m.

### DIFF
--- a/openprescribing/frontend/management/commands/ppu_sql/savings_for_decile.sql
+++ b/openprescribing/frontend/management/commands/ppu_sql/savings_for_decile.sql
@@ -17,7 +17,7 @@ FROM (
       AVG(brand_count) AS avg_brand_count_per_practice,
       MAX(category) AS category,
       SUM(quantity) AS quantity,
-      AVG(price_per_unit) AS price_per_unit,
+      SUM(net_cost) / SUM(quantity) AS price_per_unit,
       SUM(possible_savings) AS possible_savings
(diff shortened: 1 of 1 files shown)


evansd:
I was actually waiting for the failure on the PR, but it never came!

I don't think there's any point adding tests for this now while I'm MatrixStore-ising it all anyway.

sebbacon:
> I don't think there's any point adding tests for this now while I'm MatrixStore-ising it all anyway

Right. I guess the main thing is to bear in mind new tests should cover this case. If you're using the same randomise-and-compute-in-different-ways approach as elsewhere, that should cover it!

PR #2052, opened 2019-09-11
Link to analyse page from measure pages

inglesp:


openprescribing/frontend/utils/bnf_hierarchy.py
```diff
@@ -6,20 +6,25 @@ def simplify_bnf_codes(bnf_codes):
...
 
     all_bnf_codes = get_all_bnf_codes()
+
+    # Drop any BNF codes for which we don't have prescribing.
+    bnf_codes = set(bnf_codes) & all_bnf_codes
+
     return _prune_paths(bnf_codes, all_bnf_codes)
 
 
 def get_all_bnf_codes():
```
sebbacon: Should this be cached?

inglesp: No, it's only called by the pipeline and anyway it's pretty quick.

inglesp:
> Do we need to update notes about how to write a measure (I think we have it documented in the wiki somewhere?) and then notify the others?

See https://github.com/ebmdatalab/openprescribing/wiki/Measure-definitions.

> Can we definitely remember to add a link to analyse form from dm+d browser so you can analyse all drugs matching a certain advanced query? (i.e. if not in this PR then do it ASAP while it's fresh?)

See #2065.

> Have you considered smoke testing this against the current data, i.e. sanity checked that the measures produce matching BNF codes?

Have done so manually for a handful of measures.

PR #2053, opened 2019-09-11
Report current git sha along with Sentry errors

evansd:
This will hopefully aid debugging.


evansd:
Not sure, to be honest. Sentry's search and filtering seem a bit limited so that might not be possible.

PR #2068, opened 2019-09-17
Upgrade to Python 3

inglesp:
- [x] Get unit and functional tests passing on Travis
- [x] Get Ansible tests passing on Travis
- [x] Get linting passing on Travis
- [x] Get end-to-end test passing
- [x] Run all fetchers
- [x] Set up Python 3 virtualenv on staging
- [x] Install on staging and kick tyres
- [ ] Build docker image for Python 3.5
- [x] Set up Python 3 virtualenv on production
- [ ] Deploy to production
- [ ] Get up and stretch


inglesp:
This is not deployable, as there are some outstanding tasks (see PR description).  But it is ready for a first review.  The commits are in the order I made them, as my attempt to reorder to make reviewing easier ended in rebase hell, but each commit should be fairly small and self-contained.

inglesp:
To deploy on staging I did the following:

1. Create virtualenv: `python3 -m venv .venv3`
1. Update `VIRTUALENV_PATH` in `environment`
1. Delete all `.pyc` files: `find . -name *.pyc -delete`
1. Deploy
1. Restart service: `sudo systemctl restart 'app.openprescribing_staging.*'` (this might not have been necessary had I done 2. and 3. before deploying)

.travis.yml
```diff
@@ -16,7 +16,7 @@ matrix:
       env: LANG=en_US.UTF-8
     - name: "Linting"
       script:
-        - pip install black==19.3b0
```
evansd: :disappointed: This does make me reconsider the 3.5 choice to be honest. I didn't think we were losing much by using it, but I note that Django 3 won't support it either so maybe it's worth reconsidering.

.travis.yml
```diff
@@ -26,7 +26,10 @@ services:
   docker
 language: python
 python:
-  - "3.5"
+  # The application runs against 3.5, which is set in the Dockerfile that is used to
```
evansd: Ah, I see. This still isn't ideal though as it means we can't use the lint script from within the development environment.

deploy/clean_up_bq_test_data.sh
```diff
@@ -2,6 +2,6 @@
 export DJANGO_SETTINGS_MODULE=openprescribing.settings.test
 LOGFILE=$(mktemp -t clean_up_bq_test_data-$(date +%Y-%m-%d)-XXXX.log)
 
-. /webapps/openprescribing/.venv/bin/activate
```
evansd: This is actually the opposite of what I was thinking. My suggestion was to replace the existing virtualenv with a symlink to `.venv35` so that we could switch virtualenvs _without_ needing to update references everywhere.

deploy/clear_cache.py
```diff
@@ -18,10 +18,8 @@ def list_cloudflare_zones():
         "X-Auth-Email": os.environ["CF_API_EMAIL"],
     }
     result = json.loads(requests.get(url, headers=headers).text)
-    zones = map(
-        lambda x: {"name": x["name"], "id": x["id"]}, [x for x in result["result"]]
-    )
-    print (zones)
+    zones = [{"name": x["name"], "id": x["id"]} for x in result["result"]]
+    print(zones)
```
sebbacon: Ha that was a weird thing you replaced!

deploy/clean_up_bq_test_data.sh
inglesp: That makes much more sense.  I'll back out this change and set up virtualenv symlinking on staging.

inglesp:
@evansd Regarding Python version and black: can we stick with 3.5 for now, and then upgrade again in a week or two once the dust has settled here?

evansd:
@inglesp Oh yes, absolutely. It's certainly no worse than the status quo. I was just hoping we wouldn't settle on 3.5 for the long term.

PR #2105, opened 2019-10-02
Fix numbers checker

inglesp:


sebbacon:
Of course if it takes longer than 60s we've got other things to worry about, so possibly redundant, but happy to merge it!

evansd:
@inglesp Now we've fixed the underlying issue in #2106 I think we should back out the timeout change

PR #2106, opened 2019-10-02
Fix incorrect aggregation SQL

evansd:
In order to generate the correct SQL it's necessary that the definition
of `calc_value` come *after* the definition of `numerator` and
`denominator`. Previously, by luck of the order in which the keyword
arguments were processed, it worked to have all the definitions in one
`annotate` call. But the upgrade to Python 3 exposed the error here, so
now we use two `annotate` calls.


evansd:
Merged as it fixes a reasonably serious bug and there's no one around to reveiw

evansd:
Incidentally, the original code is actually correct in Python 3.6+ as the dict implementation changed and now the order of keyword arguments is preserved by design.

PR #2107, opened 2019-10-03
Create task to make some BQ data public

inglesp:
* We're using tables not views, because you cannot preview a view.
* The dataset is currently called public_draft.  This can change once
  we're happy with the structure of the data.
* Table/column names are very much up for discussion, as these cannot be
  changed once the data is published.
* There's no dm+d data yet.
* For prescribing, list_size, and star_pu, we could also have versions
  containing just the latest three months of data.
* We could consider limiting data to just CCGs and GP practices.
* We could consider publishing unaggregated prescribing data too.


inglesp:
The data is currently accessible at https://console.cloud.google.com/bigquery?project=ebmdatalab&p=ebmdatalab&d=public_draft&page=dataset.

inglesp:
Further considerations:

* licensing
* BQ permissions

openprescribing/bq_public_tables/practice.sql
```diff
@@ -0,0 +1,24 @@
...
+    join_provider_date,
+    leave_provider_date,
+    CASE
+      WHEN COALESCE(address1, address2, address3, address4, address5) IS NULL THEN NULL
+      ELSE CONCAT(
+        address1,
+        CASE WHEN address2 IS NULL THEN '' ELSE CONCAT(', ', address2) END,
+        CASE WHEN address3 IS NULL THEN '' ELSE CONCAT(', ', address3) END,
+        CASE WHEN address4 IS NULL THEN '' ELSE CONCAT(', ', address4) END,
+        CASE WHEN address5 IS NULL THEN '' ELSE CONCAT(', ', address5) END
```
sebbacon: Why are we not leaving the fields as-is?

One reason this concat might not be good is if someone wanted to print envelopes (for example): the separate fields presumably denote sensible line breaks, whereas there will be commas within lines in the data.

Alternatively we could concat with line breaks

inglesp: Good question, not sure.  I'll remove it.

HelenCEBM:
Great, happy to work with Caroline on writing guidance.

Some thoughts:
- It would be useful to describe what period each dataset covers (e.g. ccgs and practices are current - so you can't see which CCG your practice belonged to last year for example, bnf is current, prescribing and list sizes cover 2010-present, or last 5 years?) - will they be automatically updated each month?
- People may often want to recreate our measures so that could be a good thing to include in guidance (e.g. how to copy and adapt from our measure definitions) - or they might want to adapt from an Analyse query. 
- we might need to make a note of which tables are not currently available that some measures rely on e.g. ADQs and opioid OME, if we release without them. These might be useful to release, though (in due course), as these will be the complex measures that can't currently be recreated in the Analyse page. 
- People may wish to select "all generics" or similar for a query. 
- Is prescribing table normalised? We should describe if so. 
- Are there any ways that the results might differ from what's on the site? (E.g setting=4)

PR #2128, opened 2019-10-08
Add include_in_alerts flag to measures

inglesp:


evansd:
Don't we need another commit which sets `include_in_alerts: false` on all existing non-core measures?

openprescribing/frontend/views/bookmark_utils.py
```diff
@@ -274,7 +274,7 @@ def _best_or_worst_performing_in_period(self, period, best_or_worst=None):
         worst = []
         measure_filter = {
             "month__gte": self.months_ago(period),
-            "measure__tags__contains": ["core"],
+            "measure__include_in_alerts": True,
```
sebbacon: This needs to be added at https://github.com/ebmdatalab/openprescribing/pull/2128/commits/4599722d046032d35b72b2c4d0714fcf244bf59e#diff-22c4b84611cefe89e70c6debe46babd7R339 too (should prob grep for `tag.*core`)?

inglesp: Oh wow, how did I miss those?!

openprescribing/measure_definitions/lpzomnibus.json
```diff
@@ -23,6 +23,7 @@
   "tags_focus": [
     "lowpriority"
   ],
+  "include_in_alerts": false,
```
sebbacon: From our conversation earlier, I'd understood the decision was to include the omnibus but not the individual LP measures?

sebbacon: Oh I'm sorry I think I misunderstood:

> Fair enough, but I now think it'd be useful for individual alerts to start to go out, as we're some time down the line on the programme.

I guess your code does this!

PR #2131, opened 2019-10-09
Add management command to dump alert signups

inglesp:


evansd:
Just checking the use case for this. We have some alert signup stuff in the web admin already.

evansd:
Oh sorry, just seen the Slack thread. Ignore my question. Although in general I think it's useful to include at least context in PRs for the sake of our future selves -- even if it's just a copy/paste of a portion of Slack conversation.

PR #2138, opened 2019-10-10
Move logic out of _build_measure_options

inglesp:
This adds significant code duplication, but increases clarity.

Next step is to remove duplication where sensible.


inglesp:
This is something I've been hacking on while measures have been building etc over the last couple of days.

I would like to get it merged soon, even though there is work to do, to avoid the branch getting stale.

Thoughts?

PR #2149, opened 2019-10-16
create measure using new dmd view

richiecroker:


openprescribing/measure_definitions/pregabalinddd.json
```diff
@@ -0,0 +1,52 @@
...
+  "is_cost_based": false,
+  "low_is_good": null,
+  "tags": [
+    "core",
+    "pain",
+    "safety"
+  ],
+  "numerator_type": "custom",
+  "numerator_columns": [
+    "SUM(p.quantity * r.strnt)/300 AS numerator, "
```
sebbacon: Why divided by 300? Can you add in a SQL comment?

openprescribing/measure_definitions/pregabalinddd.json
```diff
@@ -0,0 +1,52 @@
...
+  ],
+ "numerator_from": "{hscic}.normalised_prescribing_standard p  LEFT JOIN {measures}.measures.dmd_unit_dose_strength r ON CONCAT(SUBSTR(p.bnf_code,0,9),SUBSTR(p.bnf_code,-2)) = CONCAT(SUBSTR(r.bnf_code,0,9),SUBSTR(r.bnf_code,-2)) ",
+  "numerator_where": [
+    "p.bnf_code LIKE '0408010AE%' AND --Pregabalins"
```
evansd: Is there a stray `AND` here?

openprescribing/measure_definitions/pregabalinddd.json
```diff
@@ -0,0 +1,52 @@
...
+  "tags": [
+    "core",
+    "pain",
+    "safety"
+  ],
+  "numerator_type": "custom",
+  "numerator_columns": [
+    "SUM(p.quantity * r.strnt)/300 AS numerator, "
+  ],
+ "numerator_from": "{hscic}.normalised_prescribing_standard p  LEFT JOIN {measures}.measures.dmd_unit_dose_strength r ON CONCAT(SUBSTR(p.bnf_code,0,9),SUBSTR(p.bnf_code,-2)) = CONCAT(SUBSTR(r.bnf_code,0,9),SUBSTR(r.bnf_code,-2)) ",
```
sebbacon: you've made a new view here. It is better to include it as a subselect unless it makes it *completely* incomprehensible, otherwise views are at risk of disappearing, etc

openprescribing/measure_definitions/pregabalinddd.json
```diff
@@ -0,0 +1,52 @@
...
+  "numerator_where": [
```
sebbacon: As I understand it, each VMP can have several VTIs and therefore DDDs.  I think this means you will end up with DDDs being repeated for each BNF code and I'm not sure what the consequences would be. I would compile what this measure is doing (roughly speaking) into a single bigquery SQL and debug it there, though - hard to think about just from reading the code.

openprescribing/measure_definitions/pregabalinddd.json
```diff
@@ -0,0 +1,52 @@
...
+  "is_cost_based": false,
+  "low_is_good": null,
+  "tags": [
+    "core",
+    "pain",
+    "safety"
+  ],
+  "numerator_type": "custom",
+  "numerator_columns": [
+    "SUM(p.quantity * r.strnt)/300 AS numerator, --divide by 300 to get DDD (https://www.whocc.no/atc_ddd_index/?code=N03AX16) "
```
evansd: As we discussed on Slack, you'll need to remove the trailing comma here. Other measures _do_ have the trailing comma but they manage to work because there's some code which automatically removes them. This code isn't reliable though, so I'm removing it and we can then stick to the simple rule that `numerator_columns` should never have a trailing comma.

openprescribing/measure_definitions/pregabalinddd.json
richiecroker: It's to calculate the DDD (which is 300).  Added comment

openprescribing/measure_definitions/pregabalinddd.json
richiecroker: fixed

openprescribing/measure_definitions/pregabalinddd.json
richiecroker: when it comes to DMD calcs, for my (and possibly @brianmackenna 's ) sanity, views are really helpful in building the measures.  Not sure in this case there's a much easier way to do.

openprescribing/measure_definitions/pregabalinddd.json
richiecroker: I've taken a look at this.  @seb of course you are correct.  I have fixed by including the `ing` field in the view, which we can then use in the `WHERE` criteria of the measure along with the `bnf_code` to filter the appropriate drug.

We may find this is a eventually much better way of managing e.g. the opioids measure as we can pull the effective "dose" from DM+D, allowing new products to be automatically added.

CarolineMorton:
Can this PR be closed @richiecroker

richiecroker:
it relates to another piece of work that is ongoing, and they probably need to come together.  Can we leave open for now, it'll be the first thing I'll do if we get more OP cash.

CarolineMorton:
no problem. we are just rationalising branches on a bunch of repos

PR #2152, opened 2019-10-16
Change gabapentinoid measures to use dm+d rather than BNF code

richiecroker:


inglesp:
What's the latest on this @richiecroker?

richiecroker:
@brianmackenna can you review please.  This has been pushed to preview, and appears to give very similar results to the previous version.  This is now a "non-view" version, in line with our conversations about salbutamol.

The differences between the two measures appear to be the introduction of pregabalin _tablets_, which are not included in the original calculation view.  Therefore it would be good to get this approved as soon as possible.  Thanks!

richiecroker:
I'll also transfer this over to the pregabalin measure as well, so the same methodology is used.

richiecroker:
@brianmackenna pregabalinmg measure also now done and deployed to preview.  Same profile, very similar but some small differences due to lack of tablets in the original code

richiecroker:
When deployed to live, we also need to remove unnecessary views and tables:

- [ ] measures.dmd_unit_dose_strength_missing_fix
- [ ] measures.pregabalin_total_mg
- [ ] measures.vw__gabapentin_ddd

richiecroker:
we also need to rationalise the gabapentinoid markers, we have pregabalin (mg) AND pregabalin (items), as well as both gabapentinoids (DDD), AND percentage of pregabalin as Lyrica.  I think we can retire pregabalin (items) and Lyrica (median=0.65%).

Thoughts @brianmackenna @chrisjwood16 ?

chrisjwood16:
@richiecroker I agree, seems reasonable to retire Lyrica and pregabalin (items) measures as they don't add much any longer.

PR #2157, opened 2019-10-18
Refactor _build_measure_options

inglesp:


inglesp:
> I can't really tell from code-reading if it's all functionally the same, but I recall we have great test coverage for this?

Yes, this is thoroughly tested.

Issue #2144, opened 2019-10-11
Respimat ontology - what type of inhaler is it?


brianmackenna:
What is the product form or ontology of  _Respimat_ devices?

Background
_Respimat_ is a  type of inhaler device that has been around for approximately 10 years. If pushed I would always refer to it as a soft-mist inhaler. In the NHS dm+d it the [tiotropium respimat](https://openprescribing.net/dmd/vmp/12197411000001102/) is classified as a `pressurizedinhalation.inhalation`. 

**Why is this important?**
`pressurizedinhalation.inhalation` is generally the classification that is given to Metered Dose Inhalers (MDIs) which the NHS is trying to reduce and we have produced a new [environmental inhalers measure](https://openprescribing.net/measure/environmental_inhalers/all-england/) on. This [study](https://link.springer.com/article/10.1007%2Fs12325-019-01028-y) says respimat devices are lower polluting than MDIs and also frames them as a different type of device from an MDI. It is therefore important to correctly classify what type of inhaler they are. We should exclude them from the measure but we should first ascertain from the company if they If they are indeed a `pressurizedinhalation.inhalation` . If they aren't it makes more sense to amend dm+d and our measure and analyses by others will be refined.

brianmackenna:
I have emailed the company for clarification.

brianmackenna:
The manufacturer have responded and agree that dm+d is not completely accurate BUT `pressurizedinhalation.inhalation` is the closet within context of dm+d authoring guidelines.  

We should exclude `respimat` from the measure definition and discuss at a later date with `dm+d` team as we have not seen authoring guidelines for the `ontology` documentation.

brianmackenna:
The BSA have responded to say that any user can request a new ontology by filling out a fo[rm and emailing it to them](https://www.nhsbsa.nhs.uk/sites/default/files/2017-07/issueform.doc). This appears to differ from #2140 hospital only prescribing status where the company needs to make the change. I will fill out the form and also let the company know who appear to be under the impression that they aren't allowed request a change.

brianmackenna:
The dm+d team have been in touch and they agree with our assessment. They are going to assign the following for soft mist inhalers

- Form = Inhalation solution
- Route = Inhalation
- Ontology / Form/ Route = inhalationsolution.inhalation 


A good news story!

PR #2170, opened 2019-10-25
Update environmental_inhalers.json

brianmackenna:
excluded `Respimat` devices from the measure numerator due to them having a low effect on environment. See #2144

### DIFF
--- a/openprescribing/measure_definitions/environmental_inhalers.json
+++ b/openprescribing/measure_definitions/environmental_inhalers.json
@@ -46,10 +46,15 @@
     "  WHERE form_route IN ('pressurizedinhalation.inhalation', 'powderinhalation.inhalation')",
     "  AND bnf_code LIKE '03%' ",
     "  AND bnf_code NOT LIKE '0301011R0%' ",
+    "  AND bnf_name NOT LIKE '%Respimat%' ",
+    "  AND bnf_code NOT LIKE '0301040X0AA%' ",
+    "  AND bnf_code NOT LIKE '0301011Z0AA%' ",
+    "  AND bnf_code NOT LIKE '0301020Q0AAACAC' ",
+    "  AND bnf_code NOT LIKE '0301020Q0AAAEAE' ",
     ")"
   ],
   "date_reviewed": [
-    "2019-10-09"
+    "2019-11-14"
   ],
   "next_review": [
     "2020-10-09"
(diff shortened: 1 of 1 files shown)


brianmackenna:
No - (I think offhand) Respimat would have  impact more inline with DPIs, so it would be preferable to have people on them compared to MDIs if appropriate for the purposes of the environment.

I would be thankful for a sense check though as I am in a rush to get papers polished before leaving.

 Its not super urgent so could wait for discussion with @richiecroker or other team members next week

inglesp:
OK.  The code change is fine.  I'll wait for a nod from @richiecroker before merging though.

richiecroker:
> OK. The code change is fine. I'll wait for a nod from @richiecroker before merging though.

@inglesp can I just check my understanding of the code please?

```    "  WHERE form_route = 'pressurizedinhalation.inhalation'",
    "  AND bnf_code LIKE '03%' ",
    "  AND bnf_code NOT LIKE '0301011R0%' ",
    "  AND bnf_name NOT LIKE '%Respimat%' ",
```

means that in order to return a code it has to be both NOT `0301011R0%` AND `respimat`, of which there aren't any, so it'd return null?

Should it be:

 ```   "  WHERE form_route = 'pressurizedinhalation.inhalation'",
    "  AND bnf_code LIKE '03%' ",
    "  AND (bnf_code NOT LIKE '0301011R0%' ",
    "  OR bnf_name NOT LIKE '%Respimat%') ",
```
instead?

evansd:
> in order to return a code it has to be both NOT `0301011R0%` AND `respimat`

@richiecroker I don't think this is right. In order to return a code it has to be both NOT `0301011R0%` AND **NOT** `respimat`, doesn't it?

richiecroker:
> > in order to return a code it has to be both NOT `0301011R0%` AND `respimat`
> 
> @richiecroker I don't think this is right. In order to return a code it has to be both NOT `0301011R0%` AND **NOT** `respimat`, doesn't it?

do you know what, i'll just check it in BQ

richiecroker:
ok, that works.  Fair enough - as you were!

brianmackenna:
@richiecroker  has pointed out that `Respimat` does not exclude  generic Respimat as the are named "dev". Will update

brianmackenna:
I have manually excluded generic `Respimat`. Discussed with @richiecroker  and from our point of view it is good to go BUT  we should revisit when we are rewriting rules on new BSA data format in an attempt to future proof it.

PR #2202, opened 2019-11-10
Fix CCG name display

### LINKED ISSUE #2201: Bug in STPs - not listing CCGs correctly
https://openprescribing.net/stp/E54000028/

image

inglesp:
Fixes #2201.


evansd:
I'm going to break convention and deploy this on @inglesp's behalf as it's a straightforward fix which it would be good to have done

PR #2224, opened 2019-11-19
Add bnf_name to dmd_objs_with_form_route.sql

inglesp:


inglesp:
Yes, I mentioned that in the commit message.

 > It doesn't matter that we're renaming the `name` field as nothing uses it.

evansd:
Very sorry! As it was so small I didn't review commit by commit so I missed it

PR #2232, opened 2019-11-20
Don't build changed measures on deploy

inglesp:
This matches the version of the fabfile that ebmbot has

We should revisit this, and either:

* work out why it wasn't working and fix it,
* or make deploying measures something you do via Slack


sebbacon:
It does make me sad that we've not been able to fix this, because it used to work, and it's useful. Perhaps one to pair on one Wednesday?

PR #2233, opened 2019-11-21
Semaglutide

richiecroker:


brianmackenna:
Looks ok to me unless we need to reconfigure to suit new measure structures? @inglesp

inglesp:
> Looks ok to me unless we need to reconfigure to suit new measure structures?

No, because it's not a vanilla measure -- it queries the raw prescribing data.

PR #2234, opened 2019-11-21
Mark some measures as having no analyse_url

inglesp:
Because of non-standard numerators, certain measures cannot be
adequately represented via the analyse page:

* bdzper1000 -- numerator_columns: SUM(p.quantity * r.percent_of_adq)
* gabapentinoidsddd -- numerator_columns: SUM(gaba_ddd)
* ktt9_uti_antibiotics -- numerator_columns: SUM(p.quantity * r.adq_per_quantity)
* pregabalinmg -- numerator_columns: SUM(lyrica_mg) AS numerator"
* sildenafil -- numerator_where returns different things depending on the month

### DIFF
--- a/openprescribing/frontend/management/commands/import_measures.py
+++ b/openprescribing/frontend/management/commands/import_measures.py
@@ -326,7 +326,8 @@ def create_or_update_measure(measure_def, end_date):
             get_num_or_denom_bnf_codes(measure, num_or_denom, end_date),
         )
 
-    measure.analyse_url = build_analyse_url(measure)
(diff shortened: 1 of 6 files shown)


inglesp:
> Looks fine as is, but wouldn't the logic be better inside `build_analyse_url`...?

Maybe, but then we'd have to pass `v.get("no_analyse_url")` to `build_analyse_url` and I don't think that'd be much nicer.

evansd:
Oh, my bad. I didn't realise it was only a property on `v` and not something on the `measure` which we were already passing in. As you were...

PR #2242, opened 2019-11-25
Add task to create public measure tables in BQ

inglesp:
The tables created are already in the `public_draft` dataset.

### DIFF
--- a/openprescribing/bq_public_tables/_measure_template.sql
+++ b/openprescribing/bq_public_tables/_measure_template.sql
@@ -1,0 +1,54 @@
+WITH practice_numerator AS (
+{numerator_sql}
+),
+
+practice_denominator AS (
+{denominator_sql}
+),
+
+month_practice AS (
+    SELECT
+        m.month,
+        p.id AS practice_id
+    FROM public_draft.practice AS p
+    CROSS JOIN (
+        SELECT DISTINCT month FROM practice_numerator
+        UNION DISTINCT
+        SELECT DISTINCT month FROM practice_denominator
+        ORDER BY month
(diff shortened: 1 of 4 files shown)


inglesp:
> Is there an argument for making these views? Mainly pedagogic - users can see how they have been constructed

I think that the SQL is too overwhelming to be much use pedagogically.  Better is to document carefully, like [this](https://openprescribing.net/docs/querying-the-raw-data-yourself/).

Also, it makes querying the table significantly slower.

inglesp:
> Have you already sanity checked that this produces the same output as our main code 

Yes, it does, more or less.

The difference is when a practice has zero for both its numerator and its denominator in a given month.  In OpenPrescribing, we treat set `calc_value` and `percentile` to `NULL`.  Here, we set `ratio` and `percentile` to `0`.

This makes more sense in some situations (eg if a practice prescribes no Levetiracetam, then it makes sense to say that its ratio of Keppra to Levetiracetam is zero)...

...and less in others (eg if a practice is closed and has no prescribing and no patients, it makes sense to say that doesn't have a ratio.

Another effect on this change is shift, slightly, the percentile values for all practices.  That is: the OpenPrescribing method calculates the percentiles over all practices with non-`NULL` ratios, while this method calculates the percentiles over all practices.  This means that percentile values here differ slightly from those on the website.

Thoughts?

sebbacon:
OK - I suggest documenting this in the excellent documentation already produced. i.e. "you can query measure data here but note it differs very slightly from the data on openprescribing in the following ways"?

inglesp:
Have updated the docs.

PR #2261, opened 2019-11-26
Expunge allauth

inglesp:
This is the first part of reworking the alert signup process.

There are a handful of user-facing changes in this PR:

* Users are no longer signed in when they manage their bookmarks.  Since that was the only way to sign in, we have removed everything where behaviour depends on the user being signed in.  Most users will not notice a difference.
* Alerts will start to be sent to users who have not verified their email addresses.  (This used to be a requirement but was dropped in January 2019.)  Before merging, we must decide what to do about bookmarks that haven't been approved.  There are about 1000 such bookmarks, created between December 2016 and January 2019.  I would suggest dropping these from the database.
* Email messages confirming alert signup are now tagged with `"alert_signup"` rather than `"allauth"`.  This is presumably something that we can (but don't?) use in the Mailgun interface.


inglesp:
Closing this, as more to follow soon once I've added alert signups for PCNs.

PR #2271, opened 2019-11-27
Rework alert signups

inglesp:
This PR reworks both the workflow of how users sign up to alerts, and how the signup process is handled by the code.

The workflow is now much simpler: from a page with an alert signup form, a user enters their email address and submits the form.  (They no longer have to choose whether to sign up for the newsletter.)  This creates a bookmark object in our database (meaning they will be emailed next time we send alerts).  They are sent an email telling them that they have signed up for alerts, but they don't have to confirm anything (existing behaviour) and are not told they have to confirm something (new behaviour!).  They see something like:

image

The link to "signed up for our newsletter" goes to the [contact page](https://openprescribing.net/contact/).

The code is also much simpler, and we have dropped our dependencies on `allauth` and `mailchimp3`.  This should make further tweaks to the signup process much less scary.

As part of the changes we have removed the `approved` field from the bookmark models.  Previously, we've only sent alerts to users who've approved their bookmarks, and there are about 1000 bookmarks that haven't been approved.  I think we should delete these from the database, but I'd like to get a second opinion!


sebbacon:
To clarify (not looking at the code til tomorrow sorry) does this mean anyone can now sign anyone else up to emails?

inglesp:
> To clarify (not looking at the code til tomorrow sorry) does this mean anyone can now sign anyone else up to emails?

They already can.

sebbacon:
Sorry I am somewhat confused! I thought people previously only got emails when they confirmed / activated their account.

inglesp:
> I thought people previously only got emails when they confirmed / activated their account.

We stopped doing that in January in #1397 (Automatically approve alert signups).

openprescribing/frontend/views/views.py
```diff
@@ -1062,6 +983,171 @@ def _financial_ytd_total(monthly_totals):
...
+    or an IntegrityError if the submitted pct_id/practice_id/pcn_id doesn't
+    correspond to an existing PCT/Practice/PCN (which shouldn't happen because
+    the entity id should be set in the initial form by _build_bookmark_form().)
+    """
+
+    form_cls = BOOKMARK_CLS_TO_FORM_CLS[bookmark_cls]
+    form = form_cls(request.POST)
+    form.full_clean()
+    email = form.cleaned_data["email"]
+    user, _ = User.objects.get_or_create(username=email, defaults={"email": email})
```
evansd: I think we should either use a case-insensitive match here or, more simply, force the email to lowercase before doing anything else with it. Either way, it would be better not to end up with duplicate users because of inconsistencies in email case.

inglesp: Good idea.

We have 37 users whose email address is not lowercase.  Should we convert those to lowercase?  And if so, should we use a data migration or do it with a one-off script?

evansd: I think we should convert the existing users to lowercase using a one off script. Doing it with a data migration is just extra work for no real benefit.

inglesp: Any reason not to do this:

```
from django.contrib.auth.models import User

for user in User.objects.all():
    user.username = user.username.lower()
    user.email = user.email.lower()
    user.save()
```

?

evansd: Not that I can think of!

PR #2274, opened 2019-11-28
Show list of PCNs with at least one practice in CCG

inglesp:


openprescribing/frontend/models.py
```diff
@@ -208,6 +208,11 @@ def name_and_status(self):
     def get_absolute_url(self):
         return reverse("ccg_home_page", kwargs={"ccg_code": self.code})
 
+    def pcns(self):
+        "Return all PCNs with at least one practice in this CCG."
+         
+        return PCN.objects.filter(practice__in=self.practice_set.all())
```
evansd: I'm sure this is probably fine as is, but I think you could also do `.filter(practice__ccg_id=self.code)` and get a more efficient query.

inglesp: Good point, but unfortunately they're both wrong.

inglesp: (They both return one PCT per practice in the CCG.)

PR #2277, opened 2019-11-28
Remove (most) versions from requirements.in

inglesp:
They're no longer necessary.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -1,50 +1,47 @@
 # This file lists all the packages that openprescribing depends on directly.
-# See instructions at the top of base.txt for how to generate a full list of
-# pinned requirements.
-
-# Note that most of the pinned version numbers here are not essential (they
-# were taken from the versions running in production at the point that this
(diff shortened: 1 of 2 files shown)


inglesp:
Hm... I think we should record that we're deliberately running old versions of these libraries somewhere.  Where would you suggest?

evansd:
We record it in dependabot, when we tell to ignore certain versions. It's a shame that that information is then locked away in a proprietary tool, but unless we're militant about keeping `requirements.in` in sync with dependabot (which I suspect we won't be) then I think it's better not to have it in there at all.

evansd:
P.S I have very little attachment to this opinion so do feel free to ignore me

inglesp:
It's not obvious how to get this information out of dependabot.  (I think you have to look for PRs with comments saying "OK I won't merge that" or something.)

But I don't have very strong opinions either so I'm going to merge this and move on.

PR #2283, opened 2019-12-02
improved reference to BTS/SIGN for high dose definition

richiecroker:


inglesp:
This is now on staging at http://staging.openprescribing.net/measure/icsdose/ and could be merged when @brianmackenna approves it.

PR #2286, opened 2019-12-03
Prepare for django upgrade

inglesp:


inglesp:
Closing in favour of #2289.  Turning `DeprecationWarning`s into errors doesn't really work, because we only want to catch `RemovedInDjango20` warnings and you can't do that.  (https://bugs.python.org/issue22543)

PR #2294, opened 2019-12-05
Fix "PCN" where we meant "CCG"

### LINKED ISSUE #2292: Oragnisation Hierachy bug PCN ->CCG
On an STP page ([south east London example](https://openprescribing.net/stp/E54000030/)) the STP says there are "x PCNs in this STP". This should read CCG for hierachy, however when you click "show more" it does correctly show CCGs so it is just text that needs amending. 

image

evansd:
Closes #2292


inglesp:
Merging this now to give @ebmbot a kick.

PR #2295, opened 2019-12-05
Bump django 2.1

inglesp:


inglesp:
This was completely painless.  Suggest we let 2.0 bed in for a couple of days before merging early next week.

PR #2296, opened 2019-12-05
Move to using new URL syntax

inglesp:


evansd:
Looks good. In general the `path` syntax is so much more readable than regexes.

If we are missing some `get_object_or_404`s calls then at worst there'll be some 500s that should have been 404s, which we'll be notified about it any case.

If we're really cared about stricter validation in the URL definitions it would be easy enough to define a [custom path converter](https://docs.djangoproject.com/en/2.1/topics/http/urls/#registering-custom-path-converters) for e.g ODS codes. But I don't think there's much advantage to that.

Incidentally, I have an aesthetic objection to uppercase in URLs, and a technical objection to case-sensitivity. So when we come to reworking the organisation code I'd quite like to make the code matching case insensitive and lowercase by default.

inglesp:
> If we're really cared about stricter validation in the URL definitions it would be easy enough to define a custom path converter for e.g ODS codes. But I don't think there's much advantage to that.

Thought about this but felt like excessive work.

> Incidentally, I have an aesthetic objection to uppercase in URLs, and a technical objection to case-sensitivity.

Me too!

PR #2303, opened 2019-12-09
Fix CSV download URL on numerator breakdown page

evansd:
Previously, the URLs were built seperately for the DataTable request and
for the download link. This made it easy to omit the `org_type`
parameter from the download link. In addition, there is some backwards
compatibility code which makes the org type optional and, if it is
omitted, tries to guess the type from the length of the code. So in
fact, the download link worked OK for practices and CCGs which is
probably why we never noticed the issue before. The addition of PCNs
made the problem more obvious.

The fix here is to generate the URL once, inside the view, and then use
it both places.

### DIFF
--- a/openprescribing/frontend/views/views.py
+++ b/openprescribing/frontend/views/views.py
@@ -455,7 +455,8 @@ def measure_for_one_entity(request, measure, entity_code, entity_type):
         "measure": measure,
         "measure_options": measure_options,
         "current_at": ImportLog.objects.latest_in_category("prescribing").current_at,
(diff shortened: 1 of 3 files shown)


evansd:
Good point. I hadn't even clocked that I was getting rid of it (hence code reviews!) and I've added it back now.

inglesp:
Have just remembered there's `_build_api_url()` in `views.py` that does just what you want.

evansd:
So there is. Thanks!

PR #2308, opened 2019-12-10
Call create_normalised_prescribing_view directly in e2e setup

inglesp:
Since 0375e05b, create_normalised_prescribing_view is not called as part
of generate_presentation_replacements.  And I'm fairly sure that the
only reason we called generate_presentation_replacements in the e2e
setup was to ensure that normalised_prescribing_* existed.


inglesp:
I think it's me who keeps breaking them!  But thanks :)

PR #2333, opened 2019-12-18
Bump django 2.2

inglesp:


inglesp:
Again, let's leave 2.1 a couple of days to bed in.

PR #2360, opened 2020-01-09
Improving unsubscribe wording

HelenCEBM:
Also adding explanation of how to sign up for more alerts.


HelenCEBM:
Mainly I wanted to make it more clear that the link lets you choose which alerts to remove and it has a confirmation step, rather than immediately unsubscribing you as soon as you click it.  Feel free to change wording!

PR #2363, opened 2020-01-09
Add measure definition pages with SQL

### LINKED ISSUE #2258: Add links, for all measures, to data / code / text explaining how measure was built
We have this done for all that can be driven from analyse page; but those relying on e.g. quantity cannot be done this way. For those we could swiftly zip thru and manually make a link to the code driving the measure, not necessary for it to be perfect text, just discoverable and intelligible to an expert crowd.

evansd:
Closes #2258

### DIFF
--- a/openprescribing/frontend/templatetags/template_extras.py
+++ b/openprescribing/frontend/templatetags/template_extras.py
@@ -95,3 +95,8 @@ def fancy_join(lst, sep=", ", final_sep=" and "):
         head, tail = lst[:-1], lst[-1]
         lst = [sep.join(head), tail]
     return final_sep.join(lst)
+
+
+@register.filter
+def username_from_email(email):
+    return email.split("@")[0]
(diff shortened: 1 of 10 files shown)


sebbacon:
Thanks, now I see it on the website, some small suggestions:

Inconsistent full stop:
image

---

We should consider stripping out `WHERE 1=1` as this is just an artefact of the implementation and hinders comprehension:
image

---

> Below are the database queries which are used to create this measure. These are run in Google BigQuery against our copy of the public prescribing dataset. 

"These are run against a copy of the BSA prescribing data [link to About] which we store in Google BigQuery"?

> We're working on making our BigQuery tables publically available 

publicly 

> at which point it will be possible to run and modify these queries yourself. But even where code and database queries are not directly useable by others we believe it is always preferrable to make them public. 

preferable

evansd:
Feedback from @HelenCEBM (copied from Slack):

 * should there be a link from the "definition" page to "see the full definition in github"?

* should there be a caveat to explain for things like high dose opioids that we rely on drug lists defined in additional tables and "we're working on making these available" or something

 * could also be useful to include the "last reviewed" date or similar?

PR #2367, opened 2020-01-13
Update denominator

HelenCEBM:
... to account for the new measures added June 2017 (summed population denominator needs to be divided by 25 not 18).

Also adds a test to ensure that the divisor stays correct in future.

### DIFF
--- a/openprescribing/frontend/tests/commands/test_import_measures.py
+++ b/openprescribing/frontend/tests/commands/test_import_measures.py
@@ -2,7 +2,9 @@ from __future__ import print_function
 
 import csv
 import itertools
+import json
 import os
+import re
 import tempfile
 from mock import patch
 from random import Random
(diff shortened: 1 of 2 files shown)


richiecroker:
Are we able to feed `measure_count` into the code, rather than hard-coding `25`, so that if we make further changes to the measures, this will be automatically reflected in the denominator?

evansd:
> Are we able to feed measure_count into the code, rather than hard-coding 25, so that if we make further changes to the measures, this will be automatically reflected in the denominator?

Not easily, unfortunately. But I've done the next best thing which is to create a test that the numbers match which will prevent us from deploying if we forget to update them.


> Should we also have a test that measure_count is the same as the count of measure names prefixed lp (excluding lpz)

Excellent idea! Done.

PR #2369, opened 2020-01-14
Change measure to reflect quantity and capsules only

richiecroker:
Amended measure definition:
1)  Change items to quantity
2)  Restrict measure to capsules only


richiecroker:
@brianmackenna are you looking at correct URL: http://staging.openprescribing.net/measure/lyrica/ is correct one

PR #2378, opened 2020-01-15
Brian vitb check

brianmackenna:
@evansd  : the master branch created by @richiecroker  is ok for deployment. I created this branch for merging as it has my check on it. 

Greatful for ideas on how to streamline pull requests if I am messing up workflow


evansd:
Thanks @brianmackenna. When I edit a file through the GitHub interface I get the option to "Commit directly to this branch" as well the option to create a new branch and a PR. That's the one to use in cases like this, but it might be that you don't have that due to permissions so let me know if you can't see it next time.

PR #2394, opened 2020-01-20
MatrixStore Powered PPU

### LINKED ISSUE #2411: PPU substitutions code doesn't work as intended
I'm documenting this just to confirm my understanding. The issue, assuming I've understood it correctly, will be fixed as part of the MatrixStore-powered reimplementation in #2394 

Can others check over and see if the below makes sense?

---------------------

We have a [formulation swaps](https://github.com/ebmdatalab/price-per-dose/issues/11) spreadsheet which we use to mark different formulations of a given presentation (e.g. tablets and capsules) as equivalent for the purposes of the PPU calculation. But there are two issues with the way this is currently implemented.

### 1. We don't handle branded prescribing correctly

The formulation swaps are implemented by creating a `prescribing_with_merged_codes` table which performs a substitution on BNF codes using a `CASE` statement. For instance, if the spreadsheet says that generic A can be substituted for generic B then whenever we see the BNF code B we replace it with A. Later we apply various BNF code matching rules to also include all prescribing of branded A. But we never include branded B.

Interestingly the code which renders the bubble charts, which are driven directly from the prescribing data and not from the PPU table, _does_ seems to handle substitution of branded prescribing correctly. But not the code which calculates the target PPU or the available savings.

As a concrete example, if you look at savings for Co-codamol 30mg/500mg tablets at [this CCG](https://openprescribing.net/ccg/15E/price_per_unit/?date=2019-11-01) it shows a total quantity of 2,318,288. If you look at what the [bubble chart API](https://openprescribing.net/api/1.0/bubble/?format=json&date=2019-11-01&bnf_code=0407010F0AAAHAH&highlight=15E&focus=1) returns and add up the quantities you get a figure which is greater by 34,952. The difference is exactly equal to the quantity of branded capsules (e.g. "Kapake_Cap 30mg/500mg"). So the PPU figure is including generic tablets, branded tablets, generic capsules but not branded capsules.

This omission can sometimes have quite large effects on the target PPU. For instance, on [this chart](https://openprescribing.net/all-england/0301020Q0AAABAB/price-per-unit/?date=2019-11-01) it's fairly obvious from the big green blob of "Braltus_Pdr For Inh Cap 10mcg+Zonda Inh" priced at 86p that the best decile price should be around 86p. But looking at the Target PPU column in the table we can see that it's been calculated at £1.03.

### 2. We don't handle multiple formulation swaps correctly
 
Sometimes we want to say that not only is A equivalent to B, but C is also. For instance, on the last two lines of the current spreadsheet we have:

Formulation | Code | Alternative Formulation | Alternative Code
-- | -- | -- | --
Tiotropium_Pdr For Inh Cap 18mcg | 0301020Q0AAABAB | Tiotropium_Pdr For Inh Cap 10mcg + Dev | 0301020Q0AAADAD
Tiotropium_Pdr For Inh Cap 18mcg + Dev | 0301020Q0AAAAAA | Tiotropium_Pdr For Inh Cap 10mcg + Dev | 0301020Q0AAADAD

But the PPU code can't handle the same presentation occurring multiple times in the spreadsheet i.e. it can't handle saying as we do above "A is equivalent to B and C is also equivalent to B". If it encounters the same presentation a second time it just ignores it:
https://github.com/ebmdatalab/openprescribing/blob/e7104cf8305793172cd82989dd0b70471f247613/openprescribing/frontend/management/commands/import_ppu_savings.py#L53-L58

evansd:
This PR replaces the existing PPU implementation with one driven from the MatrixStore. It's a biggie, and would definitely benefit from a commit-by-commit review (using the commits tab to get the logical ordering).

This code gives savings figures that are reasonably close to the previous implementation but with a number of deliberate changes. In particular it fixes #2411 which finds some additional savings. There are also changes to the way the minimum thresholds are calculated (see commit message for explanation).

I had originally hoped that we could drop the prescribing table after this, however I've since realised that the ghost branded generics query uses it so there'll be another piece of work in replacing that. Hopefully nowhere near as involved as this one though.


evansd:
I've got this to the point where I can reproduce the PPU savings calculated under the current system with two caveats:

 1. There's disagreement within the codebase as to whether a CCG's numbers should include its non-setting 4 practices. There aren't many of these (for the most part, if a practice belongs to a CCG then it's setting 4) but they do exist. Historically, the Analyse form has included practices of all settings so the MatrixStore copied this. The PPU calculation includes only setting 4 practices.

 2. We impose a minimum savings threshold on PPU savings. If the available savings on a class of drug is less than £200 we don't include it in the database at all. This means that when we aggregate the CCG figures up to the All England level a proportion of the CCGs are missing entirely. This doesn't have much effect on the total savings (necessarily, these are CCGs with little available savings) but it _does_ effect the mean PPU and total quantity calculations. The mean PPU ends up too high because we're excluding all the best performing CCGs. And the total quantity ends up too low.

I think the answer to 1 is to decide on the most generally appropriate definition of "all practices in a CCG" and use that consistently everywhere, accepting that some of our numbers will change slightly.

I think the answer to 2 is that the way the new MatrixStore code does the calculation is the correct way and that are existing numbers are wrong, albeit probably not in a way that anyone cares too much about.

evansd:
I have since discovered a third discrepancy:

3. We apply a minimum savings threshold of £10 when calculating savings for practices (i.e. we treat as zero any savings less than this amount), but we don't apply this threshold when we sum up practice savings to get CCG savings. This leads to a small but slightly confusing discrepancy between e.g the saving for Fluticasone at the top of the table on this [CCG page](https://openprescribing.net/ccg/15E/price_per_unit/?date=2019-11-01) and the saving shown when you click through to the [practice breakdown](https://openprescribing.net/ccg/15E/0302000N0AABGBG/price_per_unit/?date=2019-11-01).

evansd:
There's a feature called `trim` in the current PPU bubble chart API which is designed to remove very high percentile PPU values:
https://github.com/ebmdatalab/openprescribing/blob/debb386b32e2b5ee56f17ad23fe24f51caa119aa/openprescribing/api/views_spending.py#L117-L135

It doesn't seem to have been used at all as for as long as we have logs:
```bash
ssh largeweb2.ebmdatalab.net \
  zgrep 'trim=' /webapps/openprescribing/logs/nginx-access.log*
``` 

Can I just not implement this? @sebbacon @inglesp

inglesp:
> Can I just not implement this?

Fine by me.

A couple of out of interests:

* is it not implemented in the UI, or is it a UI thing that nobody's using?
* what would it take to implement something like this with the MatrixStore?

evansd:
Good questions.

> is it not implemented in the UI, or is it a UI thing that nobody's using?

It's not implemented in the UI; you have to supply a special query parameter to the PPU page which then gets included in the URL for the API call.

> what would it take to implement something like this with the MatrixStore?

I don't think it would be any more difficult than the other stuff I've already implemented, so it's not a problem if we decide we do need it. But I thought this would be a good opportunity to remove unused code.

inglesp:
image

evansd:
:smile:  Literally every commit on this branch is "WIP" though so I'd say I've maintained a steady information level of precisely zero.

openprescribing/frontend/price_per_unit/savings.py
```diff
@@ -0,0 +1,144 @@
...
+        ),
+        bnf_codes,
+    )
+
+    quantity_sum = MatrixSum()
+    net_cost_sum = MatrixSum()
+    for quantity, net_cost in results:
+        quantity_sum.add(get_submatrix(quantity, cols=date_slice))
+        net_cost_sum.add(get_submatrix(net_cost, cols=date_slice))
+    return quantity_sum.value(), net_cost_sum.value()
```
inglesp: We have almost identical code in `prescribing_breakdown.get_prescribing`.

openprescribing/api/views_spending.py
```diff
@@ -198,6 +225,34 @@ def price_per_unit(request, format=None):
...
 
 
+def _get_concession_bnf_codes(date):
+    """
+    Return list of BNF codes to which NCSO price concessions were applied in
+    the given month.
+
+    Note that because concessions are defined at the product-pack level and BNF
+    codes refer to products it's possible that the same BNF code may be
+    returned multiple times.
```
inglesp: Why not add a `SELECT DISTINCT`?  Also, why not use the ORM?

openprescribing/api/views_spending.py
```diff
@@ -294,50 +349,6 @@ def ghost_generics(request, format=None):
...
-          MODE() WITHIN GROUP (ORDER BY price_concession)
-            AS price_concession,
-          MODE() WITHIN GROUP (ORDER BY name) AS name
-        FROM cte
-        GROUP BY date, presentation
-        """.format(
-        original_sql=original_sql,
-        pct_name=entity_name if entity_type == "CCG" else "NULL",
-        practice_name="NULL" if entity_type == "CCG" else entity_name,
-    )
```
inglesp: Goodbye SQL I've never understood!

openprescribing/frontend/price_per_unit/savings.py
```diff
@@ -89,57 +100,83 @@ def get_total_savings_for_org(date, org_type, org_id):
...
-def get_total_savings_for_org_type(date, org_type):
+def get_total_savings_for_org_type(
+    db,
+    substitution_sets,
+    date,
+    group_by_org,
+    min_saving,
+    practice_group_by_org,
+    target_centile,
+    practice_min_saving,
```
inglesp: Are these last four params really "global state"?

openprescribing/frontend/price_per_unit/substitution_sets.py
```diff
@@ -46,6 +47,28 @@ def __init__(self, id, presentations, name=None, formulation_swaps=None):
...
         # formulation changes are involved then this is None.
         self.formulation_swaps = formulation_swaps
+        # `cache_key` is used to identify the state of this SubstitutionSet for
+        # caching purposes i.e.  SubstitutionSet instances should have the same
+        # cache_key if and only if they have same list of presentations
+        hashobj = hashlib.md5(str(self.presentations).encode("utf8"))
+        self.cache_key = hashobj.digest()
+
+
+class DictWithCacheID(dict):
```
inglesp: For once, I think this is an acceptable example of subclassing `dict`!

openprescribing/openprescribing/settings/base.py
```diff
@@ -330,25 +330,51 @@
...
+        # from blocking each other (readers are never blocked). This involves
+        # splitting the cache into multiple shards (8 by default).  However
+        # this seems to have a noticeable impact on performance and given that
+        # our workload is extremely light on writes (the first few requests
+        # after a new data load will do a whole batch of writes and then
+        # nothing) we don't really need this. Setting the shard count to 1
+        # effectively disables the fan-out behaviour. See:
+        # http://www.grantjenks.com/docs/diskcache/tutorial.html#fanoutcache
+        "SHARDS": 1,
+        # In the event that a write does get blocked this timeout etermines how
```
inglesp: s/etermines/determines/

openprescribing/matrixstore/cachelib.py
```diff
@@ -0,0 +1,60 @@
...
+            cache_key = _get_cache_key(cache_key_base, args, kwargs)
+            result = cache.get(cache_key, default=MISSING)
+            if result is MISSING:
+                result = func(*args, **kwargs)
+                cache.set(cache_key, result)
+            return result
+
+        return wrapper
+
+    return decorator
```
inglesp: Apologies if I've missed something in the PR -- are there tests for this/for the diskcache?

openprescribing/frontend/tests/price_per_unit/test_savings.py
```diff
@@ -0,0 +1,320 @@
...
+
+        self.assertEqual(round_floats(result), round_floats(expected))
+
+    @classmethod
+    def tearDownClass(cls):
+        cls._remove_patch()
+        super().tearDownClass()
+
+
+def invent_generic_bnf_code(index):
```
inglesp: Does this live in a data factory?

openprescribing/frontend/tests/price_per_unit/test_savings.py
```diff
@@ -0,0 +1,320 @@
...
+        factory.create_months("2020-01-01", 2)
+        factory.create_practices(10)
+        for bnf_codes in substitution_sets:
+            for bnf_code in bnf_codes:
+                factory.create_presentation(bnf_code=bnf_code)
+        factory.create_prescribing(
+            factory.presentations, factory.practices, factory.months
+        )
+
+        # Copy these objects to the database
```
inglesp: I don't understand this comment.

openprescribing/frontend/price_per_unit/substitution_sets.py
```diff
@@ -0,0 +1,274 @@
...
+            group.extend(other_group)
+            for member in other_group:
+                groups[member] = group
+    for element, group in groups.items():
+        # Each group will occur multiple times in this loop, once for each of
+        # its elements. But we want to return just the unique groups, so we
+        # pick (arbitrarily) the first element of each group to identify it
+        # with.
+        if element == group[0]:
+            yield group
```
inglesp: This doesn't appear to be independently tested -- it's a little hairy, so could it be?

openprescribing/frontend/price_per_unit/substitution_sets.py
```diff
@@ -0,0 +1,274 @@
...
+    names = Presentation.names_for_bnf_codes(list(bnf_codes))
+    # Add in names for our invented BNF codes which represent their entire BNF
+    # sub-paragraph
+    for sub_para, name in GENERIC_SUB_PARAGRAPHS.items():
+        names[sub_para + "AAA0A0"] = name
+    return names
+
+
+def generic_equivalent_for_bnf_code(bnf_code):
+    """
```
sebbacon: Possibly some opportunity for reusing this in the new analyse form stuff you're working on @inglesp? If so perhaps belongs in a more-generic module?

openprescribing/frontend/price_per_unit/savings.py
```diff
@@ -0,0 +1,144 @@
...
+# Note this is "standard_practice" rather than "practice" as we only want to
+# include setting 4 practices (i.e. ordinary GP practices)
+CONFIG_TARGET_PEER_GROUP = "standard_practice"
+
+# Savings are calculated in two stages: first at the practice level, and then
+# aggregated up into organisation savings. At both stages we apply a minimum
+# threshold such that savings below that threshold are treated as zero. This is
+# particularly important at practice level as it prevents "negative savings"
+# from practices already performing better than the target PPU.
+# (Note: all values in pence)
```
sebbacon: From this comment alone, it's hard to understand what is meant here.

The rationale for treating small savings as zero is (IIRC) that trivial savings are unlikely ever to be worth making.

The negative savings thing is handled separately and not really relevant to this config variable, I think?

It would also be useful to be clear what the `all_standard_practices` option means, i.e. savings under this at all-England level are considered trivial (right?)

openprescribing/openprescribing/settings/base.py
```diff
@@ -330,25 +330,51 @@
...
+        # long to wait before giving up and just not writing the value to the
+        # cache. It's by no means a disaster if this happens: we'll just cache
+        # the value the next time round.
+        "DATABASE_TIMEOUT": 0.1,
+        "OPTIONS": {
+            "size_limit": cache_size_limit,
+            # By default DiskCache writes larger values to individual files on
+            # disk and just uses SQLite for the index and to store smaller
+            # values. However by bumping the below limit up to something huge
+            # we can get it to store values of any size in SQLite. By also
```
sebbacon: Why is this a good thing?

openprescribing/pipeline/management/commands/run_pipeline_e2e_tests.py
```diff
@@ -68,7 +68,6 @@ def run_end_to_end():
     client.create_table("ccgs", schemas.CCG_SCHEMA)
     client.create_table("stps", schemas.STP_SCHEMA)
     client.create_table("regional_teams", schemas.REGIONAL_TEAM_SCHEMA)
-    client.create_table("ppu_savings", schemas.PPU_SAVING_SCHEMA)
```
sebbacon: Just to sanity check that none of our colleagues mind this disappearing...?

openprescribing/frontend/price_per_unit/profile.py
```diff
@@ -0,0 +1,37 @@
...
+    attempts.sort()
+    profile_file = "profile.{}.prof".format(
+        datetime.datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
+    )
+    attempts[0][1].dump_stats(profile_file)
+    print(
+        "{}s (best of {}), profile saved as: {}".format(
+            attempts[0][0], num_attempts, profile_file
+        )
+    )
```
sebbacon: I think we can probably drop this? It's neat but I bet we never use it again

openprescribing/frontend/tests/price_per_unit/test_savings.py
```diff
@@ -0,0 +1,320 @@
...
+                bnf_code=presentation["bnf_code"], name=presentation["name"]
+            )
+
+        cls.substitution_sets = substitution_sets
+        cls.factory = factory
+        cls._remove_patch = patch_global_matrixstore(
+            matrixstore_from_data_factory(factory)
+        )
+        # Clear the cache on this memoized function
+        get_substitution_sets.cache_clear()
```
sebbacon: Why is the necessary with a DummyCache backend?

openprescribing/frontend/tests/price_per_unit/test_savings.py
```diff
@@ -0,0 +1,320 @@
...
+    @classmethod
+    def setUpTestData(cls):
+        # Create some sets of substitutable presentations, which each consist
+        # of a generic plus some branded equivalents
+        substitution_sets = []
+        for i in range(4):
+            generic_code = invent_generic_bnf_code(i)
+            brands = invent_brands_from_generic_bnf_code(generic_code)
+            substitution_sets.append([generic_code] + brands)
+
```
sebbacon: Can we test exclusions and manual mappings, too?

openprescribing/frontend/price_per_unit/substitution_sets.py
```diff
@@ -0,0 +1,274 @@
...
+class SubstitutionSet:
+    """
+    Represents a set of presentations which we believe can be reasonably
+    substituted for one another
+    """
+
+    def __init__(self, id, presentations, name=None, formulation_swaps=None):
+        # The code which identifies this set of substitutions. As it happens we
+        # define this as the lexically smallest generic BNF code in the set but the
+        # rest of the codebase just treats this as an opaque string identifier.
```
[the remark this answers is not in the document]
evansd: I'm not sure it is. The point of the comment was just to say that's how we happen to do it, but that nothing in particular hangs on it. Unless I've misunderstood what you were thinking of enforcing?

openprescribing/frontend/price_per_unit/substitution_sets.py
```diff
@@ -0,0 +1,274 @@
...
+        formulations[code] = row["Formulation"].strip()
+        formulations[alternative_code] = row["Alternative formulation"].strip()
+    # We combine pairs of equivalent codes into groups of equivalent codes.
+    # This means if our substitutions table says "A and B are equivalent" and
+    # "B and C are equivalent" then we can handle this correctly.
+    for code_group in groups_from_pairs(code_pairs):
+        # Get the unique formulations involved in this group
+        formulations_for_group = [
+            f for f in set(map(formulations.get, code_group)) if f
+        ]
```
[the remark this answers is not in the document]
evansd: Almost but not quite. We need to know not just that the the code is in the `formulations` dict but that it isn't an empty string. You have nevertheless inspired me to simplify this line though.

openprescribing/matrixstore/sql_functions.py
```diff
@@ -5,13 +5,20 @@
...
     accumulator = None
 
     def step(self, value):
-        if value is None:
-            return
-        matrix = deserialize(value)
+        if value is not None:
+            self.add(deserialize(value))
+
+    def add(self, matrix):
```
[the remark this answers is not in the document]
evansd: This did cross my mind. But I feel like it might end up as a case of syntactic sugar rotting semantic teeth, as the `MatrixSum` object doesn't otherwise behave like a matrix. In particular, you have to call `value()` on it to get a result back.

openprescribing/frontend/price_per_unit/prescribing_breakdown.py
```diff
@@ -0,0 +1,97 @@
...
+          bnf_code, quantity, net_cost
+        FROM
+          presentation
+        WHERE
+          bnf_code IN ({})
+        """.format(
+            ",".join("?" * len(bnf_codes))
+        ),
+        bnf_codes,
+    )
```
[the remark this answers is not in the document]
evansd: Yes, I agree. And also with the comment below about almost identical queries appearing elsewhere. There's also very similar code powering the spending and NCSO APIs. I was holding off factoring these away though until I had a better sense of the different use cases we would have for it. I think once I've done Ghost Branded Generics (which is the last bit of code using `frontend_prescribing` I'll have a firm idea and it will be a relatively easy job to replace this code with a function in the matrixstore library.

openprescribing/frontend/price_per_unit/savings.py
evansd: No, they're not. They're more like "configuration values which affect the return value of the function" but I didn't have a convenient name for that. Maybe just "global state and configuration" would do actually.

openprescribing/frontend/tests/price_per_unit/test_savings.py
evansd: I've expanded this comment:
```
# The DataFactory creates data that can be written to the MatrixStore
# but doesn't automatically create anything in the database, so we do
# that manually here
```
But maybe the route of the confusion is that we've ended up with two DataFactory classes: one that lives in `matrixstore.tests` (which deliberately avoids writing to the database) and another that lives in `frontend.tests` which automatically populates the database.

It would be good to unify these, but it's a slightly tricky and unrewarding job I fear.

openprescribing/frontend/tests/price_per_unit/test_savings.py
evansd: Probably. But then it should really use `next_id()` rather than requiring an index passed in manually. And then it would need to deal with numbers larger than 9 and it didn't feel worth it until we had an actual need to reuse it.

openprescribing/api/views_spending.py
evansd: Because I copied it from the big fat query in `spending_utils.py` and the deleted the stuff I didn't need. But you're quite right, looking at it now it's pretty easy to do via the ORM.

openprescribing/frontend/tests/price_per_unit/test_savings.py
evansd: This is actually a good illustration of what's confusing about our current caching set up. We've got two sorts of cached values: 1. Things which are held in memory for just the lifetime of the process 2. Things which are written to an external cache backend (previously Redis, soon to be DiskCache).

Caching of type 1 uses `functools.lru_cache` and caching of type 2 uses `django.core.cache`. I was thinking it might be good to have a single unified wrapper which could do either, depending on the arguments used. And then a single place to disable caching for tests. But I haven't quite figured out how that would all work yet.

openprescribing/frontend/price_per_unit/profile.py
evansd: Actually, I've had a similar thing around for a while as I've been working on the MatrixStore and I've ended up using it so often I thought I ought to commit it!

sebbacon: OK - in that case, what about mentioning this somewhere - perhaps the MatrixStore README, as an example of profiling? I can imagine it'll easily get lost otherwise.

openprescribing/frontend/price_per_unit/substitution_sets.py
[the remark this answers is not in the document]
inglesp: I guess this is just a theoretical concern, that we might end up relying on an invariant that's not enforced.  Suggestion withdrawn.

openprescribing/pipeline/management/commands/run_pipeline_e2e_tests.py
evansd: As discussed at the DL day, Rich is the only person who used this and he'd be happy with just a CSV download option

openprescribing/frontend/tests/fixtures/price_per_unit/formulation_swaps.csv
```diff
@@ -0,0 +1,12 @@
+The forma ,Code,Sept_Quantity,Formulation,Alternative formulation,Really equivalent?,,Alternative name,Alternative quantity,Number of words in formulation,Alternative code,DM+Dnonbioequivalent
```
inglesp: Is `The forma ` a mistake?

openprescribing/frontend/tests/price_per_unit/test_substitution_sets.py
```diff
@@ -0,0 +1,58 @@
...
+            self.assertNotEqual(substitution_set_a, substitution_set_b)
+
+    def assertNotSubstitutable(self, presentation_a, presentation_b):
+        self.assertSubstitutable(presentation_a, presentation_b, invert=True)
+
+    def test_substituting_generic_for_brand(self):
+        self.assertSubstitutable("generic_metformin", "branded_metformin")
+
+    def test_excluded_presentations_ignored(self):
+        self.assertNotSubstitutable("generic_inhaler", "gsk_inhaler")
```
inglesp: I'm probably missing something, but can't see where this exclusion is defined.

openprescribing/frontend/tests/fixtures/price_per_unit/formulation_swaps.csv
evansd: Yes, quite possibly; but it's a mistake in the original Google Doc which I've downloaded verbatim so I'd rather not fiddle with it.

openprescribing/frontend/tests/price_per_unit/test_substitution_sets.py
evansd: It's in the big EXCLUSIONS_RE regex in the substitution sets code.

PR #2395, opened 2020-01-20
Add `measures` ImportLog event

### LINKED ISSUE #2228: all-england can be broken during deploy
If PPU savings are imported before measures are recomputed, the all-england page breaks because it tries to find `MeasureGlobal` objects with a month corresponding to the latest `ImportLog` with `category='ppu'`.

### LINKED ISSUE #2391: All England measure savings incorrect following monthly import
After importing Nov 2019, the measure savings at All England level are shown as zero.

evansd:
This is created when all measures have finished importing and all PPU calculations have completed. The All England views now use this date, rather than just the PPU date, which should avoid the problems described in the two linked issues.

Closes #2228 
Closes #2391

### DIFF
--- a/openprescribing/pipeline/management/commands/create_importlog_for_measures.py
+++ b/openprescribing/pipeline/management/commands/create_importlog_for_measures.py
@@ -1,0 +1,10 @@
+from django.core.management import BaseCommand
+from frontend.models import ImportLog
+
+
+class Command(BaseCommand):
+    help = "Records the fact that measures have finished importing"
+
+    def handle(self, *args, **kwargs):
+        date = ImportLog.objects.latest_in_category("prescribing").current_at
+        ImportLog.objects.get_or_create(category="measures", current_at=date)
--- a/openprescribing/pipeline/metadata/tasks.json
+++ b/openprescribing/pipeline/metadata/tasks.json
@@ -277,6 +277,14 @@
             "upload_to_bigquery"
         ]
     },
+    "create_importlog_for_measures": {
+        "type": "post_process",
+        "command": "create_importlog_for_measures",
+        "dependencies": [
+            "import_measures",
+            "generate_ppu"
+        ]
+    },
     "import_generic_mappings": {
         "type": "post_process",
         "command": "import_generic_mappings",
--- a/openprescribing/pipeline/metadata/tasks.json
+++ b/openprescribing/pipeline/metadata/tasks.json
@@ -277,9 +277,9 @@
             "upload_to_bigquery"
         ]
     },
-    "create_importlog_for_measures": {
+    "create_importlog_for_dashboard_data": {
         "type": "post_process",
-        "command": "create_importlog_for_measures",
+        "command": "create_importlog_for_dashboard_data",
         "dependencies": [
             "import_measures",
             "generate_ppu"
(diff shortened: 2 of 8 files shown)


evansd:
Once approved but before merging we should manually create the first `measures` importlog event in the database so we don't get errors immediately after deploying.

openprescribing/pipeline/metadata/tasks.json
```diff
@@ -277,6 +277,14 @@
             "upload_to_bigquery"
         ]
     },
+    "create_importlog_for_measures": {
+        "type": "post_process",
+        "command": "create_importlog_for_measures",
+        "dependencies": [
+            "import_measures",
+            "generate_ppu"
```
sebbacon: Why is `ppu` a dependency for a task that marks measure calculation as complete?  I don't suppose it matters much, but just want to check I understand the logic.

openprescribing/pipeline/management/commands/create_importlog_for_measures.py
```diff
@@ -0,0 +1,10 @@
+from django.core.management import BaseCommand
+from frontend.models import ImportLog
+
+
+class Command(BaseCommand):
+    help = "Records the fact that measures have finished importing"
+
+    def handle(self, *args, **kwargs):
+        date = ImportLog.objects.latest_in_category("prescribing").current_at
+        ImportLog.objects.get_or_create(category="measures", current_at=date)
```
inglesp: Why can't this happen at the end of `import_measures`?

openprescribing/pipeline/metadata/tasks.json
evansd: The short answer is that `measures` is actually not the right name but I couldn't easily think of a better one at the time. We need both the measures and PPU to finish importing before we can produce an updated dashboard. But now it occurs to me that `dashboard_data` might be a reasonable name.

openprescribing/pipeline/management/commands/create_importlog_for_measures.py
evansd: The main reason is the one I discussed above i.e. we need PPU to finish processing as well. But apart from that we'd also need to make sure that it only create the ImportLog when it finished importing all measures, not when we run it on an ad hoc basis to import a single measure. And that would involve more messy logic which I didn't relish.

inglesp: Got it.

evansd:
@sebbacon @inglesp Thanks for the comments. Have just pushed a new commit which is hopefully clearer. Alternative name suggestions also very welcome.

sebbacon:
`dashboard_data` works for me!

evansd:
Have created the relevant ImportLog entry:
```
hello@largeweb2:~$ cd /webapps/openprescribing/openprescribing/
hello@largeweb2:/webapps/openprescribing/openprescribing$ source ../.venv/bin/activate
(.venv35) hello@largeweb2:/webapps/openprescribing/openprescribing$ ./manage.py shell
Python 3.5.3 (default, Sep 27 2018, 17:25:39) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from frontend.models import ImportLog
>>> date = ImportLog.objects.latest_in_category("prescribing").current_at
>>> date
datetime.date(2019, 11, 1)
>>> ImportLog.objects.get_or_create(category="dashboard_data", current_at=date)
(<ImportLog: ImportLog object (15862)>, True)
```

PR #2407, opened 2020-01-25
Fix fabfile for ebmbot

inglesp:


evansd:
> Add shlex appeal

Tempted to reject purely on this basis

PR #2419, opened 2020-01-31
Make date an optional argument to send_ncso_concessions_alerts.py

inglesp:


inglesp:
It's only ever invoked by ebmbot, so this behaviour (while, I agree, is not perfect) is no worse that what we had previously.

PR #2439, opened 2020-02-05
Add link to analyse from dmd results

inglesp:


openprescribing/dmd/templates/dmd/advanced-search.html
```diff
@@ -48,6 +48,7 @@ <h2>Found more than {{ objs|length }} {{ obj_type_human_plural }}</h2>
 <p>Please refine your search.</p>
 {% else %}
 <h2>Found {{ objs|length }} {{ obj_type_human_plural }}</h2>
+<p><a href="{{ analyse_url }}">Analyse prescribing for these drugs</a></p>
```
evansd: Should this be wrapped in a `{% if analyse_url %}` block? I think it's possible for `objs` not to be None, but for `analyse_url` to be empty because it's too long.

inglesp: Well spotted.

PR #2443, opened 2020-02-05
Support PCNs in alerts

inglesp:
There are no tests for PCN alerts, but I have tested manually and got something sensible in the email.

The existing alert-sending tests don't give me much confidence that it would be worthwhile to work out how to extend them to support PCNs.  The hairy part of this is calculating interesting measures, and from what I can tell `InterestingMeasureFinder.measure_filter_for_org()` does the right thing.  But I'd appreciate extra care when reviewing that part of the code.

### DIFF
--- a/openprescribing/frontend/forms.py
+++ b/openprescribing/frontend/forms.py
@@ -84,6 +84,19 @@ class OrgBookmarkForm(forms.Form):
     pcn_id = forms.CharField(widget=forms.HiddenInput(), required=False)
 
 
+class NCSOConcessionBookmarkForm(forms.Form):
+    email = forms.EmailField(
+        label="",
+        error_messages={
+            "required": "This can't be blank!",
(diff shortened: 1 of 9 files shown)


sebbacon:
Looks good.

Re tests, I agree there's no mileage in worrying about testing the interesting measure finder stuff in any new ways.

But what do you think about adding a simple high-level test to minimally exercise the new code path? Along these lines:

https://github.com/ebmdatalab/openprescribing/blob/e604cfc20b2d66af5bf75f1ec2278efb315fa18e/openprescribing/frontend/tests/commands/test_send_monthly_alerts.py#L453-L465

The test could also assert `PCN` is in the email text, per Dave's wondering?

inglesp:
> But what do you think about adding a simple high-level test to minimally exercise the new code path?

I tried this and got bogged down creating the necessary fixtures.

PR #2449, opened 2020-02-06
Ensure ansible playbook uses python 3

sebbacon:


sebbacon:
I use ansible on my computer and had to rebuild an image, and it didn't work.

PR #2452, opened 2020-02-07
Add "Impact and Investment Fund" tag

brianmackenna:
new dashboard to support PCNs with their new contract. will amend measures on seperate pull request

PR #2453, opened 2020-02-07
Iif update measuredef

brianmackenna:
I have assigned the category "iif" (see #2452 ) to support PCNs with new contract. The list of low priority needs to be reviewed once the published list of 20/25 is available. For the moment I have guessed which of the 20 are included. (I think its a good guess seen as though I work on low priority for NHSE but I'm not 100% sure)

### DIFF
--- a/openprescribing/measure_definitions/environmental_inhalers.json
+++ b/openprescribing/measure_definitions/environmental_inhalers.json
@@ -26,6 +26,7 @@
   "tags": [
     "core",
     "greenernhs",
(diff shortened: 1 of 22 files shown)


sebbacon:
Looks fine but you'll also need to add something to https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/common/measure_tags.json, so the tag has a description

brianmackenna:
Is #2452 not it?

PR #2462, opened 2020-02-11
Drop normalised prescribing legacy

inglesp:


inglesp:
That ship has sailed, but thanks!

PR #2483, opened 2020-02-24
Fix PPU tests so they can be run locally

evansd:


evansd:
Sorry, I explained in the commit but should have copied to the PR description:

> Under Python 3.6 `json.loads` accepts bytes, but under 3.5 it only
> accepts strings. We haven't noticed this because our tests run under
> Python 3.6 but it prevents us running tests locally.

PR #2487, opened 2020-02-26
Ensure we send alert signup about the correct bookmark

inglesp:
Fixes https://github.com/ebmdatalab/sysadmin/issues/85.


inglesp:
Failed linting, which was fixed in #2488.

PR #2501, opened 2020-03-05
Update advanced-search.html

brianmackenna:
added dm+d description and blog link for anyone that lands directly on this page.


brianmackenna:
thanks - soz about missing </em>.

I write on to Git browser - is there a better way I should be doing it that would do tests? For measures I uses a json validator.

inglesp:
Not really, I'm happy to act as a human HTML validator!

PR #2505, opened 2020-03-05
Don't show numerator breakdown for raw prescribing measures

### LINKED ISSUE #2135: Hide "presentations contributing to variation" table where not appropriate
On some occasions, such as the proposed "seven-day prescribing" measure, the presentations contributing to variation table is not correct:
image
In this example the quantity and items relate to the *total* number of prescriptions, not those for seven days only.

For these occasions, the most simple thing to do is to have a flag in the measure definition (e.g. `presentation_table = FALSE` so that the table is not shown.  

How feasible is this?

inglesp:
Fixes #2135.

### DIFF
--- a/openprescribing/measure_definitions/semagglutidequantity.json
+++ b/openprescribing/measure_definitions/semagglutidequantity.json
@@ -33,6 +33,7 @@
 		"AND",
 		"bnf_code LIKE '0601023AW%' --Semaglutide"
 	],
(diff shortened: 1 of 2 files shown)


inglesp:
Closing this -- after discussion with Brian we still want to show the breakdown for semagglutidequantity, even though it's potentially misleading.

inglesp:
Will do the seven-day prescribing separately later.

PR #2514, opened 2020-03-10
Redirect branded->generic codes on PPU pages

### LINKED ISSUE #2508: Links from DMD browser to PPU pages are broken in some instances
For example:
https://openprescribing.net/dmd/ampp/4881611000001105/
links to the PPU page using the branded BNF code:
https://openprescribing.net/all-england/1502010J0BNAACE/price-per-unit/
where in fact we want the generic:
https://openprescribing.net/all-england/1502010J0AACECE/price-per-unit/

It may be that we just change the way the DMD browser generates links, or we may decide we should make the PPU pages accept branded codes and work out the appropriate generic itself.

evansd:
Fixes #2508

### DIFF
--- a/openprescribing/frontend/views/views.py
+++ b/openprescribing/frontend/views/views.py
@@ -60,6 +60,9 @@ from frontend.views.spending_utils import (
     NATIONAL_AVERAGE_DISCOUNT_PERCENTAGE,
 )
 from frontend.price_per_unit.savings import get_total_savings_for_org
+from frontend.price_per_unit.substitution_sets import (
+    get_substitution_sets_by_presentation,
+)
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


openprescribing/frontend/views/views.py
```diff
@@ -878,6 +889,19 @@ def all_england_price_per_unit_by_presentation(request, bnf_code):
     return render(request, "price_per_unit.html", context)
 
 
+def _get_generic_bnf_code(bnf_code):
```
inglesp: Does this have to use substitution sets?  I think you can do it just by transforming the BNF code.

My understanding is that given a branded BNF code that ends `YZ`, you could get the generic equivalent by replacing the last six characters with `AAYZYZ`.

evansd: That doesn't account for formulation swaps though. And in general, I'd rather have that logic centralised as far as possible.

inglesp: Ah right.  In which case the name of the function is slightly misleading, as it's doing more than getting a generic BNF code.

evansd: Yeah, we don't really have a good term for this yet. I originally called it "get_canonical_bnf_code" (because I was thinking in terms of the URL) but that isn't quite right so I changed it to "generic". Open to suggestions.

inglesp: `get_primary_substitutable_bnf_code` is a bit of a mouthful, but it's less confusing than `get_generic_bnf_code`.

PR #2518, opened 2020-03-11
Fetch BNF SNOMED mapping automatically

inglesp:
This was previously done manually because we weren't sure whether the HTML page structure would be reliably scrapable.


openprescribing/dmd/management/commands/fetch_bnf_snomed_mapping.py
```diff
@@ -0,0 +1,58 @@
...
+
+        rsp = requests.get(
+            base_url + "/prescription-data/understanding-our-data/bnf-snomed-mapping"
+        )
+        doc = BeautifulSoup(rsp.text, "html.parser")
+
+        href = None
+        for a in doc.find_all("a"):
+            if a["href"].endswith(".zip"):
+                assert href is None
```
sebbacon: Maybe add helpful failure message here - "Found more than one zipfile at <url>, expected only one"?

openprescribing/dmd/management/commands/fetch_bnf_snomed_mapping.py
```diff
@@ -0,0 +1,58 @@
...
+        filename = href.split("/")[-1]
+        datestamp = filename.split(".")[0].split("%20")[-1]
+        release_date = datestamp[:4] + "_" + datestamp[4:6] + "_" + datestamp[6:]
+
+        dir_path = os.path.join(
+            settings.PIPELINE_DATA_BASEDIR, "bnf_snomed_maping", release_date
+        )
+        zip_path = os.path.join(dir_path, filename)
+
+        if os.path.exists(zip_path):
```
evansd: We're checking for the existence of the zip file, but the xlsx is what this script ultimately creates, so if anything goes wrong with the extraction it will get stuck in a partially completed state. Could we change the check to "a file matching `*.xlsx` exists"?

PR #2519, opened 2020-03-11
Replace "/all-england/" URL with "/national/england/"

evansd:
This gives us a consistent `<org-type>/<org-id>` URL pattern which will
make later refactoring easier. It also doesn't hurt that it suggests the
possibility of other national datasets...

We automatically redirect from old to new URLs.


evansd:
Travis failure is due to not being able to run the lint check because it can't download a file from SauceLabs. But I know the lint passes so I'm going to merge anyway.

Issue #2516, opened 2020-03-10
Import republished prescribing data


inglesp:
Because of the One Drug Database changes, the BSA are republishing the last five years of prescribing data, ie data from 2015_01 to 2019_12.

We want to:

* import this republished data as quickly as possible (ideally before the 2020_01 data is published)
* keep the old data around (for reference, and in case we need to roll back)
  * see #2513 for discussion
* verify that the new data is sensible
* work out whether any measure definitions need changing
* work out whether PPU formulation swaps need changing
* minimise the amount of time that the site has incomplete data

In order to do this, we should split up the reimport into several steps.

#### Hack about with code

 - [x] download new prescribing data to `PIPELINE_DATA_BASE_DIR/prescribing_v2`
 - [x] change `convert_hscic_prescribing.py` to read data from `hscic/prescribing_v2`
 - [x] change `convert_hscic_prescribing.py` to append data to `hscic.prescribing_v2`
 - [x] hack `convert_hscic_prescribing.py` so that it stops after appending data to `hscic.prescribing`
 - [x] remove `import_hscic_prescribing.py` and everything that depends on it
 - [x] create `hscic.raw_prescribing_v1` and create `hscic.raw_prescribing_v2`
 - [x] change definition of `hscic.raw_prescribing_normalised` to read from `hscic.raw_prescribing_v2`
 - [x] update definition of `hscic.normalised_prescribing_standard`
 - [x] define `hscic.normalised_prescribing`
 - [x] update measure definitions to use `hscic.normalised_prescribing`
 - [x] create temporary `Measure1` model, with same fields as `Measure`
 - [x] add `from frontend.models import Measure1 as Measure` to `import_measures`
 - [x] change `BQ_MEASURES_DATASET` to `measures_v2`
 - [x] update e2e test setup

#### Back up existing data

 - [x] rename `hscic/prescribing` to `hscic/prescribing_v1`
 - [x] copy `hscic.prescribing` to `hscic.prescribing_v1`
 - [x] copy `hscic.bnf` to `hscic.bnf_backup`

#### Delete temporary data

 - [x] remove `tmp_eu.formatted_prescribing_2019_12`

#### Create new tables/directories

 - [x] create `hscic/prescribing_v2`
 - [x] create `hscic.prescribing_v2`
[...]

PR #2524, opened 2020-03-12
Prepare to import republished prescribing data

inglesp:
This PR implements the changes outlined under "Hack about with code" in #2516.


inglesp:
As usual, GH's showing the commits on the conversation tab in an arbitrary order.  Suggest reviewing commit-by-commit via the commits tab.

openprescribing/frontend/management/commands/measure_sql/normalised_prescribing_standard.sql
```diff
@@ -0,0 +1,26 @@
+SELECT
```
sebbacon: Nit: the views in the current directory are no longer all measure-related, despite what the directory name implies. It would be nice to call it something like `views_sql` rather than `measures_sql`.

inglesp:
> I don't fully understand the consequences of c11e71c. Does this mean that we will no longer show practices that are not in epraccur? Does that only affect the odd new practice in the most recent month of data?

We already don't show them because you get a 500 if you go their page.  But now we're not even pretending to import them.

> is there any danger of measurevalues being created accidentally while Measure1 is in use? Is it worth temporarily disabling the relevant part of the import_measures command?

Very good point, I'll do that.

openprescribing/frontend/management/commands/measure_sql/normalised_prescribing_standard.sql
inglesp: It was already not completely measure-related.  The time to do this will be when we drop `import_measures` -- have got too many balls in the air right now.

openprescribing/matrixstore/build/import_prescribing.py
```diff
@@ -139,7 +139,7 @@ def build_matrices(prescriptions, practices, dates):
     grouped_by_bnf_code = groupby(prescriptions, lambda row: row[0])
     for bnf_code, row_group in grouped_by_bnf_code:
         items_matrix = sparse_matrix(shape, integer=True)
-        quantity_matrix = sparse_matrix(shape, integer=True)
+        quantity_matrix = sparse_matrix(shape, integer=False)
```
sebbacon: So we've talked about keeping this an integer but multiplying by 1000. Now's not the time but we should keep that in mind.

PR #2525, opened 2020-03-13
Drop import_measures --definitions_only from pipeline

inglesp:
There's no need, as definitions get imported as part of import_measures.

This is being fixed now because it should depend on matrixstore_publish
(because it looks up all prescribed presentations from the MatrixStore)
and rather than fixing that, I'm just removing it.


inglesp:
I'm going to merge this myself now, because I'd like to check the e2e tests are passing before merging #2524.  A post-merge sense-checking review would be appreciated though!

PR #2526, opened 2020-03-13
Don't restrict measure BNF codes to those with prescribing

inglesp:
Since cb80cd89, the numerator_/denominator_where attributes have been
constructed based on presentations matching numerator_/
denominator_bnf_codes_filters, which are then restricted to
only those presentations with prescribing.

But doing so causes the end-to-end tests to fail because the end-to-end
tests import only a limited amount of prescribing, so several measures
end up with no matching presentations.  This causes the numerator_/
denominator_where fragments to contain invalid SQL.

This change will have no practical effect on the measure calculations.


inglesp:
Again, self-merging but a post-hoc review would be nice.

PR #2527, opened 2020-03-13
Allow e2e tests to handle measures with no prescribing

inglesp:
The change in a49f885f didn't work, because some measures (eg
aafpercent) are defined in terms of lists of BNF codes that we don't
actually populate in the tests.  So this reinstates the filter removed
in a49f885f, and hacks around the problem.  We expect to revisit this code
soon.


inglesp:
Merging myself again.

PR #2549, opened 2020-03-19
Add command to send Esyma alerts

inglesp:
These will be ready to go out when @sebbacon has reviewed and @brianmackenna is happy with the sample emails I've sent him.


brianmackenna:
HI - very happy with that. I've made a commit with slight tweak on the bullet point.

PR #2550, opened 2020-03-19
Add diskcache to .gitignore

inglesp:
The disckcache directory seems to be created when running tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,4 @@ venv/
 tmp
 logs
 deploy-log.json
+/diskcache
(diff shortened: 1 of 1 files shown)


inglesp:
I've got some extra SQLite files in there:

```
$ tree diskcache/
diskcache/
└── 000
    ├── cache.db
    ├── cache.db-shm
    └── cache.db-wal
```

Would it be better to add `*.db-shm` and `*.db-wal` instead?

evansd:
I would definitely prefer adding `/diskcache`. And possibly removing `*.db` as well, unless that's going to cause other problems.

PR #2553, opened 2020-03-22
Bmk update to injectable abx

brianmackenna:


inglesp:
Let me know when you have the "why it matters".

brianmackenna:
Ah it was in a commit comment that don't seem to easily pull through

unsure where it goes in new measure def structure. to do: we need to exclude 
`bnf_code NOT LIKE "0501070I0%"

why it matters is [here on this google doc](https://docs.google.com/document/d/1DNJoMsNfCv4CIHR3Xo7p9AtQjnwknskPY1_glpqUsXg/edit)

inglesp:
> Ah it was in a commit comment that don't seem to easily pull through

I missed that.  It's hidden behind the `...`:

image

Will update the definition.

inglesp:
I've updated the PR, and it's on staging at http://staging.openprescribing.net/measure/injectable_antibiotics/ccg/00K/.

inglesp:
I've committed some stray files, sorry.  Will sort out later.

PR #2554, opened 2020-03-22
Change measure definitions that depend on presentation names

inglesp:


openprescribing/measure_definitions/tamoxifen.json
```diff
@@ -31,10 +31,9 @@
   "low_is_good": null,
   "numerator_type": "custom",
   "numerator_columns": [
-    "SUM(quantity/ CASE WHEN bnf_name LIKE '%Oral%' THEN 10 ",
-    "WHEN bnf_name LIKE '%Liq%' THEN 10 ",
-    "WHEN RTRIM(bnf_name) LIKE '%10mg' THEN 2 ",
-    "WHEN RTRIM(bnf_name) LIKE '%40mg' THEN 0.5 ",
+    "SUM(quantity/ CASE WHEN bnf_name LIKE '%oral%' THEN 10 ",
```
sebbacon: Does this need `UPPER` too?

openprescribing/measure_definitions/lyrica.json
```diff
@@ -29,14 +29,34 @@
   "low_is_good": true,
   "numerator_type": "bnf_quantity",
   "numerator_bnf_codes_query": [
-    "SELECT bnf_code FROM {hscic}.presentation WHERE name LIKE '%_Cap%'"
```
evansd: We're now including tabs where the previous query included only caps. Just wanted to confirm that this was intentional.

inglesp:
> I'm not quite clear though on what our case sensitivity requirements are

Good point -- it's different on a case-by-case basis.  I've ensured that the new queries are capturing exactly the same presentations as the old ones.

openprescribing/measure_definitions/lyrica.json
inglesp: Yes, thanks.

openprescribing/measure_definitions/tamoxifen.json
inglesp: No, but thanks for checking.

PR #2556, opened 2020-03-24
Update formulation swaps

inglesp:
This PR changes the expected structure of the formulation swaps CSV file (except it doesn't, because the columns we care about have the same name).

The file is generated by downloading a Google Sheet as a CSV.  As part of these changes, we are using a new sheet, which is linked to from the old sheet, and from a comment in the code.

This PR also changes the contents of the CSV file to account for BNF codes changing as part of the One Drug Database changes.

Before this PR (and with the pre-ODD prescribing data) the total savings for All England in 2019_12 was £23,291,161.

With the changes in this PR (and with post-ODD prescribing data) the total savings was £23,205,113.  This is 99.6% of the old value.  The change is most likely to be explained by the quantity-related ODD changes.

(The total savings were computed with `get_total_savings_for_org("2019-12-01", "all_standard_practices", None)`.)


inglesp:
> I tried the link given in the curl command and it seems to be a protected sheet so I can't access it.

Can you try now, with the updated link in this PR?

> If I could access, would the CSV which Google generates have quotes or not? Ideally the local file would just be a mirror of whatever Google gives us.

Yes, afraid so.  Unless you can think of a way to get the CSV not to have quotes, l'll add another commit with the file with quotes for future diffing.

PR #2557, opened 2020-03-24
Update measure definitions to read from hscic.presentation

inglesp:
hscic.bnf has columns for chapter/sections/paragraph/etc, but we do not
yet have an updated copy of the full BNF hieararchy.  These measures
only use presentation codes, which are present in hscic.presentation --
and which we can populate based on the names/codes in the prescribing
data.


inglesp:
This is green in Travis!

PR #2566, opened 2020-03-26
Gracefully handle orgs with no prescribing

evansd:
At present, visiting the homepage of an organisation without any prescribing triggers a KeyError. This PR addresses this, and also refactors some of the testing to make things simpler.

Probably easiest reviewed commit-by-commit, though there are only a handful of them.


openprescribing/matrixstore/tests/decorators.py
```diff
@@ -32,9 +41,17 @@ def setUpClass(inner_cls):
         # Have to wrap this in a staticmethod decorator otherwise Python thinks
         # we're trying to create a new class method
         inner_cls._stop_patching = staticmethod(stop_patching)
+        new_settings = override_settings(
+            CACHES={
+                "default": {"BACKEND": "django.core.cache.backends.dummy.DummyCache"}
+            }
+        )
+        new_settings.enable()
+        inner_cls._new_settings = new_settings
```
inglesp: Given that we also have to do this in price_per_unit/test_savings.py, is it worth extracting this more generally?  Or too much faff?

openprescribing/frontend/tests/test_views.py
```diff
@@ -167,21 +167,16 @@ def test_pcn_bookmark_created(self):
...
-        super().setUpClass()
-        patcher = mock.patch("frontend.views.views.get_total_savings_for_org")
-        mocked = patcher.start()
-        mocked.return_value = 0
-        cls.patcher = patcher
-
-    @classmethod
-    def tearDownClass(cls):
-        cls.patcher.stop()
-        super().tearDownClass()
```
inglesp: Good to see this go!

openprescribing/matrixstore/tests/decorators.py
evansd: I did look into this actually. But it turns out the PPU tests are subtlety different (it's to do with the fact we have two distinct `DataFactory` classes) so it isn't straightforward to do this. A clean up for another day, I think.

evansd:
@inglesp Thanks for the review. As you'll see there was, in the usual manner of things, A Bunch More Faff. Would you mind casting your eye over it again when you get a mo?

inglesp:
All looks sensible!

PR #2576, opened 2020-04-04
Esmya ulipristal 

brianmackenna:
New measure supporting previously issued alert. No supporting notebook due to COVID workload and urgency of this measure. Will revisit at review date.


inglesp:
On staging at http://staging.openprescribing.net/measure/ulipristal/.

PR #2581, opened 2020-04-08
Always use dashboard_data ImportLog when choosing what to display

inglesp:
A dashboard_data ImportLog is created at the very end of the monthly
import, while the prescribing ImportLog is (now) created once
convert_hscic_prescribing has been run.


inglesp:
Closing in favour of #2583.

PR #2586, opened 2020-04-09
Fix issue with mixed ndarray/sparse-matrix types

evansd:
We just happened not to have any mixed types with the old data so we
didn't notice this problem before.


evansd:
Oh hey coveralls. Nice to see you with your totally incorrect result. If you bump into Travis tell him to get in touch sometime.

PR #2612, opened 2020-04-27
Ensure data is in BQ before downloading it

inglesp:
The last two monthly imports have failed because the data that the
matrixstore_build expects to find in BigQuery is not actually there.

### DIFF
--- a/openprescribing/matrixstore/build/download_practice_stats.py
+++ b/openprescribing/matrixstore/build/download_practice_stats.py
@@ -28,9 +28,11 @@ def ensure_stats_downloaded_for_date(date):
     filename = get_practice_stats_filename(date)
     if os.path.exists(filename):
         return
+    client = Client("hscic")
+    check_stats_in_bigquery(date, client)
     logger.info("Downloading practice statistics for %s", date)
     temp_name = get_temp_filename(filename)
-    result = Client("hscic").query(
+    result = client.query(
         """
         SELECT *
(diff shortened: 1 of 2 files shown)


openprescribing/matrixstore/build/download_practice_stats.py
```diff
@@ -28,19 +28,37 @@ def ensure_stats_downloaded_for_date(date):
...
         WHERE month = TIMESTAMP("%s")
         """
         % (date,)
     )
+
     with gzip.open(temp_name, "wt") as f:
         writer = csv.writer(f)
         writer.writerow(result.field_names)
         for row in result.rows:
             writer.writerow(row)
```
evansd, line 48: Couldn't we just do:
```py
row = None
for row in result.rows:
    writer.writerow(row)
if row is None:
   raise ValueError("No practice statistics found")
```
That feels less error prone, and less work, as we're not duplicating the query.

openprescribing/matrixstore/build/download_prescribing.py
```diff
@@ -136,6 +137,21 @@ def extract_data_for_date(date, bq_client):
     )
 
 
+def check_prescribing_data_in_bigquery(date, bq_client):
```
evansd, line 140: I can see that we can't avoid doing a separate query here.

inglesp:
Test failure is unrelated -- something flaky in ansible.

PR #2631, opened 2020-05-12
Handle presentations without names in PPU code

evansd:


evansd:
Exactly my feelings on writing this!

PR #2632, opened 2020-05-12
Revert "Ensure data is in BQ before downloading it"

inglesp:
Reverts ebmdatalab/openprescribing#2612

This causes the e2e tests to fail.

### DIFF
--- a/openprescribing/matrixstore/build/download_practice_stats.py
+++ b/openprescribing/matrixstore/build/download_practice_stats.py
@@ -28,11 +28,9 @@ def ensure_stats_downloaded_for_date(date):
     filename = get_practice_stats_filename(date)
     if os.path.exists(filename):
         return
-    client = Client("hscic")
-    check_stats_in_bigquery(date, client)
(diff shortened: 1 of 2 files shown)


evansd:
Isn't it a problem of some kind if there isn't any practice stats data in BQ when the end-to-end tests run?

inglesp:
> Isn't it a problem of some kind if there isn't any practice stats data in BQ when the end-to-end tests run?

We don't have five years of data for the end-to-end tests to use, so this assert gets hit.

evansd:
Ah of course, this is a month-by-month check rather than a check over the entire five years.

I still feel like this is a useful check as it would have saved us on a couple of occasions recently. I'd be more in favour of adding a setting like `ALLOW_EMPTY_PRACTICE_STATISTICS` which is False by default but True in the end-to-end tests.

I can try to find a bit of time to do this if you're deep in the middle of [_the other thing_] (it's hard to remember which of our communication channels are public sometimes!).

inglesp:
Good idea.  Closing in favour of #2653.

PR #2708, opened 2020-06-22
Update research page

inglesp:


brianmackenna:
The changes I made all look fine but I notice - seem to be changed to &#8211which appears to be the Unicode Decimal Code. Is this on purpose? Does it matter?

inglesp:
Not on purpose, and doesn't matter.  Probably just an artifact of how I copied/pasted the HTML from the blog.

PR #2719, opened 2020-07-03
Remove unused presentation metadata from MatrixStore

evansd:
I think this ended up sticking around from the very first
proof-of-concept MatrixStore code, despite the fact that we have never
-- and now will never -- use it.

Removing it now while it occurs to me.


evansd:
I have no idea why the ansible build test is suddenly failing because it can no longer compile PyYAML.

evansd:
The PyYAML problem above seems to have magically fixed itself over the weekend...

So this PR is in fact green, although Travis has failed to update the status:
https://travis-ci.org/github/ebmdatalab/openprescribing/builds/704730846

PR #2733, opened 2020-07-16
Fix broken Browserstack link in README

StevenMaude:
Adding the https prefix is required to fix this link.

Otherwise, GitHub treats it as a relative link to a file inside the
repository and, in this case, this file doesn't exist.


inglesp:
Thanks @StevenMaude!  Tests are failing because as an outside collaborator, Travis won't let the tests use secret environment variables.

PR #2754, opened 2020-07-27
Update to PyArrow 1.0.0

evansd:
This release now promises stability of the binary serialization format:
https://arrow.apache.org/docs/format/Versioning.html

Unfortunately it breaks compatibility with the older version of PyArrow
we were using so that we can no longer read existing MatrixStore files
after we update -- hence the need to update the test fixture. This means
there will be some coordination required around deployment, with a build
of the latest data created in the new format ready to swap into place.

Note that while this version does have native scipy.sparse matrix
support which would in theory mean we could drop some of the custom code
in our serializer, in practice the performance isn't good enough for our
purposes (the profile test case jumps from 2.5s to 4.3s). The PyArrow
code makes the same mistake we initially did which is to allow the
horribly expensive scipy.sparse initialisation code to run
unnecessarily.

Closes #2751


evansd:
The latest prescribing file has been rebuilt using the new format:
`matrixstore_2020-05_2020-07-27--16-25_6bad2978c4a7b439.sqlite`

PR #2780, opened 2020-08-18
Add workaround for vagrant/ansible on MacOS 10.15

madwort:
By default on MacOS 10.15, vagrant's tooling does not have sufficient permissions to install correctly.


madwort:
Nb. could put this in a footnote if you prefer?

madwort:
Needs more fixes than just this - closing for now

PR #2783, opened 2020-08-18
Setup Github Actions

madwort:
* Run docker-compose from github workflow
* Add google cloud credentials
* Document how to update the encryption key for github actions
* Generate SAUCE_HOST
* Use SAUCE_HOST for "nonfunctional" test suite requires it to be set, although it doesn't actually use it (AFAICS)
* Disable functional tests, to emulate current Travis behaviour


sebbacon (commented):
Looks great. Is there an argument for just pasting the entire google credentials file as a secret into the GH settings? Rather than going via GPG?

madwort:
Yeah, I did consider pasting the whole file in as a secret:

Pros:
* fewer moving parts
* Github automatically redacts secrets in log output (but obviously, not things decrypted with secrets)

Cons:
* Github may not be able to correctly redact a JSON secret - ["To help ensure that GitHub redacts your secret in logs, avoid using structured data as the values of secrets. For example, avoid creating secrets that contain JSON or encoded Git blobs."](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#about-encrypted-secrets)

The setup used here is pretty much straight out of [their docs](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#limits-for-secrets).

sebbacon (approved):

inglesp (approved):
All looks readable and sensible.

The only thing missing from the Travis config (as far as I can tell) is the sending of notifications.  Is that something we can do with Actions?

madwort:
Cool, yeah, email notifications were turned on for me automatically as soon as I started working on actions for this repo - are you getting them too? If not, I wonder if there's some per-user config in github that you'll need to do? 

Slack notifications are also possible but not setup yet - I can do in a follow-up PR?

PR #2788, opened 2020-08-25
Create environment file

madwort:
* `wait_for_postgres.sh` wants to read it, and it's currently not created until later in the process
* also note for future reference that `wait_for_postgres.sh` seems to be unnecessary (for at least some environments), but I don't think it's a problem to keep it


inglesp (approved):
:+1:

PR #2799, opened 2020-08-26
Create Github Actions Slack notification

madwort:
* use https://github.com/voxmedia/github-action-slack-notify-build as it seems to be one of the few plugin options that will easily link to the PR page (which I think is very useful!)
* authenticate with `SLACK_BOT_TOKEN` in Github Secrets
* notify failures separately - can be useful to be aware of (e.g.) linting fail vs unit test fail
* although it is possible to get the plugin to update a message as the CI progresses, this proved to be fiddly to do with multiple jobs, so I haven't set that up

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,6 @@
-name: CI
+name: "OpenPrescribing CI"
 
-on:
-  push:
+on: [ pull_request ]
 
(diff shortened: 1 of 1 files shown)


inglesp (approved):
```
color: danger
```

My favourite colour.

PR #2806, opened 2020-09-01
Pin setuptools to a working version

madwort:
* v50 was just released and doesn't work for us
* pin to last known working (49.6.0)
* seems like this is a common problem https://github.com/pypa/setuptools/issues/2350

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,8 @@ Install library dependencies (current as of Debian Jessie):
 
 Ensure pip and setuptools are up to date:
 
-    pip install -U pip setuptools
+    pip install -U pip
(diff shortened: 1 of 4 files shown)


inglesp (approved):
Fine.  When will we be able to unpin, and how will we remember to do this?

madwort:
It _looks like_ we'll be able to unpin when we upgrade from python 3.5, but will obviously need to test to verify that. 

Also - it seems that it's not installed where it looks like it's installed...!

madwort:
Seems like this might actually be https://github.com/pypa/setuptools/issues/2350

Issue #2379, opened 2020-01-15
Errors thrown by functional tests


evansd:
We had the first flaky failure on the functional tests that we've seen in a long while. This isn't itself too troubling as these are much less frequent than they used to be and a certain amount of flakiness is inevitable with the convoluted remote browser setup. However I noticed when looking at it that the functional tests are throwing a lot of errors like this:
```
----------------------------------------
Exception happened during processing of request from ('172.18.0.1', 37630)
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/socketserver.py", line 317, in _handle_request_noblock
    self.process_request(request, client_address)
  File "/usr/local/lib/python3.6/socketserver.py", line 348, in process_request
    self.finish_request(request, client_address)
  File "/usr/local/lib/python3.6/socketserver.py", line 361, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/local/lib/python3.6/socketserver.py", line 721, in __init__
    self.handle()
  File "/usr/local/lib/python3.6/http/server.py", line 418, in handle
    self.handle_one_request()
  File "/usr/local/lib/python3.6/http/server.py", line 406, in handle_one_request
    method()
  File "/code/openprescribing/frontend/tests/functional/mock_api_server.py", line 104, in do_GET
    self.wfile.write(data.encode("utf8"))
AttributeError: 'dict' object has no attribute 'encode'
----------------------------------------
```
(From https://travis-ci.org/ebmdatalab/openprescribing/jobs/633828763 although this log won't stick around forever)

This is note to myself (mainly) to look into this at some point.

evansd:
Fixed in #2808

PR #2808, opened 2020-09-01
Use BrowserStack to run functional tests

madwort:
* Set up functional tests to run on BrowserStack
* Most tests pass, a couple of tests are fixed in this PR
* Still a couple of failing tests as below:

This test fails consistently, and looks like it could be fixed:

```
test_summary_totals_on_analyse_page (frontend.tests.functional.test_charts.AnalyseSummaryTotalsTest)
```

```
FAIL: test_summary_totals_on_analyse_page (frontend.tests.functional.test_charts.AnalyseSummaryTotalsTest)
  File "/code/openprescribing/frontend/tests/functional/test_charts.py", line 106, in test_summary_totals_on_analyse_page
    self.assertEqual(element.text.strip(), value)
AssertionError: "Aug '16" != "Sep '16"
- Aug '16
+ Sep '16
```

These two tests fail intermittently with a timeout: 

```
test_map_slider (frontend.tests.functional.test_charts.MapTest) 
test_nothing_hidden_by_default (frontend.tests.functional.test_charts.SmallListTest) 
```


openprescribing/frontend/tests/functional/mock_api_server.py
```diff
@@ -98,6 +98,7 @@ def do_GET(self):
         else:
             logger.error("Unhandled path %s" % self.path)
             self.send_response(requests.codes.not_found)
+            data = str(data)
```
madwort, line 101: Some functional tests request `/api/1.0/spending/` from this mock api server, which isn't implemented, so it falls back to attempting to return the default `{}`, which being a `dict` doesn't support `encode('utf-8')`. If we convert to string here, our test doesn't crash.

madwort, line 101: Some functional tests request `/api/1.0/spending/` from this mock api server, which isn't implemented, so it falls back to attempting to return the default `{}`, which being a dict doesn't implement `.encode('utf-8')`. If we convert to string here, our test doesn't crash.

evansd: Ah you've fixed #2379!

madwort: aha!

evansd:
All those failed tests pass for me when I run these locally, but a different test which passes in CI fails. Fun times! I shall have to do further investigation.

evansd:
Looks like this is nearly there but not quite. The test suite now passes (with 3 expected failures) but it's then running another command afterwards which fails with:
`Cannot create container for service test-production: invalid IP address in add-host: ""`

Context:
```
The TRAVIS variable is not set. Defaulting to a blank string.
The TRAVIS_JOB_NUMBER variable is not set. Defaulting to a blank string.
The TRAVIS_BUILD_NUMBER variable is not set. Defaulting to a blank string.
The TRAVIS_BRANCH variable is not set. Defaulting to a blank string.
The TRAVIS_PULL_REQUEST variable is not set. Defaulting to a blank string.
The TRAVIS_JOB_ID variable is not set. Defaulting to a blank string.
The COVERALLS_REPO_TOKEN variable is not set. Defaulting to a blank string.
The SAUCE_USERNAME variable is not set. Defaulting to a blank string.
The SAUCE_ACCESS_KEY variable is not set. Defaulting to a blank string.
The USE_SAUCELABS variable is not set. Defaulting to a blank string.
The SAUCE_HOST variable is not set. Defaulting to a blank string.
Starting openprescribing_db-test_1 ... 

Starting openprescribing_db-test_1 ... done
Pulling test-production (ebmdatalab/openprescribing-py3-base:latest)...
Cannot create container for service test-production: invalid IP address in add-host: ""
latest: Pulling from ebmdatalab/openprescribing-py3-base
Digest: sha256:80dd8366985d7c858074455a2c66e23319bf4c2721a96f3690f43c3d5106b312
Status: Downloaded newer image for ebmdatalab/openprescribing-py3-base:latest
##[error]Process completed with exit code 1.
```

@madwort I'm going to punt this back to you if that's OK as it may be something easy to fix if you've been looking at this recently. If not them, well, we tried our best...

madwort:
ah yeah, leave that to me!

.github/workflows/main.yml
```diff
@@ -30,26 +30,38 @@ jobs:
...
+          local-identifier: random
+          local-logging-level: all-logs
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Retrieve Google Cloud credentials
+        run: ./.github/scripts/decrypt_secret.sh
+        env:
+          GOOGLE_CLOUD_GITHUB_ACTIONS_PASSPHRASE: ${{ secrets.GOOGLE_CLOUD_GITHUB_ACTIONS_PASSPHRASE }}
+      - name: Functional tests
+        run: docker-compose run --service-ports test && docker-compose run --service-ports test-production
```
madwort, line 57: replicating what happened in Travis - running two test suites with an `&&` - `test` now pass but will separate out `test-production` (& possibly disregard)

evansd:
There must be a German word for the sense of satisfaction you get from a green tick, mixed with the shame of knowing you got it by marking a bunch of tests xfail.

evansd (approved):

Issue #2813, opened 2020-09-03
Feedback form: error 500


brianmackenna:
Sending feedback via the form prompts a server 500 error. It isn't critical as thre is an email contact but very annoying for users I imagine.

![image](https://user-images.githubusercontent.com/40460354/92083912-9655dc80-edbe-11ea-933d-27fbbae2d8c8.png)

inglesp:
This is caused by our Mailgun account being suspended.  (This is the same problem that caused sending the NCSO alerts to fail last night.)

evansd:
Account is now unsuspended.

PR #2814, opened 2020-09-03
Turn off email

inglesp:
We can revert this PR once #2813 is resolved.


evansd (approved):
Seems sensible. I would probably have just stuck `style="display: none"` on the form element but this works as well.

PR #2823, opened 2020-09-08
Silence pending deprecation warning from `cryptography`

evansd:
Its presence is causing all our cronjobs to become chatty.

### DIFF
--- a/openprescribing/openprescribing/settings/base.py
+++ b/openprescribing/openprescribing/settings/base.py
@@ -3,8 +3,18 @@ from os.path import abspath, basename, dirname, join, normpath
 from sys import path
 from django.core.exceptions import ImproperlyConfigured
 from common import utils
+import warnings
 
 import sys
+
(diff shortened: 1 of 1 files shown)


openprescribing/openprescribing/settings/base.py
```diff
@@ -3,9 +3,19 @@
...
 
 import sys
 
+# This `cryptography` warning is making our cronjobs chatty
+warnings.filterwarnings(
+    "ignore", "Python 3.5 support will be dropped in the next release"
+)
+if sys.version_info >= (3, 6):
+    warnings.warn(
+        "Warning supression on line above is now redundant and should be removed"
```
madwort, line 16: nice!

evansd: It's a bit meta, isn't it? A warning about a thing which turns off a warning. I'm just keen to work against the inevitable accumulation of cruft that all projects develop.

inglesp (approved):

PR #2826, opened 2020-09-10
Update black to 20.8b1

madwort:


evansd (approved):
Looks good. I think the only reason our lint tools are installed separately (and specified in two places, as you discovered!) is that we used to run on py2.7 and black was py3 only. So there's probably a simplification we can make at some point, but let's not worry about that now.

Issue #2828, opened 2020-09-11
Remove all Travis-related stuff


evansd:
Since #2808 we no longer use Travis CI for anything in this project. I've disabled the Travis webhook which ought to stop it running, but for hygiene's sake it would be good to grep the codebase and remove anything Travis related that we no longer need.

PR #2830, opened 2020-09-14
Remove Travis config file

evansd:
Hopefully this will prevent Travis attempting to run on this repo.
There doesn't seem to be any per-repository setting to disable it, and I
don't want to deactivate it for our whole account as there are other
repos that use it.

This partially addresses #2828 but doesn't close it as I think there are
still other bits of Travis-related code that could do with cleaning up.


evansd:
Argh, it doesn't work. Travis is still there as an expected check. How on earth do we get rid of it?

evansd:
Ah! I think it was appearing because it was marked as required in the "branch protection settings" in Github. I've disabled that and now it doesn't seem to be showing.

madwort (commented):
Yep, good to remove this as a start.

madwort (approved):

PR #2837, opened 2020-09-16
Run Github Actions CI on push to master

madwort:
* Pushes to pull requests are already automatically built
* If we enable builds for all pushes, opening a PR triggers two builds


evansd (approved):
:+1:

PR #2838, opened 2020-09-16
add in missing codes for some products

richiecroker:


inglesp (approved):
Fine from my non-clinical perspective!  Nudge me on Slack when you'd like this re-imported.

brianmackenna (approved):
I have added two products that @AnnaB45 has identified 

(they only account for 3 items per annum at the minute.) There are more products that might be possible additions but this can wait for resource #2843

PR #2859, opened 2020-09-23
Tweak dependabot config

madwort:
* ignore more packages that do not support python 3.5. These were previously ignored with github PR comments, but dependabot no longer respects these.
* add a quick notes on policy to the README file


inglesp (approved):
Brill, thanks.

PR #2864, opened 2020-09-24
Support downloading practice level PPU data for CCG

evansd:
Previously @richiecroker used BigQuery to get this as it ended up being
stored there as a side-effect of the calculation method.

This PR patches on another argument to what is already a pig's ear of an
API. (Note: this uses "org" rather than "entity" as that's what we've
settled on elsewhere.) It also adds a dropdown option to the download
button (where applicable) to allow downloading this data via the UI.

localhost_8000_ccg_02W_price_per_unit_


inglesp (approved):
:ship: :it:

PR #2871, opened 2020-09-29
Fix DT fetcher properly

inglesp:


evansd (approved):
Lol! At least I get equal blame for not spotting it in the last review

PR #2874, opened 2020-09-30
Pin google-cloud-bigquery-* to < 2.0.0

madwort:
* pin these packages because 2.0.0 drops support for python 3.5, makes breaking changes to the api, and causes our unit tests to fail
* see also https://github.com/ebmdatalab/openprescribing/pull/2873 and https://github.com/ebmdatalab/openprescribing/pull/2875

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,6 +21,9 @@ updates:
     versions:
     - "< 3"
     - ">= 2.a"
+  - dependency-name: google-cloud-bigquery
+    versions:
+    - "< 2.0.0"
   - dependency-name: google-cloud-bigquery-storage
     versions:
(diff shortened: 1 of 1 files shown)


inglesp:
Why do we need to do this?

madwort:
this is the new `@dependabot ignore`

inglesp:
Oh right.  And why are we pinning this version?  Perhaps we could explain in a comment?

madwort:
yep sure - revised the PR description!

inglesp (approved):
I meant a comment in the yaml file itself, but we can probably assume that by default a pinning is because of Python 3.5, and add comments to the yaml file when this is not the case.

madwort:
🤝

PR #2883, opened 2020-10-06
Update handlebars to 4.0.14

madwort:
* in general, we're not confident in our js tests, but bumping to the latest 4.0.x is probably safe


inglesp:
Closing in favour of #2880.

PR #2898, opened 2020-10-21
Amend the dependabot ignore configuration

madwort:
* we should be specifying versions to ignore, not version to accept


.github/dependabot.yml
```diff
@@ -23,10 +23,10 @@ updates:
     - ">= 2.a"
```
madwort, line 23: I think I was confused by these auto-generated requirements, which I now don't think make sense...

inglesp (approved):

PR #2902, opened 2020-11-05
Add script to create historial practice table

inglesp:
This is probably a one-off, but was a bit fiddly and might be useful
again.

See https://ebmdatalab.slack.com/archives/C33TWNQ1J/p1604424022241500


evansd (approved):
TIL: `runscript`

inglesp:
> TIL: `runscript`

It's part of [django-extensions](https://django-extensions.readthedocs.io/en/latest/).

PR #2905, opened 2020-11-09
Fix fetch_and_import_drug_tariff properly

inglesp:
Turns out `year` needs to be defined.

Hadn't tested this, sorry.  Annoying that `flake8` hadn't picked this up.


evansd (approved):
Ah yeah, I've been caught by this before. It's not quite clever enough to spot if a var is only defined in one branch of a conditional.

PR #2906, opened 2020-11-17
Reinstate import export

inglesp:
This is a manual revert of 84a9b94a008eee95d325fdaee51c5b4cd9403d29

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,7 +59,7 @@ services:
       - USE_SAUCELABS=${USE_SAUCELABS}
       - GOOGLE_APPLICATION_CREDENTIALS=/code/google-credentials.json
     extra_hosts:
-      - "saucehost:${SAUCE_HOST}"
+      - "saucehost:${SAUCE_HOST-127.0.0.1}"
     ports:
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -59,7 +59,7 @@ services:
       - USE_SAUCELABS=${USE_SAUCELABS}
       - GOOGLE_APPLICATION_CREDENTIALS=/code/google-credentials.json
     extra_hosts:
-      - "saucehost:${SAUCE_HOST}"
+      - "saucehost:${SAUCE_HOST-127.0.0.1}"
```
madwort, line 62: tbh, I'm not quite sure what this does, it looks like a default value, but I think it's probably just string concat. However, we've stopped using Sauce Labs, so we just need docker-compose to not fail here because it's missing this value. Ideally we'd do a deeper dive & remove this whole bit of cruft, but that doesn't seem worth doing as we're not sure of the future of the project right now.

madwort: (and we do this same thing elsewhere in docker-compose)

evansd (approved):

Issue #2517, opened 2020-03-11
Alerts still not getting sent out without hassle


sebbacon:
Running the command with 

`./manage.py send_monthly_alerts --max_errors=50`

```
        /webapps/openprescribing/openprescribing/frontend/views/bookmark_utils.py:205: RuntimeWarning: Mean of empty slice
  self.target_means.append(np.nanmean(window))
/webapps/openprescribing/.venv35/lib/python3.5/site-packages/numpy/lib/nanfunctions.py:1667: RuntimeWarning: Degrees of freedom <= 0 for slice.
  keepdims=keepdims)
console: [opttick] - GTM-5PX77GZ: exception: TypeError: undefined is not a function (evaluating 'r.getEntries()')
console: [opttick] - GTM-5PX77GZ: exception: TypeError: undefined is not a function (evaluating 'r.getEntries()')
Error posting to https://api.mailgun.net/v3/openprescribing.net/messages:
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='api.mailgun.net', port=443): Read timed out. (read timeout=30)
Sending a message to xxx@nhs.net from OpenPrescribing <feedback@openprescribing.net>
Traceback (most recent call last):
  File "/webapps/openprescribing/.venv35/lib/python3.5/site-packages/urllib3/contrib/pyopenssl.py", line 485, in wrap_socket
    cnx.do_handshake()
  File "/webapps/openprescribing/.venv35/lib/python3.5/site-packages/OpenSSL/SSL.py", line 1934, in do_handshake
    self._raise_ssl_error(self._ssl, result)
  File "/webapps/openprescribing/.venv35/lib/python3.5/site-packages/OpenSSL/SSL.py", line 1646, in _raise_ssl_error
    raise WantReadError()
OpenSSL.SSL.WantReadError

During handling of the above exception, another exception occurred: 

Traceback (most recent call last):
  File "/webapps/openprescribing/.venv35/lib/python3.5/site-packages/urllib3/connectionpool.py", line 376, in _make_request
    self._validate_conn(conn)
  File "/webapps/openprescribing/.venv35/lib/python3.5/site-packages/urllib3/connectionpool.py", line 994, in _validate_conn
    conn.connect()
  File "/webapps/openprescribing/.venv35/lib/python3.5/site-packages/urllib3/connection.py", line 360, in connect
    ssl_context=context,
[...]

PR #2911, opened 2020-11-20
Change wording on concessions

brianmackenna:
I have amended wording on email alerts (rationale below). I have changed 
- NCSO template
- All England alert template (although see #2517 as they aren't going atm)

Is there anywhere else it needs changing? I don't think so but not sure with templates

------

We have received feedback that are wording on price concessions is not perfect. We say

> We estimate that in November 2020 total price concessions for out-of-stock medicines will cost

I think "out of stock"  was my/our attempt at describing types of items in plain english but it is not technically true as concessions are not applied to things that can't be gotten! More accurately they are described by wording suggested by a user

> We estimate that in November 2020 total price concessions for products reported to only be available above Drug Tariff price


richiecroker (approved):
happy with this change

inglesp (approved):
Should cover it!

PR #2918, opened 2020-11-25
actually fix dmd link

brianmackenna:
sorry!


inglesp (approved):

madwort:
Nb. I've updated the nginx config so that the old link (without a trailing slash) will also work!

PR #2926, opened 2020-12-07
Fixes for Python 3.8

evansd:
Contains fixes for running under Python 3.8 and on newer versions of the base OS.

These are backwards compatible and so can be merged while we're still running under Python 3.5.


inglesp (approved):
Thanks for doing all the legwork!

Issue #2927, opened 2020-12-08
Review ignored dependency upgrades now we're on Python 3.8


evansd:
As per https://github.com/ebmdatalab/openprescribing/issues/2103#issuecomment-558556893 we've had to ignore various dependency upgrades that weren't compatible with Python 3.5. Now we're running 3.8 we should review these.

#2245

Other unmerged dependency PRs:
https://github.com/ebmdatalab/openprescribing/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+is%3Aunmerged+label%3Adependencies+comments%3A%3E1+updated%3A%3E%3D2019-11-27

inglesp:
Have you seen the configuration in https://github.com/ebmdatalab/openprescribing/blob/master/.github/dependabot.yml?

evansd:
Ah, I hadn't. I guess we just want to delete the entire "ignore" section?

inglesp:
Worth a go.

madwort:
I think this is done

PR #2928, opened 2020-12-08
Remove dependabot ignore rules

evansd:
We think most of these are only there because of Python version
incompatibilities which should no longer apply now we're running 3.8.

In future we should include comments with each ignore rule so we know
why each is there. (I'm assuming the intial version of this file was
auto-generated so not finger-pointing here!)

Possibly addresses #2927

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,47 +10,8 @@ updates:
   - dependency-type: direct
   - dependency-type: indirect
   ignore:
-  - dependency-name: bcrypt
-    versions:
-    - "> 3.1.7"
-  - dependency-name: django
-    versions:
-    - ">= 3.a"
-    - "< 4"
-  - dependency-name: fabric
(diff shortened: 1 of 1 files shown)


inglesp (approved):
LGTM, but worth bringing in @madwort as he's touched this file since it was autogenerated.

madwort:
This is a mix of the pre-existing items, which I don't know much about, and new ignores which are based only on test failures, so if tests pass I think we should just remove these ignores. The rationale for some of them is documented in the PR - e.g. https://github.com/ebmdatalab/openprescribing/pull/2874 "google-cloud-bigquery-2.0.0 drops support for python 3.5" - but that requires a bit of digging to get to, an inline comment would probably have been easier...?

evansd:
That makes sense. I guess for now it's OK to remove them all and then add back as and when dependabot creates failing PRs.

PR #2935, opened 2021-01-09
Changed deprecated Docker instruction

cjwright83:
MAINTAINER is deprecated and should be replaced with LABEL.
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.8
 
-MAINTAINER Seb Bacon version: 0.2
+LABEL maintainer="Seb Bacon version: 0.2"
 ENV PYTHONUNBUFFERED 1
 RUN mkdir /code
 WORKDIR /code


madwort (approved):

madwort:
Thanks @cjwright83 , just check how to trigger a CI run that has access to our secrets...

cjwright83:
Thanks @madwort. Should this CI problem be raised as an issue?

evansd:
Hi @cjwright83, thanks for the contribution!

The CI problem is a general issue that affects all CI runs in Github which use secrets (see e.g. [here](https://github.community/t/feature-request-allow-secrets-in-approved-external-pull-requests/18071) and [here](https://github.community/t/allow-secrets-to-be-shared-with-forks-from-trusted-actions/16525/10)). I think the workaround @madwort used is probably the best we can do for now.

PR #2936, opened 2021-01-11
Changed deprecated Docker instruction

madwort:
* duplicate of https://github.com/ebmdatalab/openprescribing/pull/2935 but pushed by @madwort


evansd:
I guess the solution is to merge the original now it's green and close this dupe?

Issue #2939, opened 2021-02-02
Semaglutide injection measure definition including tablets


brianmackenna:
@andrewbarron85 has drawn my attention to the fact that the [semaglutide measure](https://openprescribing.net/measure/semagglutidequantity/definition/) is now including tablets ([NLP STP example](https://openprescribing.net/measure/semagglutidequantity/stp/E54000028/))

A quick review looks like we did not foresee release of tablet preps when we wrote the measure. The very quick fix is to include `bnf name NOT LIKE tab`. Other fixes include using `dmd.formroute` or curation of bnf codes.

brianmackenna:
We thought #2940  fixed this issue but tablets are still being reported on the [measure](https://openprescribing.net/measure/semagglutidequantity/national/england/#semagglutidequantity).

inglesp:
You're missing a `\n` from [this line](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/measure_definitions/semagglutidequantity.json#L46).

Because it's missing, the [generated SQL](https://openprescribing.net/measure/semagglutidequantity/definition/) is:

```
SELECT
     CAST(month AS DATE) AS month,
     practice AS practice_id,
     SUM(total_quantity) AS numerator
 FROM hscic.raw_prescribing_normalised
 WHERE quantity_per_item>3 AND bnf_code LIKE '0601023AW%' --Semaglutide AND bnf_name  LIKE '%inj%'
 GROUP BY month, practice_id
```

and not:

```
SELECT
     CAST(month AS DATE) AS month,
     practice AS practice_id,
     SUM(total_quantity) AS numerator
 FROM hscic.raw_prescribing_normalised
 WHERE quantity_per_item>3 AND bnf_code LIKE '0601023AW%' --Semaglutide
       AND bnf_name  LIKE '%inj%'
 GROUP BY month, practice_id
```

That is, everything after the comment is erroneously included in the comment.

See [lpneedles.json](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/measure_definitions/lpneedles.json#L36) for where this is done correctly.

I'm going to check all other measures to see whether we hit this elsewhere.

If we get more funding, reworking measure definitions is top of my wishlist.

richiecroker:
Thanks @inglesp: [hopefully fixed and PR created](https://github.com/ebmdatalab/openprescribing/pull/3066).

inglesp:
This is the only measure affected.

```
[...]

PR #2940, opened 2021-02-02
Update semagglutidequantity.json

brianmackenna:
This is a slightly temporary fix for #2939 . I have included all the possible variations of tabs/caps, this might be slightly overkill and there are nicer ways of fixing but this is most time efficient I think!

### DIFF
--- a/openprescribing/measure_definitions/semagglutidequantity.json
+++ b/openprescribing/measure_definitions/semagglutidequantity.json
@@ -54,7 +54,7 @@
 	],
 	"denominator_from": "{hscic}.raw_prescribing_normalised",
 	"denominator_where": [
-		"bnf_code LIKE '0601023AW%' --Semaglutide"
+		"bnf_code LIKE '0601023AW%' --Semaglutide",
 		"AND",
 		 "(",
 	    "bnf_name  NOT LIKE '%tablet%' OR",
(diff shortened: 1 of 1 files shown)


richiecroker:
this looks fine, although we might come a cropper again if liquid specials start appearing.

One alternative may be to have

```
bnf_code LIKE '0601023AW%' --Semaglutide
AND
bnf_name  LIKE '% inj %'
```

madwort:
nb @brianmackenna I think it might not be your change that's the problem with CI, will look into it when I can

madwort:
rebased to add the ansible CI fix

richiecroker:
@brianmackenna just thinking about this again - the code won't work, as you'll need `bnf_code NOT LIKE...AND` rather than `OR`
I've amended the code to just include `%inj%` which seems to work in tests.

richiecroker (changes requested):
I've reviewed and made further changes to code as per comments.  Can you review please @brianmackenna

brianmackenna (commented):
This looks good  and much simpler solution than initially proposed. Thanks!

richiecroker:
@inglesp can you merge this please - @brianmackenna and I are both happy.

inglesp (approved):

PR #2950, opened 2021-02-22
Use BSA's API to get URL of latest prescribing data

inglesp:


bloodearnest (approved):
Will pretty much always approve a removal of html scraping 💪

PR #2952, opened 2021-02-24
Don't delete Hospital Only AvailabilityRestriction

### LINKED ISSUE #2951: dm+d import failure
`openprescribing/deploy/fetch_and_import_dmd.sh` started consistently failing on Feb 22 with the following error:

```
psycopg2.errors.ForeignKeyViolation: update or delete on table "dmd_availabilityrestriction" violates foreign key constraint "dmd_amp_avail_restrictcd_4160c735_fk_dmd_avail" on table "dmd_amp"
DETAIL:  Key (cd)=(8) is still referenced from table "dmd_amp".
```

A clean fetch & import returns the clean version of the same error:

```
psycopg2.errors.ForeignKeyViolation: insert or update on table "dmd_amp" violates foreign key constraint "dmd_amp_avail_restrictcd_4160c735_fk_dmd_avail"
DETAIL:  Key (avail_restrictcd)=(8) is not present in table "dmd_availabilityrestriction".
```

Digging into the source XML files, we find in `f_lookup2_3180221.xml`:

```xml
 <AVAILABILITY_RESTRICTION>
  <INFO>
   <CD>0001</CD>
   <DESC>None</DESC>
  </INFO>
  <INFO>
   <CD>0002</CD>
   <DESC>Restricted Availability</DESC>
  </INFO>
  <INFO>
   <CD>0003</CD>
   <DESC>Individual Patient Supply</DESC>
  </INFO>
  <INFO>
   <CD>0004</CD>
   <DESC>Imported</DESC>
  </INFO>
  <INFO>
   <CD>0005</CD>
   <DESC>Clinical Trial</DESC>
  </INFO>
  <INFO>
   <CD>0006</CD>
   <DESC>Special</DESC>
  </INFO>
  <INFO>
   <CD>0007</CD>
   <DESC>Extemp</DESC>
  </INFO>
  <INFO>
   <CD>0009</CD>
   <DESC>Not available</DESC>
  </INFO>
 </AVAILABILITY_RESTRICTION>
```

which is missing  `8 │ Hospital Only`.

But plenty of references persist elsewhere:

```bash
$ grep "<AVAIL_RESTRICTCD>0008</AVAIL_RESTRICTCD>" f_amp2_3180221.xml  | wc
    1594    1594   71730
 ```

@inglesp to update the code & contact the upstream data provider

inglesp:
It has been removed from the latest dm+d release, but is still
referenced by ~1500 AMPs.

Fixes #2951

### DIFF
--- a/openprescribing/dmd/management/commands/import_dmd.py
+++ b/openprescribing/dmd/management/commands/import_dmd.py
@@ -14,7 +14,7 @@ from django.db.models import fields as django_fields
 from django.db.models import ForeignKey
 
 from dmd import models
-from dmd.models import AMP, AMPP, VMP, VMPP
+from dmd.models import AMP, AMPP, VMP, VMPP, AvailabilityRestriction
 from dmd.view_schema import schema
 from dmd.obj_types import cls_to_obj_type
 from frontend.models import ImportLog, Presentation
@@ -217,7 +217,12 @@ class Command(BaseCommand):
     def import_model(self, model, elements):
         """Import model instances from list of XML elements."""
 
-        model.objects.all().delete()
+        if model == AvailabilityRestriction:
+            # AvailabilityRestriction 8 (Hospital Only) has been removed from the dm+d
+            # releases, but is still referenced by ~1500 AMPs.
+            model.objects.exclude(pk=8).delete()
+        else:
+            model.objects.all().delete()
 
         boolean_field_names = [
             f.name


madwort:
@inglesp I guess `openprescribing/dmd/tests/data/dmd/1/f_lookup2_3010419.xml` needs updating for tests to pass?

madwort:
This branch seems like it would work in production, but the unit tests fail annoyingly - my suspicion is that the tests import dm+d twice without cleaning the db in between. The test data currently contains HospitalOnly, and a test fails because it attempts to re-import HospitalOnly without having deleted it first. However, if I /remove/ the test data, it fails because HospitalOnly is never imported & other parts of the test dataset still use it. I think the easiest option is to match what we're seeing in production - which is to remove it from test data, manually add it, and keep the rest of the test dataset that relies on it. I've done this in #2953

PR #2963, opened 2021-03-22
Increase max_length of PCT.code

inglesp:
In April 2021, new CCGs were created with 5-character codes


evansd (approved):
I think in the context of having no budget to fix this properly but wanting to keep it going as far as possible this is the best approach.

PR #2966, opened 2021-04-01
Update scipy to v1.5.4

madwort:
* minimum update to (attempt to) unblock dependabot


evansd (approved):
Having looked at the v1.5.0 [release notes](https://docs.scipy.org/doc/scipy/reference/release.1.5.0.html) I think is a safe upgrade for us, and given that the tests pass I'm reasonably confident.

madwort:
ansible CI is failing because of https://github.com/ebmdatalab/openprescribing/pull/2967 - in that test run it's using system python on debian/stretch, which means python-3.5 - switching to debian/buster gives us python-3.7 which should be compatible with this.

PR #2925, opened 2020-12-04
Run tests under 3.8

evansd:
Now we're running 3.8 in production we should test using it as well.

This also removes a warning that's no longer relevant.

Instructions for updating the Python version are now in wiki:
https://github.com/ebmdatalab/openprescribing/wiki/Installing-a-new-Python-version-for-OpenPrescribing

PR #2967, opened 2021-04-01
Upgrade ansible images to buster

madwort:
* this should upgrade system python to 3.7, when running ansible tests


evansd (approved):
If it's of any help, here's the PR which upgraded things to 3.8:
https://github.com/ebmdatalab/openprescribing/pull/2925

Issue #2959, opened 2021-03-11
Fix Dependabot CI


madwort:
@dependabot-bot no longer has access to secrets for CI runs, so they fail (e.g. #2957 )
See https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/
Probably this is worth a read too: https://securitylab.github.com/research/github-actions-preventing-pwn-requests

madwort:
Also - dependabot "security updates" (which are different to "version updates") are currently failing to install e.g. https://github.com/ebmdatalab/openprescribing/security/dependabot/requirements.txt/django/open/update-logs/79578134 - this appears to be due to attempting to install scipy 1.4.1 on python 3.9, which I think doesn't have a wheel available and is failing to compile.

benbc:
I've turned off dependabot alerts and dependabot security updates for this repo until this problem has been fixed.

benbc:
@sebbacon I think that we should consider fixing this problem even though there is no active development on OpenPrescribing at the moment because Dependabot is our best safeguard against vulnerabilities caused by our dependencies.

evansd:
Huge thread here on the issue: https://github.com/dependabot/dependabot-core/issues/3253 (with an [FAQ](https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544) at the end).

It seems like maybe we could do something with a `pull_request_target` workflow, but I haven't fully got my head around that.

madwort:
this is indeed potentially a can of worms - for good security reasons, github has disabled forks having access to github actions secrets. However, dependabot runs from a fork & we need github actions secrets to run our CI. 
There's not an obvious way to square the circle, but there are different compromises. 

For instance - we could use `pull_request_target`, which would allow forks to access github actions secrets by running the actions with the permissions of the target rather than the head of the branch (i.e. the permissions of `origin/master` not  `fork/update`). However, this introduces a potential security hole - a compromised upstream dependency could automatically pwn our secrets before we knew about it (which in this case I think includes a key to access our old Google Compute account vms/dbs). Upstream dependencies do get pwned sometimes, but it's very rare, maybe we take the risk. At GitHub scale, they're worried about this, which is why it's under heavy development.

At the moment, for security updates / vulnerabilities, I've been re-running CI with my own permissions, but not for "version updates" - of which there are *a lot*.

madwort:
[...]

PR #2968, opened 2021-04-01
Add .python-version file

madwort:
* test whether this will persuade dependabot to use the correct version
* 3.8.6 because that appears to be what is currently used in production


evansd (approved):
Worth a shot!

madwort:
@evansd what's worrying me about this one is that I have no idea whether it'll affect the production deployment...

evansd:
A valid concern, but it won't have any effect in this case.

madwort:
Great, it's worked - first unblocked PR here: https://github.com/ebmdatalab/openprescribing/pull/2969 
Now to fix https://github.com/ebmdatalab/openprescribing/issues/2959 ...

PR #3045, opened 2021-04-09
Reduce number of dependabot open PRs

madwort:
* at least until we get the backlog cleared a bit


evansd (approved):
It's a bit overwhelming atm, isn't it? :smile:

PR #3047, opened 2021-04-09
Use pysqlite3 instead of stdlib sqlite3

evansd:
This is so we can make use of the bundled version of SQLite in
pysqlite3-binary which (since [this PR][1]) has been compiled with a
much higher memory-map limit. Since, our 6G prescribing files were
exceeding the previous 2G limit its possible that this will have
significant performance benefits for us. The easiest way of finding out
is just to try it and see. We have nginx logging response times so a
before/after comparison should be fairly straightforward.

[1]: https://github.com/coleifer/pysqlite3/pull/22

### DIFF
--- a/openprescribing/matrixstore/tests/commands/test_matrixstore_build.py
+++ b/openprescribing/matrixstore/tests/commands/test_matrixstore_build.py
@@ -21,6 +21,7 @@ import json
 import numbers
 import shutil
 import sqlite3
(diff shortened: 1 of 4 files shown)


ghickman (approved):

sebbacon:
Add a calendar entry to check the times to find out?! Not super important, but interested!

PR #3050, opened 2021-04-13
Experimental PR to run tests for Dependabot PRs

evansd:
Dependabot PRs run without access to secrets so the tests (which rely on the BigQuery test account credentials) can't run.

However, because CI runs are attached to the commit ID it's possible to create a separate PR with the some commit which will cause the tests to run and then update the status of the original dependabot PR.

This is an experiment at a workflow where we leave this PR open and force push dependabot commits to it in order to force test runs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -196,7 +196,7 @@ protobuf==3.13.0
     #   googleapis-common-protos
 psycopg2-binary==2.8.6
     # via -r requirements.in
-pyarrow==2.0.0
+pyarrow==3.0.0
     # via -r requirements.in
 pyasn1-modules==0.2.8
     # via google-auth


evansd:
Gah, I now discover that unless we change the HEAD commit of the `dependabot-ci-runs` branch _before_ merging the original dependabot PR then this PR gets automatically closed.

madwort:
@evansd are you aware you can now give dependabot its own secrets - which I did - which initially worked (possibly for the `dependabot-security-updates` bot) - and then stopped working (possibly for the `dependabot-version-updates` bot)... see https://github.com/ebmdatalab/openprescribing/settings/secrets/dependabot

evansd:
@madwort Ah, I didn't realise this. That's great, thanks! I'll take a look

madwort:
Hmm, so just got around to checking that & I was entirely wrong - that will only pass secrets into `dependabot.yml`, whereas we want to pass secrets into the CI runs. So obviously something else was happening before.

PR #3051, opened 2021-04-13
Experimental PR to run tests for Dependabot PRs: attempt 2

evansd:
Dependabot PRs run without access to secrets so the tests (which rely on the BigQuery test account credentials) can't run.

However, because CI runs are attached to the commit ID it's possible to create a separate PR with the some commit which will cause the tests to run and then update the status of the original dependabot PR.

This is an experiment at a workflow where we leave this PR open and force push dependabot commits to it in order to force test runs.

It's important that _before_ merging the original dependabot PR we force push another, different commit to this branch to prevent this PR from being automatically marked as merged.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -162,7 +162,7 @@ mock==3.0.5
     # via -r requirements.in
 networkx==2.4
     # via -r requirements.in
-numpy==1.18.5
+numpy==1.20.2
     # via
     #   -r requirements.in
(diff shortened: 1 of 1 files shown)


evansd:
Weird workaround no longer needed, see: https://github.com/ebmdatalab/openprescribing/issues/2959#issuecomment-859584770

PR #3060, opened 2021-04-21
Use django-2.2.20

madwort:
* fix `CVE-2021-28658`
* see also https://github.com/ebmdatalab/openprescribing/security/dependabot/requirements.txt/Django/open

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -14,7 +14,7 @@ django-cors-headers
 django-crispy-forms
 django-dotenv
 django-import-export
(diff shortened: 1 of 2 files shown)


madwort:
Annoyingly, dependabot didn't create a PR for this, because "there is already a PR for version 3.0", although that PR is closed

madwort:
("dependabot" now refers to at least 3-4 different systems, which may or may not correctly interoperate with each other...)

PR #3063, opened 2021-04-23
Handle missing drug tariff data for a month

### LINKED ISSUE #3059: All England alerts - concession information showing zero
In the latest version of the All England alert concession information is showing as zero which based on the [dashboard](https://openprescribing.net/national/england/concessions/) should be showing a figure circa £4million

image

evansd:
In this case, even if we have price concession data for this month we
can't use that data because we don't have the standard tariff price
against which to compare it.

I don't know if we _should_ have tariff data and something has broken
with the import. But either way this feels like something we should
handle gracefully.

Fixes #3059


inglesp (approved):
Looks plausible!

PR #3075, opened 2021-05-04
Fix fetch_dmd

inglesp:
A line that used to say eg

        Released on Monday, 3 May 2021

now says eg

        Released on Monday, 3 May 2021.


evansd (approved):
:face_with_head_bandage:

PR #3101, opened 2021-06-01
Add fields to measures for OpenCodelists codelist names

inglesp:
In most cases, the new fields are derived from the existing
numerator_short or denominator_short fields.

A couple of measures have XXXs that will need to be updated by a clueful
clinician.


richiecroker:
@inglesp i've made changes to XXX, except for denominator in `ktt9_uti_antibiotics.json` as the numerator and denominator are the same (numerator based on quantity)

inglesp:
Great, thanks.  Should the other `kttX` codelists have names beginning with "NICE KTTX", in the same way as you've done for ktt9_cephalosporins?

PR #3114, opened 2021-06-10
chore(deps): Use django 2.2.22

madwort:


madwort:
Force merged, because this is like a dependabot PR, and tests pass.

PR #3152, opened 2021-07-07
Fix SNOMED map fetcher to handle multiple urls

evansd:
The BSA have now started including links to historical versions of the
mapping which, while helpful in itself, breaks our assumption that
there's a single ZIP file. We now find all matching ZIP files and use
the latest.


inglesp (approved):
Thanks!

PR #3222, opened 2021-09-13
Upgrade Numpy and Pandas

evansd:
There's an interdependency here which means that upgrading them individually doesn't work.

It also turns out that something subtle changed in the rounding behaviour of Pandas when doing a `groupby`/`sum` in v1.3.0, though I couldn't spot an obvious smoking gun in the release notes:
https://pandas.pydata.org/docs/whatsnew/v1.3.0.html

However here's a small-ish reproducing example which finds the sum of
each of two lists, both with and without using `groupby`:
```py
import pandas

group_1 = [0.156, 0.118, 0.153, 0.149, 0.112]
group_2 = [0.126, 0.110, 0.160, 0.137, 0.155]

df1 = pandas.DataFrame(
    {
        1: group_1,
        2: group_2,
    }
)

df2 = pandas.DataFrame(
    {
        "group": [1] * len(group_1) + [2] * len(group_2),
        "value": group_1 + group_2,
    }
)

df1_sum = df1.sum()
df2_sum = df2.groupby("group")["value"].sum()

print(f"Pandas v{pandas.__version__}")
print("sum", df1_sum.to_list())
print("groupby/sum", df2_sum.to_list())
```

Running this under Pandas 1.2.5 and 1.3.0 gives:

    Pandas v1.2.5
    sum [0.6880000000000001, 0.6880000000000001]
    groupby/sum [0.6880000000000001, 0.6880000000000001]

    Pandas v1.3.0
    sum [0.6880000000000001, 0.6880000000000001]
    groupby/sum [0.688, 0.6880000000000001]

The values in question come from the list sizes of the practices in two PCNs which we generate in the tests. Under Pandas <= 1.2.5 these summed as equal to each other and, as the numerators in question were also equal, the two PCNs had identical measure ratios. This meant they were assigned equal rank and thus equal percentile. Under Pandas >= 1.3.0 they are no longer equal and thus are given different rank and a more than trivially different percentile.

We avoid this problem by rounding to 10 decimal places after every groupby-sum operation.  However it feels like there's a wider question here as to what's going on with Pandas that it might be useful to get to the bottom of.


evansd:
Here's the one unit test failure:
```
FAIL: test_practice_statistics_measure (frontend.tests.commands.test_import_measures.ImportMeasuresTests)
  File "/code/openprescribing/frontend/tests/commands/test_import_measures.py", line 156, in test_practice_statistics_measure
    self.validate_calculations(
  File "/code/openprescribing/frontend/tests/commands/test_import_measures.py", line 315, in validate_calculations
    self.validate_measure_value(mv, pcns.loc[mv.pcn_id])
  File "/code/openprescribing/frontend/tests/commands/test_import_measures.py", line 375, in validate_measure_value
    self.assertAlmostEqual(mv.percentile, series["ratio_percentile"])
AssertionError: 18.5483870967742 != 19.35483870967742 within 7 places (0.8064516129032206 difference)
```

StevenMaude:
I'll approve as the actual source changes in this PR look fine. It would be good to understand the behaviour difference, though.

None of the marked, and mostly closed, regression issues seemed to obviously match from a quick look at the [pandas issue tracker](https://github.com/pandas-dev/pandas/issues?q=is%3Aissue+REGR+groupby). Maybe worth just opening one there, if none of the existing issue match, and seeing what the authoritative response on the correct behaviour here is?
 
Bisecting is a bit tedious here. Even though there are only a few hundred commits between pandas 1.2.5 and 1.3.0, pandas will need building each time.

StevenMaude (approved):

evansd:
Thanks @StevenMaude. I've since updated this PR to solve the problem using rounding rather than relying on the tweaking the random seed to avoid problematic examples.

inglesp (approved):
Looks fine.  Perhaps a less-invasive change would to be use `assert_almost_equal`?

evansd:
> Perhaps a less-invasive change would to be use `assert_almost_equal`?

Ah, that isn't possible but I don't think I did a good job of explaining why. The problem is that the very very slightly different denominators cause the ratios to be very slightly different with then causes the ranking and corresponding percentile to be much more than slightly different.

inglesp:
> > Perhaps a less-invasive change would to be use `assert_almost_equal`?
> 
> Ah, that isn't possible but I don't think I did a good job of explaining why. The problem is that the very very slightly different denominators cause the ratios to be very slightly different with then causes the ranking and corresponding percentile to be much more than slightly different.

Of course.

PR #3245, opened 2021-11-03
Send Slack notifications to #technoise

inglesp:
This will affect the overnight fetching jobs, the end-to-end test notifications, and deploy notifications.

I've added `SLACK_TECHNOISE_POST_KEY` to the environment variables file.

### DIFF
--- a/deploy/notify_deploy.py
+++ b/deploy/notify_deploy.py
@@ -10,7 +10,7 @@ def notify_slack(message):
     # Set the webhook_url to the one provided by Slack when you create
     # the webhook at
     # https://my.slack.com/services/new/incoming-webhook/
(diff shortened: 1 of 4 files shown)


evansd (approved):
Thanks for doing this.

I think ideally we wouldn't have the name of the channel in the environment variable at all. It's just a "post to Slack" URL and it's the webhook configuration, not the application, which defines which channel it ends up in. But I'm not about to insist that you change it :smile:

inglesp:
> I think ideally we wouldn't have the name of the channel in the environment variable at all. It's just a "post to Slack" URL and it's the webhook configuration, not the application, which defines which channel it ends up in

Am on the fence -- the webhook URL is linked to a channel, so this leaves it open for different webhooks to go to different places.  But I've already spent too long thinking about this!

PR #3253, opened 2021-11-16
Add in capsules and chewtabs

richiecroker:
Following email from user, small number of rosuvastatin caps now prescribed, which weren't excluded from numerator originally.  Have also added in atorvastatin 20mg chewtabs, and rearranged presentations so in strength order, therefore easier to review.


brianmackenna (approved):
It has been so long since I reviewed a change on OP - feel rusty but this all looks sensible. Does staging still work so we can have brief check?

Issue #2617, opened 2020-04-29
Lutein and Antioxidants Measure showing (nearly) zero


brianmackenna:
The [lutein and antioxidant measure](https://openprescribing.net/measure/lplutein/national/england/#lplutein) is showing (nearly) zero  since October. 

I have checked BQ and it appears raw data has these low numbers
<details><summary>Code snippet</summary>
<p>

```SELECT 
     SUM(actual_cost) AS cost,
     SUM(quantity) AS QTY,
     SUM(items) AS item
 FROM ebmdatalab.hscic.prescribing_2020_01
 WHERE bnf_code IN ("091000000BBTAA0", "091000000BBTPA0", "091000000BBUQA0", "091000000BBUSA0", "091000000BBUZA0", "091000000BBVFA0", "091000000BBVGA0", "091000000BBVQA0", "091000000BBVSA0", "091000000BBVTA0", "091000000BBVXA0", "091000000BBVZA0", "091000000BBWHA0", "091000000BBWNA0", "091000000BBWPA0", "091000000BBWQA0", "091000000BBWVA0", "091000000BBWWA0", "091200000BEJYA0", "091200000BEXXA0", "091200000BFLHA0")
GROUP BY
practice
```

</p>
</details>

This makes me think it might be to do with change in BNF mapping @inglesp ? And we are still awaiting the new mapping file from BSA.

Unless someone else knows the reason off top of their head this is low priority for investigation until at least after BSA release the new file.

inglesp:
Nothing to do with the BNF/SNOMED mapping, since `lplutein` is defined in terms of BNF codes and not SNOMED codes.

Looks like most of the presentations in the measure definition have changed their BNF code between 2019_09 and 2019_10.

For instance, "PreserVision tablets" used to have BNF code `091000000BBTPA0` and now have BNF code `091000000BBTPBU`.

In the past, we have been given a file that lists all such BNF code changes by the BSA.  I'll ask whether they can provide us with the latest list of BNF code changes.

brianmackenna:
The BSA has provided new BNF code to old BNF code maps at https://www.nhsbsa.nhs.uk/bnf-code-changes

This has been changed in the database but we still need to rewrite the measure definition for [lutein and antioxidants](https://github.com/ebmdatalab/openprescribing/blob/master/openprescribing/measure_definitions/lplutein.json).

richiecroker:
Closed via #3256

PR #3256, opened 2021-11-18
update BNF codes

richiecroker:
Following email from user stating products were missing, BNF codes appear to have changed for the majority of products.  I have updated codes and added a couple of new products in.


brianmackenna:
Just a quick pass to say it relates to relates to #2617. will do a proper review in due course

brianmackenna:
@richiecroker did you generate the list of new BNF codes/preps manually? or did you do something systematic?

It looks good to me but I worry I may have missed a few when manually checking selections from `Multinutrient tablets` based on brand names.

richiecroker:
I generated them manually, and compared with previous list

brianmackenna (approved):
Happy with this. It's unfortunate that we can't programmatically identify new preparations here. A measure that definitely requires manual review every year or two.

PR #3257, opened 2021-11-18
Amend smi greeninhaler

brianmackenna:
I have been writing about greener NHS today and decided to tackle a long outstanding issue to programmatically account for soft mist inhalers (see #2144)  We previously manually exempted soft mist inhalers from the numerator as they were incorrectly identified in dm+d as MDIs. This appears to have been rectified .

I'm not sure why I didn't address in #2636 but I suspect as May 2020 I was being pragmatic.

It has been quite awhile since I tackled measures on OP so it would be good if someone could code review this a bit more thoroughly than normal.

I think it would also be useful to check this measure on staging before deployment.

### DIFF
--- a/openprescribing/measure_definitions/environmental_inhalers.json
+++ b/openprescribing/measure_definitions/environmental_inhalers.json
@@ -39,11 +39,7 @@
   ],
   "numerator_bnf_codes_filter": [
     "03",
-    "~0301011R0%",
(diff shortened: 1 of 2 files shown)


richiecroker (commented):
See comment on commit about changes

richiecroker (approved):

PR #3259, opened 2021-11-24
Support previewing measures directly from Github

evansd:
This adds a pair of commands:

    ./manage.py preview_measure [url_of_measure_file_on_github]
    ./manage.py delete_measure [measure_id]

The `preview_measure` command rewrites the measure ID to have the prefix `preview_` (so it's possible to use this to test a new version of an already existing measure) and removes any tags and alert configuration. There's also a little bit of nasty logic to hide it from the "all measures" view. It's still accessible to anyone who knows the URL, but we don't want to cause confusion by showing it by default.

The `delete_measure` command has a safety check so it will only work for measures that have the `preview_` prefix. 

This is not a thing of beauty but it was a reasonably quick way to unlock the workflow we need.

### DIFF
--- a/openprescribing/frontend/management/commands/delete_measure.py
+++ b/openprescribing/frontend/management/commands/delete_measure.py
@@ -7,11 +7,13 @@ from gcutils.bigquery import Client
 
 class Command(BaseCommand):
     def handle(self, measure_id, **options):
-        if not measure_id.startswith(settings.MEASURE_PREVIEW_PREFIX):
-            raise CommandError(
-                f"Not deleting '{measure_id}' because it doesn't look like a preview "
-                f"measure (it doesn't start with '{settings.MEASURE_PREVIEW_PREFIX}')"
-            )
+        if not options["delete_live_measure"]:
+            if not measure_id.startswith(settings.MEASURE_PREVIEW_PREFIX):
+                raise CommandError(
+                    f"Not deleting '{measure_id}' because it doesn't look like a "
+                    f"preview measure (it doesn't start with "
+                    f"'{settings.MEASURE_PREVIEW_PREFIX}')"
+                )
         try:
             measure = Measure.objects.get(id=measure_id)
         except Measure.DoesNotExist:
@@ -24,6 +26,7 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument("measure_id")
+        parser.add_argument("--delete-live-measure", action="store_true")
 
 
 def delete_from_bigquery(measure_id):
--- a/openprescribing/frontend/management/commands/delete_measure.py
+++ b/openprescribing/frontend/management/commands/delete_measure.py
@@ -1,0 +1,24 @@
+from django.conf import settings
+from django.core.management import BaseCommand, CommandError
+
+from frontend.models import Measure
+
+
+class Command(BaseCommand):
+    def handle(self, measure_id, **options):
+        if not measure_id.startswith(settings.MEASURE_PREVIEW_PREFIX):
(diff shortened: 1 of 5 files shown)


openprescribing/frontend/management/commands/preview_measure.py
```diff
@@ -0,0 +1,110 @@
...
+from django.core.management import BaseCommand, CommandError
+from requests.exceptions import InvalidJSONError, RequestException
+
+from .import_measures import BadRequest
+from .import_measures import Command as ImportMeasuresCommand
+from .import_measures import ImportLog, relativedelta
+
+
+class Command(BaseCommand):
+    def handle(self, measure_url, **options):
```
inglesp, line 14: Suggest changing `measure_url` to `github_url`.

openprescribing/frontend/management/commands/delete_measure.py
```diff
@@ -0,0 +1,24 @@
...
+                f"measure (it doesn't start with '{settings.MEASURE_PREVIEW_PREFIX}')"
+            )
+        try:
+            measure = Measure.objects.get(id=measure_id)
+        except Measure.DoesNotExist:
+            raise CommandError(f"No measure with ID '{measure_id}'")
+        # The ON DELETE CASCADE configuration ensures that all MeasureValues are deleted
+        # as well
+        measure.delete()
+        self.stdout.write(f"Deleted measure '{measure_id}'")
```
inglesp, line 21: Should we delete from BQ as well?

openprescribing/frontend/management/commands/preview_measure.py
```diff
@@ -19,7 +19,11 @@ def handle(self, measure_url, **options):
                 f"Importing measure preview failed for {measure_url}\n\n{e.message}"
             )
         measure_url = f"https://openprescribing.net/measure/{measure_id}/"
-        self.stdout.write(f"Measure can be previewed at:\n{measure_url}")
+        self.stdout.write(
+            f"Measure can be previewed at:\n{measure_url}\n\n"
+            f"When you've finished remember to delete the preview with:\n"
+            f"@ebmbot op delete_measure {measure_id}"
```
inglesp, line 25: Could this be `delete_preview_measure`?

inglesp (approved):
Some small suggestions, but otherwise looks great.

openprescribing/frontend/management/commands/preview_measure.py
evansd: I went back and forth on this a bit, but maybe that is better actually. Do you think it would be sensible to make the ebmbot command `delete_preview_measure` but leave the Django command as is? I can sort of imagine us wanting to add a not-just-previews flag to that command for our own use later on.

openprescribing/frontend/management/commands/delete_measure.py
evansd: Oh yes, I'd completely forgotten about that. Do you know off the top of your head how we do that?

openprescribing/frontend/management/commands/preview_measure.py
inglesp: Leaving this as is, changing the ebmbot command would be fine.

openprescribing/frontend/management/commands/delete_measure.py
inglesp: Everything you need is in `gcutils.bigquery`.  If you know the name of the tables, you can call `Client.delete_table`.  Otherwise you can iterate over all tables with `Client.list_tables`, and delete those matching the prefix.

PR #3260, opened 2021-11-24
Write user-facing errors to stdout rather than stderr

evansd:
This is not particularly UNIXy, but the ebmbot system won't show the
stderr output and we need the user to see these messages.


ghickman (approved):

evansd:
Now I've got this horrible feeling that even if I write to stdout, if the command fails it won't show any output ....

PR #3261, opened 2021-11-24
Exit zero even on failure :(

evansd:
This is currently the only way to get error message back to the user if they run
via ebmbot. We should probably fix this.

### DIFF
--- a/openprescribing/frontend/management/commands/delete_measure.py
+++ b/openprescribing/frontend/management/commands/delete_measure.py
@@ -11,19 +11,19 @@ class Command(BaseCommand):
     def handle(self, measure_id, **options):
         if not options["delete_live_measure"]:
             if not measure_id.startswith(settings.MEASURE_PREVIEW_PREFIX):
-                # We want these errors to be visble to users who run via ebmbot, but
-                # ebmbot doesn't show stderr so we can't just raise CommandError here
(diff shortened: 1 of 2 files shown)


inglesp (approved):
Sorry, I expect this is all my fault.

evansd:
Not at all, I expect all the decisions made sense in their context; I'm just being grumpy because I thought I had finished this two hours ago.

openprescribing/frontend/management/commands/delete_measure.py
```diff
@@ -11,19 +11,19 @@ class Command(BaseCommand):
     def handle(self, measure_id, **options):
         if not options["delete_live_measure"]:
             if not measure_id.startswith(settings.MEASURE_PREVIEW_PREFIX):
-                # We want these errors to be visble to users who run via ebmbot, but
-                # ebmbot doesn't show stderr so we can't just raise CommandError here
+                # We want these errors to be visble to users who run via ebmbot but the
+                # only way to achieve that is to write them to stderr and exit 0 :(
```
bloodearnest, line 15: nit: comment doesn't match reality.

I think the comment (and the the one below) should `s/stderr/stdout

inglesp:
>  I thought I had finished this two hours ago

don't say that out loud!

Issue #3263, opened 2021-11-26
Django < 3.1 is incompatible with GDAL 3


evansd:
We use GDAL (implicitly) [here](https://github.com/ebmdatalab/openprescribing/blob/87f21c585e8d81ec081c745ff85e764e8e594138/openprescribing/frontend/management/commands/import_ccg_boundaries.py#L81) when loading CCG boundary data. It's possible that we use it in other places too.

Django uses whatever version of GDAL it can find. GDAL version 3 switches axis order. This means that commands appear to run without error but all locations are rotated 90 degrees around Null Island:
https://code.djangoproject.com/ticket/30645#comment:5

This was [fixed](https://github.com/django/django/commit/58f1b07e49) in Django 3.1, however we can't currently upgrade Django and I haven't yet figured out how much work that will be (see #3262).

This work OK in production at the moment because the ancient base OS we're on still has GDAL 2, but this is a problem running locally as newer OSs only ship GDAL 3.

It's possible to workaround this by compiling on old version of GDAL yourself and telling Django to use it. Briefly

 * Download GDAL 2 from here: https://gdal.org/download.html (I used version [2.4.4](https://download.osgeo.org/gdal/2.4.4/gdal-2.4.4.tar.gz))

 * Untar 
 
 * Follow the non-root installation [instructions](https://trac.osgeo.org/gdal/wiki/BuildingOnUnix), which are basically just:
    ```
    ./configure --prefix="$(dirname $PWD)/gdal"
    make
    make install
    ```
    This took forever to compile. And hour or something maybe -- I didn't sit and watch it.

 * Then you can tell Django to use it with:
    ```
    GDAL_LIBRARY_PATH="$PWD/gdal/lib/libgdal.so"
   ```

PR #3264, opened 2021-11-26
Update CCG boundaries and fix import

evansd:
This updates the test fixture to use the latest CCG boundaries, obtained using [this script](https://github.com/ebmdatalab/fetch-boundaries/tree/2a9d0f711a81c1dbe68ba4874071545256b98c14).

It also fixes the import so that it updates the org models in-place rather than replacing them and contains some workarounds for a nasty [GDAL incompatibility](https://github.com/ebmdatalab/openprescribing/issues/3263).


ghickman (approved):
All looks sensible 👍

sebbacon:
Great. Is there somewhere we could make some nice public acknowledgements for the NHSX pals who helped out here?

sebbacon:
Also I think the whole docstring in the management commands describing the manual rigmarole could be replaced by a link to that fetcher script?

evansd:
> Is there somewhere we could make some nice public acknowledgements for the NHSX pals who helped out here?

Yes, happy to but not sure where. I'll update the original issue once this is deployed and the boundaries imported.

PR #3269, opened 2021-12-06
Update measure_tags.json

### LINKED ISSUE #3226: Remove IIF Dashboard
There have been substantial developments to NHS England Investment and Impact fund. We previously launched a dashboard

https://openprescribing.net/regional-team/Y56/measures/?tags=iif

but the most recent IIF appears to utilise nearly all patient level metrics which we cannot reproduce from NHS open data.

They are listed at https://www.england.nhs.uk/wp-content/uploads/2021/08/B0828-iii-annex-b-investment-and-impact-fund-21-22-and-22-23.pdf

We could
a) leave the dashboard there for the measures we can reproduce (on a quick skim it looks like 1 or 2) with a note in the description
b) remove the dashboard

@richiecroker any preference?

richiecroker:
removed IIF tag as per #3226 

Closes #3226

### DIFF
--- a/openprescribing/common/measure_tags.json
+++ b/openprescribing/common/measure_tags.json
@@ -51,10 +51,6 @@
     "name": "Infections",
     "description": "These measures are for treatments commonly used for treating infections, including antibiotics and dressings."
   },
-  "iif": {
-    "name": "NHS England Investment and Impact Fund",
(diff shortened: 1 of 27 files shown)


evansd:
We'll have to remove the references to the tags as well, which I've just done as it was an easy find-and-replace.

evansd (approved):

PR #3276, opened 2021-12-17
Raise more helpful errors from `infer_practice_boundaries`

evansd:
We've recently hit errors here because one of our data imports seems to have included a load of Welsh practices. But it was hard to see exactly what the problem was because the error raised was not very informative. This makes them more helpful.


openprescribing/frontend/management/commands/infer_practice_boundaries.py
```diff
@@ -77,30 +80,54 @@ def cursor_execute(sql, *params):
...
+        try:
+            clipped = region.intersection(clip_boundary)
+            # This is a workaround for the error "Relate Operation called with a
+            # LWGEOMCOLLECTION type" which happens when clipped boundaries end up
+            # including things which aren't polygons (i.e. points or lines) and we then
+            # try to do ST_Contains queries on them. Generating a zero-width buffer
+            # causes all non-polygons to get dropped. See:
+            # https://lists.osgeo.org/pipermail/postgis-users/2008-August/020740.html
+            # (Why clipping would result in non-polygons in some cases is totally
+            # unclear, but somehow it does.)
```
iaindillingham, line 95: Possibly to leave it up to the user what to do with the non-polygons? For example, if I created a Voronoi tessellation of points representing postcode units and then clipped this with a polygon representing the UK landmass, then I'd not want the leftovers made into new polygons ([sliver polygons](https://en.wikipedia.org/wiki/Sliver_polygon)). Leaving them as lines would make them easier to filter out.

Perhaps I've misunderstood the issue, though.

iaindillingham (approved):

openprescribing/frontend/management/commands/infer_practice_boundaries.py
evansd: Well there isn't really a "user" in this case: it's run as part of the import pipeline so it all needs to be automated. But my understanding (as per the linked postgis thread) is that calling `buffer(0.0)` _does_ discard any non-polygons, which I think is what we want.

PR #3280, opened 2021-12-21
Skip Welsh practices on import

### LINKED ISSUE #3279: epraccur.csv includes Welsh practices which we do not want to import
These additional practice are "mostly harmless" because we don't have any corresponding prescribing data for them. However there is at least one case where they cause an issue and that is when we attempt to infer practice boundaries by calculating a Voronoi partition over all active practices. (Because we clip practice boundaries to the English national boundary we end up assigning Welsh practices an empty boundary which triggers an error.) I can also imagine there might be other issues we haven't spotted yet so it would be safer to exclude these completely.

I believe we can do this by excluding all practices in [`epraccur.csv`](https://digital.nhs.uk/services/organisation-data-service/file-downloads/gp-and-gp-practice-related-data) where column 3 (National Grouping) equals `W00`.

https://datadictionary.nhs.uk/supporting_information/organisation_coding_frames.html

datadictionary nhs uk_supporting_information_organisation_coding_frames html

evansd:
We don't (yet) get any prescribing data for these and their presence
causes at least one issue.

There'll be some manual work to remove the practices we've already imported.

Closes #3279

### DIFF
--- a/openprescribing/frontend/management/commands/import_practices.py
+++ b/openprescribing/frontend/management/commands/import_practices.py
@@ -47,6 +47,13 @@ class Command(BaseCommand):
         count = 0
         for row in entries:
             row = [r.strip() for r in row]
+
+            # Skip Welsh practices, see:
+            # https://github.com/ebmdatalab/openprescribing/issues/3279
+            national_grouping = row[2]
+            if national_grouping == "W00":
+                continue
+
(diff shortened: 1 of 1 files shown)


inglesp (approved):
:england: :england: :england:

PR #3352, opened 2022-01-20
Add command to delete latest prescribing import

inglesp:


openprescribing/pipeline/management/commands/delete_latest_prescribing_import.py
```diff
@@ -0,0 +1,166 @@
+"""
+This command:
+
+    * deletes all prescribing data from:
+      * the filesystem
+      * BigQuery
+      * Cloud Storage
+      * the matrixstore
```
evansd, line 8: I'm not sure it's quite right to say "the matrixstore" here. It certainly took me a bit by surprise and then I had to confirm by looking at the code that it didn't actually do this. I think it might be more accurate to say:
```suggestion
    * deletes all prescribing data (both original data and extracts created by the matrixstore build) from:
      * the filesystem
      * BigQuery
      * Cloud Storage
```

openprescribing/pipeline/management/commands/delete_latest_prescribing_import.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+    client.run_job("query", [sql])
+
+
+def delete_backup_from_storage(year, month):
+    _delete_file_from_storage("backup/prescribing_v2/2021_10")
+
+
+def delete_matrixstore_bq_table(year, month):
+    _delete_file_from_storage("prescribing_export", f"prescribing_{year}_{month}")
```
evansd, line 136: I think this should be:
```suggestion
    _delete_table_from_bq("prescribing_export", f"prescribing_{year}_{month}")
```

evansd (approved):
Thanks very much for doing this. Apart from one typo'd function name I think it all looks correct. There are just a few optional suggestions I'd make.

1. I feel like it might be worth the extra upfront effort to make this idempotent. With all the various interactions that happen with GCS and BQ towards the end it's not entirely impossible that one of these will hang or error out of something like that. And then recovering becomes an awkward job of working out how far it got and commenting out the relevant bits.

2. It might be worth sticking a bunch of `prints()` in there so it's obvious what it's up to when you're running it. That also makes it more obvious how far it's got if you do have to comment things out to re-run it.

3. Consider using [these functions](https://github.com/ebmdatalab/openprescribing/blob/b78b21c5cd68500ca7d0f445bcaf7c90b212e1ff/openprescribing/matrixstore/build/download_prescribing.py#L224-L252) which define where the matrixstore build process puts stuff. Given how infrequently this command is going to get used there's a high chance of it going stale. And I think the coupling here would help guard against at least one way of that happening. In the same vein it's probably also worth making [`prescribing_export`](https://github.com/ebmdatalab/openprescribing/blob/b78b21c5cd68500ca7d0f445bcaf7c90b212e1ff/openprescribing/matrixstore/build/download_prescribing.py#L41) a module constant that can be imported here.

evansd:
:+1: reasonable responses on all fronts — merge at will!

PR #3360, opened 2022-01-27
Fix derivation of national boundary

evansd:
Previously practices too close to the border, and some too close to the
Thames in Central London, would end up outside of the national boundary
which triggered errors when trying the infer practice boundaries.


madwort:
the new boundary file looks better. I think it's fine, but that you could go a little heavier on the buffer if you wanted - see (literal) edge case comparisons:
before-sennen
after-sennen
before-woolwich
after-woolwich

evansd:
Ah good spot @madwort, thanks for checking. I originally had far too much buffer and then reduced it while checking that the import still worked, but I think I took it too far. I've bumped it up now and checked it looks more reasonable.

madwort:
Thanks, looks much better.

madwort (approved):

PR #3442, opened 2022-03-25
Remove ghost_branded_generics measure and associated view

evansd:
The underlying issue which led to the Ghost-Branded Generics problem has
been fixed.

After deployment the measure will need manually deleting from the db with:
```
./manage.py delete_measure --delete-live-measure ghost_generic_measure
```

For hygiene's sake we should probably delete the BQ view as well: `vw__ghost_generic_measure`.


inglesp:
> The underlying issue which led to the Ghost-Branded Generics problem has been fixed.

Impact!

inglesp (approved):

PR #3457, opened 2022-03-31
Add project and dataset placeholders to measure SQL

### LINKED ISSUE #3454: New views for opioid measure don't use placeholders
Specifically the view SQL files added in 
 * #2907
 * [`measure_sql/vw__opioids_total_ome.sql`](https://github.com/ebmdatalab/openprescribing/blob/3c246974c13bfddac03a2ab9350b3a9ca2a2ce1d/openprescribing/frontend/management/commands/measure_sql/vw__opioids_total_ome.sql)
 * [`measure_sql/vw__opioids_total_dmd.sql`](https://github.com/ebmdatalab/openprescribing/blob/3c246974c13bfddac03a2ab9350b3a9ca2a2ce1d/openprescribing/frontend/management/commands/measure_sql/vw__opioids_total_dmd.sql)

don't use `{project}.{dataset}` placeholders like the other views do. For example:
https://github.com/ebmdatalab/openprescribing/blob/3c246974c13bfddac03a2ab9350b3a9ca2a2ce1d/openprescribing/frontend/management/commands/measure_sql/vw__herbal_list.sql#L6-L9

I _think_ this will still work OK in production as the project will already be set correctly and the datasets will have the right names. But it breaks the end-to-end tests (see [Sentry error](https://sentry.io/organizations/ebm-datalab/issues/3140297576/?project=1254683)). And it might cause us further problems down the road so we should fix it.

From a quick scan through the SQL it looks like the new views might use datasets which are not currently listed in the `DATASETS` config:
https://github.com/ebmdatalab/openprescribing/blob/3c246974c13bfddac03a2ab9350b3a9ca2a2ce1d/openprescribing/gcutils/bigquery.py#L21-L30

But I'm not quite sure if that's true, or what the consequence would be if it was!

inglesp:
We check for the project placeholder ("{project}" ) to ensure that we
don't hardcode "embdatalab" as the project name, and so the project
placeholder must be present in any SQL that is run against BQ.  (This is
not failsafe, but it's a good reminder.)

When running against the "ebmdatalabtest" project, we give each test run
its own set of datasets, identified by a nonce.  The full names of the
dataset (eg "hscic_1234") replaces the dataset placeholders.

Fixes #3454.

### DIFF
--- a/openprescribing/frontend/management/commands/measure_sql/vw__opioids_total_dmd.sql
+++ b/openprescribing/frontend/management/commands/measure_sql/vw__opioids_total_dmd.sql
@@ -14,8 +14,8 @@ WITH simp_form AS (
       STRPOS(form.descr, ".")+ 1) #takes the dosage form out of the string (e.g. tablet.oral) TO leave route.
     END AS simple_form 
   FROM 
-    dmd.ont AS ont #the coded route for dosage form, includes vmp code 
-    INNER JOIN dmd.ontformroute AS form ON form.cd = ont.form #text description of route
+    {project}.{dmd}.ont AS ont #the coded route for dosage form, includes vmp code 
+    INNER JOIN {project}.{dmd}.ontformroute AS form ON form.cd = ont.form #text description of route
     )
 
 #subquery to normalise strength to mg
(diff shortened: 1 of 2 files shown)


evansd (approved):
Thanks very much!

PR #3479, opened 2022-04-12
Add GA + Plausible tracking code to outlier pages

tomodwyer:
So that we can analyse usage of those pages.


evansd (approved):
Looks good to me.

@Jongmassey what's the procedure for updating the rendered HTML from these templates?

PR #3480, opened 2022-04-12
fix: remove labs/outlier links for regions

Jongmassey:
We haven't produced outlier reports for regional teams yet so having a (broken) link to their outlier reports on their home page makes no sense

### DIFF
--- a/openprescribing/templates/entity_home_page.html
+++ b/openprescribing/templates/entity_home_page.html
@@ -1,249 +1,223 @@
-{% extends "base.html" %}
-{% load humanize %}
-{% load static %}
-{% load template_extras %}
-
-{% block title %}Prescribing measures for {{ entity.name }}{% endblock %}
-{% block active_class %}dashboards{% endblock %}
-{% block container_class %}landing-page{% endblock %}
-
-{% block extra_css %}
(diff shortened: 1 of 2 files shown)


openprescribing/templates/entity_home_page.html
```diff
@@ -187,13 +187,15 @@ <h3>Looking for more data?</h3>
   </div>
   <div class="col-md-6">
     <div class="landing-panel">
+      {% if entity_type!="regional_team" %}
```
evansd, line 190: I think it would make more sense to have this logic in the view and set `entity_outlier_report_url` to empty for regional teams. Then the check here becomes `{% if entity_outlier_report_url %}`.

evansd (approved):
I've made one suggestion but nothing critical.

I think it would have been better to force push over the commit which changed the formatting, rather than add another commit which changes it back because that will make `git blame` harder to work with in future. But in this particular case we can avoid that by doing a squash merge of the PR so no need to do anything this time.

openprescribing/templates/entity_home_page.html
Jongmassey: Agreed

PR #3523, opened 2022-05-10
add link to audits for greener practice

brianmackenna:
I have proposed Greener Practice link as it is useful resource with practical recommendations by practices, similar to how we link to Opioids Aware. The group already uses OP as a key part of their audit recs and is being promoted by NHSE to support with IIF


a-d-brown (approved):
Looks good to me!

PR #3683, opened 2022-09-21
Use ODS codes for STPs, and update how STPs are imported

inglesp:


evansd (approved):
Thanks for doing this. And very happy that the diff stats were so misleading as to the size of the code changes!

This looks sensible to me. I don't think it's unreasonable for STP bookmarks to break, given that STPs technically no longer exist anyway.

Obviously the deployment will need a bit of babysitting but I guess you've got plans for doing that.

PR #3685, opened 2022-09-23
Replace CCGs and STPs in URLs and text

inglesp:
This PR does not change any variable names, nor any magic strings used for identifying organisation type.  Doing so would be quite a big job with significant risk.  We know that refactoring organisation-related code is a high priority, if we ever come back to serious development on the project.

I'm confident that the measures code is well covered by tests.  I'm less confident in the organisation homepages and the analyse form, so these will need thorough manual testing on deploy.

As there are no database changes, rolling this PR back will be easy and quick.

### DIFF
--- a/openprescribing/api/urls.py
+++ b/openprescribing/api/urls.py
@@ -13,7 +13,7 @@ urlpatterns = [
     path(r"bubble/", views_spending.bubble, name="bubble"),
     path(r"tariff/", views_spending.tariff, name="tariff_api"),
     path(
-        r"spending_by_ccg/",
+        r"spending_by_sicbl/",
         views_spending.spending_by_org,
         name="spending_by_ccg",
         kwargs={"org_type": "ccg"},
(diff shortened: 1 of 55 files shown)


evansd (approved):
Looks correct to me, thanks for doing this.

I think there are still a couple of instances of "Clinical Commissioning Group" in the text which might need amending:
```
$ rg -i 'clinical commissioning group'
openprescribing/templates/index.html
25:        <p>We've identified standard <a href="https://openprescribing.net/measure/">prescribing measures</a>, and created dashboards for every Clinical Commissioning Group in the country.</p>

openprescribing/templates/all_ccgs.html
11:<p>Clinical commissioning groups (SICBLs) are NHS organisations that organise the delivery of NHS services in England. They are clinically led groups that include all of the GP groups in their geographical area.</p>
```

But nothing equivalent that I can see for "Transformation Partnership".

PR #3698, opened 2022-09-30
Import all GTINs

### LINKED ISSUE #3689: dm+d browser - missing GTIN
We've had an email from a user:

> GTIN 7613421064396 is missing
>
>Revolade 25mg tablets (Novartis Pharmaceuticals UK Ltd) 28 tablet 4 x 7 tablets
>17075611000001107

Having taken a look at both the [dm+d browser](https://openprescribing.net/dmd/ampp/17075611000001107/) and  the `gtin` table on BQ, I can only see one GTIN listed for this AMPP:
image

However, on the TRUD release for this week, there appears to be 4, with 3 having end dates:

image

How do we manage changes in GTIN?

inglesp:
Previously, we assumed that an AMPP could only have a single GTIN, and so ended up importing only the first GTIN listed in the data file.  We now import all of them.

Fixes #3689


openprescribing/dmd/management/commands/import_dmd.py
```diff
@@ -214,6 +206,83 @@ def load_elements(self, filename_fragment):
...
+
+                <xsl:template match="GTINDATA">
+                    <GTIN>
+                        <GTIN><xsl:value-of select="GTIN" /></GTIN>
+                        <APPID><xsl:value-of select="../AMPPID" /></APPID>
+                        <STARTDT><xsl:value-of select="STARTDT" /></STARTDT>
+                        <ENDDT><xsl:value-of select="ENDDT" /></ENDDT>
+                    </GTIN>
+                </xsl:template>
+            </xsl:stylesheet>
```
evansd, line 279: I don't really speak XSLT, but this looks plausible.

evansd (approved):

openprescribing/dmd/management/commands/import_dmd.py
inglesp: Nor do I!  (But I feel I should -- using it to rewrite the dm+d XML into the format that `loaddata` can read would simplify much of this import script.  #newcoder.)

PR #3741, opened 2022-11-01
Support monthly alert subscriptions for ICBs

### LINKED ISSUE #3725: Add ability to subscribe to ICB level monthly email alerts and price concession alerts
Discussed with @richiecroker earlier. With the establishment of ICBs, user requirements for email alerts is likely to change.

It currently does not appear to be possible to subscribe to monthly email alerts or price concession alerts at the ICB level (currently can be done at SICBL, PCN level and practice level).

Where an ICB is made up a single SICBL, the SICBL data will be the same as the ICB data. However a number of ICBs are made up of multiple SICBLs (e.g. NHS North East and North Cumbria Integrated Care Board which is made up of 8 SICBLs) - it is likely some users will want to receive an ICB overview (perhaps in addition to SICBL alerts they may already be subscribed to).

I think this should just be an additional option users can add. No need to change existing SICBL subscriptions to ICB subscriptions etc.

evansd:
Note this _doesn't_ support price concession alerts because we've only ever had those for the "original" set of organisations: practices, CCGs/SICBLs and All England. If we decide we need these for other organisation types there'll need to be a separate piece of work to try to factor the price concessions alerts out into something that can be used more widely. 

Closes #3725


openprescribing/frontend/views/bookmark_utils.py
```diff
@@ -260,6 +260,8 @@ def __init__(self, org, interesting_saving=1000, interesting_change_window=12):
             self.measure_filter_for_org = {"pcn": org, "practice": None}
         elif isinstance(org, PCT):
             self.measure_filter_for_org = {"pct": org, "practice": None}
+        elif isinstance(org, STP):
+            self.measure_filter_for_org = {"stp": org, "practice": None, "pct": None}
```
inglesp, line 264: Should we be adding `"stp": None` to the branches above?

inglesp (approved):
All looks reasonable, thanks.  There's no test for creating an ICB bookmark, which is fine as we don't have tests for creating other bookmarks, but it'd be good to know that you'd tried this out manually.

openprescribing/frontend/views/bookmark_utils.py
evansd: No, that would break it because it would exclude all MeasureValues for any organisation at STP level or less in the hierarchy. It's only organisation types _below_ the target organisation for which you need to add `None` filters.

I didn't remember any of this stuff but fortunately, as a side effect of doing some Pandas reshaping, the tests blow up with an error if you don't get your organisation filters right.

evansd:
Yes, I did test creating a bookmark through the browser locally.

openprescribing/frontend/views/bookmark_utils.py
inglesp: Oh, yuck.

PR #3756, opened 2022-11-10
change case of "needles"

richiecroker:


brianmackenna:
It looks like a small change @richiecroker is this a good one for me to pair with another person from Clinical Informatics team take it?

I know it slows us down now but will speed us up in the long run!

PR #3757, opened 2022-11-10
Create dm+d version

richiecroker:


richiecroker:
[Preview](https://openprescribing.net/measure/preview_lpneedles/) shows slightly more costs (approx £7200 out of £693k) and no negative values compared with non-dm+d version,so we should move to this version.

chrisjwood16 (approved):

PR #3923, opened 2023-03-07
Remove feedback form and replace with mailto

evansd:
This was actually making life more difficult for the people picking up these emails.

See discussion here:
https://bennettoxford.slack.com/archives/C31D62X5X/p1678118351759649


openprescribing/frontend/views/views.py
```diff
@@ -21,13 +21,10 @@
 from django.shortcuts import get_object_or_404, redirect, render
 from django.template.loader import render_to_string
 from django.urls import get_resolver, reverse
-from django.utils.http import is_safe_url
 from django.utils.safestring import mark_safe
 from dmd.models import VMP
-from frontend.feedback import send_feedback_mail
 from frontend.forms import (
     BookmarkListForm,
-    FeedbackForm,
```
StevenMaude, line 30: question: Can `FeedbackForm` be removed too?

https://github.com/ebmdatalab/openprescribing/blob/0344e66a785984ad4ca211a6b9600debe40a0aa9/openprescribing/frontend/forms.py#L102-L124

It only looks to be used here, and those uses are removed in this PR.

openprescribing/frontend/context_processors.py
```diff
@@ -1,15 +1,17 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 def support_email(request):
+    subject = f"OpenPrescribing Feedback ({request.get_full_path()}):"
     return {
         "SUPPORT_TO_EMAIL": settings.SUPPORT_TO_EMAIL,
         "SUPPORT_FROM_EMAIL": settings.SUPPORT_FROM_EMAIL,
+        "FEEDBACK_MAILTO": f"{settings.SUPPORT_TO_EMAIL}?subject={quote_plus(subject)}",
```
StevenMaude, line 14: thought: It's not really directly related to this PR, but if the support email that people send emails to is now a PHC one, and not `openprescribing.net`, I note that there are several mentions of contacting `feedback@openprescribing.net` in various JSON files.

Maybe these need updating, if they are in use?

https://github.com/search?q=repo%3Aebmdatalab%2Fopenprescribing+feedback%40openprescribing.net+language%3AJSON&type=code&l=JSON

GitHub search results for the feedback email address in the openprescribing repository.

StevenMaude (approved):
This looks fine, as much as I can tell from not knowing the codebase well.

openprescribing/frontend/views/views.py
evansd: Good catch, thanks. I thought I had removed this but must have missed it.

openprescribing/media/js/src/global.js
```diff
@@ -17,11 +17,6 @@ if (!window.console) {
   };
 }
 domready(function() {
-  $('.feedback-show').click(function(e) {
-    e.preventDefault();
-    window.location.href='/feedback/?from_url=' + encodeURIComponent(window.location.href);
-  });
-
```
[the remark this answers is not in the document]
evansd: I assumed there was some styling hanging off it, but if not then yes we should get rid of it. I'll double check.

openprescribing/frontend/context_processors.py
evansd: Also a good catch. I think those should be updated too.

PR #3939, opened 2023-03-15
Add tapentadol caps

richiecroker:


richiecroker:
https://openprescribing.net/measure/preview_opioidspercent/
https://openprescribing.net/measure/preview_opioidper1000/

These are the two preview measures - they seem as expected - slightly higher due to addition of tapentadol caps, but now easier to update.   Also possibly a useful template for future measure rewrites.

richiecroker:
@a-d-brown yes please - add your checked by details in the metadata, then we can publish

a-d-brown (approved):

PR #3976, opened 2023-03-31
Major refactor of NCSO importer

evansd:
This switches the importer to use the RSS feed of the PSNC's "System Supplier" email alerts, rather than scraping their website. This was supposed to have two advantages:

 * more timely updates as there's sometimes a delay between publishing the new concessions and updating the PSNC website;

 * simpler, more reliable VMPP matching as the emails contain a table with SNOMED VMPP IDs.

However, as became clear while writing the new parser, these emails are clearly hand-crafted not machine generated and there's usually at least one incorrect VMPP ID in every email.

So now we still have to do all the convoluted string matching we had to do before and we do some _extra_ work to check that the VMPP we end up with matches the one supplied in the email – and warn if it doesn't.

We are, therefore, in a very real sense worse off than we were before. Nevertheless, I commend this PR to the House for the following reasons:

 1. The sunk cost fallacy.

 2. The new code correctly excludes [invalid][1] VMPPs and handles [ambiguous][2] naming issues.

 3. By detecting incorrect VMPPs in the emails we have the opportunity to improve the system for everyone.

 4. The code is arguably slightly better factored than what we had before as it separates the fetching, parsing, matching and inserting phases more cleanly.

 5. We get updates a bit sooner.

The diffs probably aren't that helpful here so it might be easier to review the complete files:
 * [openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py](https://github.com/ebmdatalab/openprescribing/blob/evansd/price-concessions-rss/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py)
 * [openprescribing/pipeline/tests/test_fetch_and_import_ncso_concessions.py](https://github.com/ebmdatalab/openprescribing/blob/evansd/price-concessions-rss/openprescribing/pipeline/tests/test_fetch_and_import_ncso_concessions.py)

[1]: https://github.com/ebmdatalab/openprescribing/issues/3978
[2]: https://github.com/ebmdatalab/openprescribing/issues/3979


openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
```diff
@@ -223,4 +266,122 @@ def regularise_ncso_name(name):
     # Remove spaces around slashes
     name = re.sub(" */ *", "/", name)
 
+    # dm+d names often have the units appended after the pack size (e.g. "12 tablets"),
+    # we want to remove the units leaving just the "12"
+    name = re.sub(r"(\d+) [^\d]+$", r"\1", name)
```
rebkwok, line 271: Will this remove some units that we want to keep?  e.g. I can see that it would make sense to remove e.g. "ampoule" at the end of [Adrenaline (base) 1mg/10ml (1 in 10,000) dilute solution for injection ampoules 10 ampoule](https://openprescribing.net/dmd/vmpp/4523811000001103/)

But for e.g. [Adrenaline (base) 1% eye drops 7.5 ml](https://openprescribing.net/dmd/vmpp/3705611000001107/) - it would remove the "ml" - is that what it should do?

(I guess this is only being used to help name-matching, so the question is probably irrelevant)

openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
```diff
@@ -223,4 +266,122 @@ def regularise_ncso_name(name):
...
+                    date=item["date"],
+                    drug=item["drug"],
+                    pack_size=item["pack_size"],
+                    vmpp_id=None,
+                    defaults=dict(
+                        price_pence=item["price_pence"],
+                    ),
+                )
+
+            if created:
```
rebkwok, line 317: Is it helpful to log whether this was a matched or unmatched new concession?

rebkwok (approved):

openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
```diff
@@ -16,171 +17,156 @@
...
+# Match strings like "March 2020 Price Concessions"
+MONTH_DATE_RE = re.compile(
+    r"""
+    (?P<month>
+        january | february | march | april | may | june | july | august |
+        september | october | november | december
+    )
+    \s+
+    (?P<year> 20\d\d)
+    \s + price \s+ concessions
```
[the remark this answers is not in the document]
evansd: Good spot, thanks!

openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
evansd: I think it's OK as is. We report the unmatched concession elsewhere.

openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
evansd: This is a good question and it's prompted me to expand the comments a bit more here. The basic issue is that the names we get from the feed don't have the units, so we have to remove them to in order to match.

The old code effectively did the same thing, but using a substring match:
https://github.com/ebmdatalab/openprescribing/blob/f0ca7a5e41a5aefdbfba125741041c761f35b02f/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py#L178-L180

There are some rare cases where this results in ambiguous names, which I've documented here:
https://github.com/ebmdatalab/openprescribing/issues/3979

But where there's more than one matching name we now treat that as no match and require a manual resolution:
https://github.com/ebmdatalab/openprescribing/blob/531e0e0a94126f3ae808eceab7b5de4539987bc6/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py#L205-L207

PR #4000, opened 2023-04-04
new_measure__amox_500_15_caps.json

richiecroker:


richiecroker:
Preview measure here: https://openprescribing.net/measure/preview_amox_500_15_caps/

brianmackenna:
Looks good - are there any obvious conditions we can think of that have >5 days course recommended. It would be good to hint that it might not be realistic to get this to 0% - I think it reads like that now.

richiecroker:
@chrisjwood16 I don't think the "title" field actually does anything...

chrisjwood16:
> @chrisjwood16 I don't think the "title" field actually does anything...

Ah, couldn't work out what was missing, but felt weird to leave it as "title"!

chrisjwood16:
> Looks good - are there any obvious conditions we can think of that have >5 days course recommended. It would be good to hint that it might not be realistic to get this to 0% - I think it reads like that now.

I've added in H. pylori treatment as an example of where higher doses/longer courses might be used.

chrisjwood16 (approved):

PR #4001, opened 2023-04-04
updated_measure__uti_antibiotics_3_day.json

richiecroker:
…json

amend from average quantities per item to % > 3 days, and renaming as KTT no longer current.


richiecroker:
Updated measure here: https://openprescribing.net/measure/preview_uti_antibiotics_3_day/

chrisjwood16:
I've removed some text that was duplicated that was causing tests to fail.
I've also added in fosfomycin sachets as these are given as a 2nd line choice similar to pivmecillinam.

chrisjwood16 (approved):
Approved. As long as you are happy with my changes @richiecroker .

PR #4010, opened 2023-04-04
updated_measure__lpherbal.json

richiecroker:
Remove the need for a view.
If approved we'll need to remove
- `measures.vw__herbal_list` 
- `measure_sql/vw__herbal_list.sql`


chrisjwood16 (approved):
This looks good. I've also added in review fields.

PR #4068, opened 2023-05-16
Insert manual corrections to NCSO data

evansd:
See:
https://psnc.org.uk/our-news/price-concession-update-for-april-2023-chlorphenamine-2mg-5ml-oral-solution/

We can't pick these up as part of the normal import. We could just insert them directly into the database but then they'd disappear if we ever tried to do a clean re-import of the concessions data so having them embedded into the code like this seems preferable.


evansd:
Yes, quite possibly. Although the correction issued on the website doesn't quite fit the usual structure either so I don't know if we'd have painlessly picked it up in any case.

PR #4071, opened 2023-05-19
Rework outliers pages

inglesp:
The new pages are not yet linked to from the rest of the site, so this can be deployed before we get feedback from the rest of the team.

The code is in two parts:

* There is a build script that produces and serializes a DataFrame for each organisation, with that organisation's prescribing for each BNF chemical and data about how that organisation's prescribing relates to all other organisations.  The build script is not yet hooked up to the rest of the import pipeline, but it could be.
* There is a new per-organisation view that reads the DataFrame and renders tables showing the chemicals for which the organisation is most outlierish.


openprescribing/outliers/build.py
```diff
@@ -0,0 +1,130 @@
...
+     - mean
+        - the mean of the ratios for all organisations of given type
+     - std
+        - the standard deviation of the ratios for all organisations of given type
+     - zscore
+        - this organisation's z-score for this chemical, against all organisations of
+          given type
+    """
+
+    start_date, *_, end_date = generate_dates(end_date, months)
```
evansd, line 38: Ha! Lovely bit of tuple unpacking here.

openprescribing/outliers/views.py
```diff
@@ -0,0 +1,48 @@
...
+def outliers_for_one_entity(request, entity_type, entity_code):
+    entity = _get_entity(entity_type, entity_code)
+    base_analyse_url = f"/analyse/#org={entity_type}&orgIds={entity_code}"
+    bnf_code_to_name = dict(
+        list(Chemical.objects.values_list("bnf_code", "chem_name"))
+        + list(Section.objects.values_list("bnf_id", "name"))
+    )
+    df = pd.read_feather(
+        settings.OUTLIERS_DATA_DIR / entity_type / f"{entity_code}.feather"
+    )
```
evansd, line 17: Given that we're expecting these files to sometimes not exist it would be nice to handle that gracefully here, mainly to avoid the Sentry spam we'll get otherwise. 

By "gracefully" I just mean a 404 – possibly with a message along the lines of "we're reimporting data please check back in a few minutes", if it's easy to include.

openprescribing/outliers/tests.py
```diff
@@ -0,0 +1,35 @@
...
+        for i in range(6):
+            ccg.practice_set.create(code=f"ABC00{i}", setting=4)
+
+        cls._remove_patch = patch_global_matrixstore(
+            matrixstore_from_data_factory(factory)
+        )
+
+    def test_build_smoke_test(self):
+        # The data produced by the DataFactory is not rich enough to test anything
+        # interesting, but this is a reasonable smoke test that the calculations work.
```
evansd, line 28: Great, yes, this was all I thought it was worth doing. Glad to see that there was enough matrixstore factory stuff lying around (which I'd completely forgotten about) to make this relatively easy.

openprescribing/outliers/views.py
```diff
@@ -0,0 +1,48 @@
...
+        _outlier_context(row, base_analyse_url, bnf_code_to_name)
+        for row in df.sort_values("zscore")[:10].itertuples()
+    ]
+    context = {
+        "entity": entity,
+        "entity_type": _entity_type_human(entity_type),
+        "high_outliers": high_outliers,
+        "low_outliers": low_outliers,
+    }
+    return render(request, "outliers_for_one_entity.html", context)
```
evansd, line 33: This looks like it might be a relatively expensive view, so possibly when it comes to making the feature public it would be worth sticking a caching header on this for Cloudflare. I can't remember exactly what we want off the top of my head, but I'm sure I could dig it out when the time comes.

evansd (approved):
Lovely.

There's amazingly not that much code for what is potentially a pretty sophisticated and powerful feature.

PR #4155, opened 2023-07-03
Yet more fixes for NCSO importer

evansd:
We now handle leading asterisks on drug names, and we also correctly report when we have failed to match a drug name.


inglesp (approved):
:(

PR #4235, opened 2023-08-15
Update and rename lpzomnibus.json to lpzomnibus_exclusions.json

richiecroker:


richiecroker:
this version currently includes trimipramine - seeking clarification from the medicines policy unit

PR #4239, opened 2023-08-15
Amend measure  lpzomnibus

richiecroker:
Removed tadalafil from SQL, updated hyperlink, and changed denominator calculation.


speed-vm (changes requested):
@richiecroker should it say items for the hyperlink rather than areas to be consistent with Guidance doc?

richiecroker:
@speed-vm items might get a little bit lost, as it can mean two things.  How about "therapeutic areas"

speed-vm:
That maybe sounds more like asthma, or hypertension. What about medications or something to qualify items - Prescription items.

Items is in the title of the nearly all of the low priority measures though so might need to think whether want to change all?

richiecroker:
Items is the name of the NHSE document, so we wouldn't want to change those!  Let's just go with items.

speed-vm (approved):
Yes, looks good

richiecroker (commented):
all ok

richiecroker (commented):
all ok

PR #4240, opened 2023-08-15
Update lpamiodarone.json

richiecroker:


chrisjwood16:
It mentions "is in line with NICE Guidance CG180" (taken from NHS E website) - but on checking this was replaced in 2021 by [NG196](https://www.nice.org.uk/guidance/ng196). Presume we should reference/link to the appropriate guidelines.

richiecroker:
> 

@chrisjwood16 my view is that we are quoting "NHS England states"...  there are a lot of issues with their wording, and if we try and correct we'll tie ourselves in knots, as these things aren't mentioned in later NICE guidelines.  I've got a list that I will email them with, but let's use their wording for now

chrisjwood16:
> > 
> 
> @chrisjwood16 my view is that we are quoting "NHS England states"... there are a lot of issues with their wording, and if we try and correct we'll tie ourselves in knots, as these things aren't mentioned in later NICE guidelines. I've got a list that I will email them with, but let's use their wording for now

Ah just pushed an update. They do link to the correct guidance further down, so think it's justified here to put correct link?

richiecroker:
we're going to come across this time after time - they are often referring to (@brianmackenna 's?) original work here, so as we're quoting I suggest we stick with what they're saying

chrisjwood16 (approved):

PR #4247, opened 2023-08-15
Update lpfentanylir.json

richiecroker:
Changed BNF codes query, updated WIM and added admin fields


chrisjwood16:
The first few sentences don't match the NHS E guidance - "Fentanyl, in the form of tablets, lozenges, films and nasal spray, is an opioid analgesic used for breakthrough cancer pain. There is little evidence it is more effective than oral morphine (such as oramorph) but it is much more expensive. This indicator does not include fentanyl patches, which are not immediate release. " are they from previous version?

chrisjwood16 (approved):

PR #4254, opened 2023-08-21
Update lpherbal.json

richiecroker:


richiecroker:
With this marker i've had to edit out some of the NHSE text, as there appears to words missing.  I will email them and reinstate when they've corrected.

richiecroker:
wording reinstated

PR #4271, opened 2023-08-29
Update lpoxycodone.json

richiecroker:
update text


chrisjwood16:
I think this should also have under recommendations "Prescribe only if the decision to prescribe has been made after a multidisciplinary team discussion." - it seems to be in the wrong box on the website but intention seems to be there.

richiecroker:
I would suggest leave as it and write to them to tell them, we can update
if they fix it!

chrisjwood16 (approved):

PR #4273, opened 2023-08-29
Update lpsilkgarments.json

richiecroker:
amend text


a-d-brown (approved):
There was a bullet point in our text saying:
- Prescribe only if no other item or intervention is clinically appropriate.

I couldn't see that in the NHSE guidance so I've removed it.

Assuming that's straight forward, then all looks good and approved.

PR #4275, opened 2023-08-29
Update lptravelvacs.json

richiecroker:
update text and SQL

### DIFF
--- a/openprescribing/measure_definitions/lptravelvacs.json
+++ b/openprescribing/measure_definitions/lptravelvacs.json
@@ -1,5 +1,5 @@
 {
-  "name": "NHS England Low Priority Treatment - travel vaccines",
+  "name": "Items which should not routinely be prescribed in primary care - travel vaccines",
   "title": "Travel vaccines cost per 1000 patients",
   "description": "Cost of travel vaccines per 1000 patients",
   "numerator_short": "Travel vaccines cost (£)",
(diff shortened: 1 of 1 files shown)


a-d-brown:
@richiecroker 

**1.**
 There was a bullet point for:
- Deprescribe in patients currently prescribed this medicine.
...but in their guidance NHSE just say "Prescribe only for an indication named in this guidance."

For that to make sense in our wording, I've suggested changing it to:
- Prescribe only for an indication not exclusively related to travel, as outlined in Immunisation Against Infectious Disease – the Green Book.

**2.**
Japanese encephalitis vaccines appear to be missing from the analyse page link, and the underlying SQL. Specifically these code trees:
- Japanese encephalitis vaccine (Imported)(1404000AS)
- Japanese encephalitis vaccine (Inactivated,adsorbed)(1404000AR)
Are they omitted intentionally?

There is a code being searched on in the current SQL but I'm not sure it is correct?
- 1404000AB # Japanese Encephalitis Vaccine

richiecroker:
I've been inclined through this to stick to NHSE wording, as that's what
we're quoting.

re codes - good spot, the code must have changed - i'll change them now

a-d-brown (approved):
Happy to use the exact wording from NHSE, I've updated further to reflect that, i.e.
"Prescribe only for an indication named in this guidance."

PR #4336, opened 2023-09-21
Raise an explicit error for TRUD login failure

evansd:
This should make it easier to identify the issue in future.

### DIFF
--- a/openprescribing/dmd/management/commands/fetch_dmd.py
+++ b/openprescribing/dmd/management/commands/fetch_dmd.py
@@ -32,6 +32,10 @@ class Command(BaseCommand):
             "commit": "LOG+IN",
         }
         rsp = session.post(login_url, params)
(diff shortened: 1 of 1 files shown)


StevenMaude (approved):

openprescribing/dmd/management/commands/fetch_dmd.py
```diff
@@ -32,6 +32,10 @@ def handle(self, *args, **kwargs):
             "commit": "LOG+IN",
         }
         rsp = session.post(login_url, params)
+        if "The email address or password is invalid" in rsp.text:
+            raise RuntimeError(
+                f"TRUD login failure for username: {settings.TRUD_USERNAME!r}"
+            )
```
rebkwok, line 38: This could also post to the data team slack channel with `notify_slack(message, is_error=True)`

PR #4343, opened 2023-09-28
Fix chart image rendering in PhantomJS

evansd:
Google's JS is now using ES6 flavoured code which PhantomJS's ancient JS engine can't handle. This means that attempting to render chart images dies with:

    ReferenceError: Can't find variable: Map

      https://www.googletagmanager.com/gtag/js?id=G-9DPND0VYSC&cx=c&_slc=1:141

This was the quickest way I could think of the make the problem go away.

### DIFF
--- a/openprescribing/frontend/context_processors.py
+++ b/openprescribing/frontend/context_processors.py
@@ -30,7 +30,13 @@ def debug(request):
 
 def google_tracking_id(request):
     tracking_id = None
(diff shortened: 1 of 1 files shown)


inglesp (approved):
Oh ffs

evansd:
Yeah, tell me about it!

I guess at some point we'll replace with it with Playwright, but not about to do that now.

PR #4409, opened 2023-11-01
Stop filtering practices with empty list sizes

evansd:
When calculating monthly/annual totals, the Analyse page includes prescribing from practices which don't have any recorded list size. However it doesn't display such practices on the chart, which leads to confusion.

The reason it doesn't display such practices is that the MatrixStore powered API goes to some lengths to mimic exactly what the Postgres powered API did. However I suspect this behaviour was just a consequence of the way the SQL happened to work rather than a deliberate decision. And, given that it's confusing, we should stop doing it.

Related Slack thread:
https://bennettoxford.slack.com/archives/C010SJ89SA3/p1698846614341989?thread_ts=1698241192.796879&cid=C010SJ89SA3

### DIFF
--- a/openprescribing/api/views_org_details.py
+++ b/openprescribing/api/views_org_details.py
@@ -97,8 +97,6 @@ def _get_practice_stats_entries(keys, org_type, orgs):
                 value = matrix[index]
                 if value != 0:
                     has_value = True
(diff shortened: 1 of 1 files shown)


inglesp (approved):
Nice, thanks

PR #4448, opened 2023-11-22
Update lpneedles.json

richiecroker:
remove references to dm+d, use bnf name for filtering


chrisjwood16:
Just to make things more complicated the dm+d uses "insulin needles" but the EPD data sometimes abbreviates this to "ins needles" and these seem to be missed by current code. For example not picked up here:

image

Could change the code to `LIKE '%ins% needles'`

Which would also unexpectedly pick this up....

image

richiecroker:
I'm going to close this PR without deploying, but the insupen example above is the _old_ truncated `bnf_name` which was phased out a while ago:

richiecroker:
This fix doesn't work, so am closing PR.

PR #4489, opened 2023-12-12
retire measure desogestrel

a-d-brown:
Deleted measure due to price changes making the measure less relevant currently.

PR #4493, opened 2023-12-12
Don't reference real measure in test

inglesp:
The deogestrel measure is going away.  All this test has to do is make check that some measure objects are created.


a-d-brown (approved):

evansd:
Did I misunderstand this section of the test here (as is quite possible)?
https://github.com/ebmdatalab/openprescribing/blob/09a18cf587dc04b4ec6f40d9a693b1b79cb13157/openprescribing/frontend/tests/commands/test_import_measures.py#L65-L85

That looks like it's referencing a real measure as well.

evansd:
Great, sorry for the noise!

richiecroker:
#4489 is still failing though: https://github.com/ebmdatalab/openprescribing/actions/runs/7184392384/job/19567016519

inglesp:
#4489 was failing because it didn't have this fix.  It does now, and is now green.

Sorry, should've thought of this!

PR #4505, opened 2023-12-13
Replace PyArrow serialization with `pickle`

evansd:
PyArrow serialization has long been deprecated and was removed in version 12 meaning that we have been unable to upgrade:
https://arrow.apache.org/blog/2023/05/02/12.0.0-release/#python-notes

But in fact, we don't need PyArrow at all now because the venerable `pickle` library has learned to do zero-copy serialization and deserialization. It isn't quite as simple as switching to `pickle.dumps()` because Pickle's zero-copy mechanism is lower-level than PyArrow's was, so we have to design our own tiny framing format for it. But this doesn't involve too much code.

The new code is about 5x as fast as PyArrow for both serializing and deserializing. I don't necessarily expect that to translate into a noticeable overall performance improvement, but the key thing is that it isn't worse that it was before.

For now we retain the capacity to read data serialized in the old format to facilitate easy deployment. And later PR will remove this, allowing us to drop the PyArrow dependency and associated code entirely.

### DIFF
--- a/openprescribing/matrixstore/serializer.py
+++ b/openprescribing/matrixstore/serializer.py
@@ -1,3 +1,4 @@
+import pickle
 import struct
 import warnings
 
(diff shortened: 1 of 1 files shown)


inglesp (approved):
Very clever!  (I think that's a complement...)

PR #4551, opened 2024-01-08
move measure definitions and views, and create csv tables

richiecroker:
Move all measure definitions to measures\definitions

Move measure related views to
measures\view

Create table for opioids at
measures\table


openprescribing/frontend/management/commands/preview_measure.py
```diff
@@ -9,11 +9,13 @@
 from .import_measures import BadRequest
 from .import_measures import Command as ImportMeasuresCommand
 from .import_measures import ImportLog, relativedelta
+from .import_measures import upload_supplementary_tables
 
 
 class Command(BaseCommand):
     def handle(self, github_url, **options):
         try:
+            upload_supplementary_tables()
```
evansd, line 18: I wonder if there's going to be a confusion here because if the measure PR involves changing or adding supplementary tables these won't be included as part of the preview process.

I don't know if there's much we can do about this other than warn people though.

inglesp: Don't understand this.

If eg Rich changes a supplementary table, and then runs `preview_measure` via Slack, then all the supplementary tables are going to be uploaded.  (Aren't they?)

evansd: Have you seen how `preview_measure` works? It pulls the measure definition JSON over HTTP:
https://github.com/ebmdatalab/openprescribing/blob/9c9144387dde552dd3453dffcce538fc511bf324/openprescribing/frontend/management/commands/preview_measure.py#L38-L46

inglesp: Oh of course.  Bother.

Can't think of a solution that doesn't involve taking a chainsaw to the whole pipeline.

PR #4552, opened 2024-01-08
Update vw__opioids_total_dmd.sql

richiecroker:
Add CASE statement to subquery to return base substance ing code where one exists.


richiecroker:
superseded by newer branch.

PR #4604, opened 2024-01-29
Update measure ktt9 cephalosporins star.json

chrisjwood16:
Rename measure to "broad_spectrum_star" as includes cephalosporins, co-amox and quinolones (not just cephalosporins).
Removed references to outdated KTT9
Update bnf_codes_query to select all oral antibiotics of the chosen classes.


chrisjwood16:
Consider if better to use list size as denominator rather than STAR-PUs

a-d-brown (approved):

richiecroker (commented):
some minor points:

1. Can we consider wording of "why it matters"?  This is a very old measure, and possibly not in line with house style now?  Also, could still do with a hyperlinked reference
2.  Needs a "y-axis" marker (new since this measure was changed)
3.  Needs a `measure_type` flag

PR #4624, opened 2024-02-06
Update injectable_antibiotics.json

a-d-brown:
This measure is less suitable for inclusion in the Improvement Radar.

I have added the variable:

- "radar_exclude": true


richiecroker (changes requested):
it's not a BNF code only measure, it's also using dm+d.  Perhaps we need a bnf_code+dmd tag
do you want to review it briefly while we're committing, so we can update the review date (2 year on a low complexity measure)

a-d-brown:
Reviewed measure content, review dates updated.

richiecroker (approved):

PR #4631, opened 2024-02-09
Surface failures in sending emails

evansd:
Previously, errors were logged only to the `django.log` file and not surfaced in the console to the user running the command. Additionally, `BadAlertImageError` exceptions weren't treated as errors so the command as a whole would appear to have run successfully even where these were encountered.

As these seem to occur quite regularly for transient reasons, this meant that every month a random selection of subscribers would never receive their emails.


inglesp (approved):
Thanks for digging into this.

PR #4647, opened 2024-02-14
Add improvement radar links

inglesp:


openprescribing/frontend/tests/functional/test_measures.py
```diff
@@ -319,7 +319,7 @@ def test_measures_for_one_practice(self):
         self._verify_link(
             panel_element,
             ".inner li:nth-child(3)",
-            "Compare all Sub-ICB Locations in England on this measure",
+            "Compare all SICBLs in England on this measure",
```
evansd, line 322: How did these pass before?

evansd (approved):
Surgical!

Although it looks like the tests are complaining about trying to stick the string `"false"` in a boolean field. I can't quite see why this should be happening though.

openprescribing/frontend/tests/functional/test_measures.py
inglesp: They didn't!

inglesp: Oh!  These tests are failing in CI.  Bamboozling.

evansd: I wonder if you haven't rebuilt your JS locally?

inglesp: 5am thought: out-of-date JS?

inglesp: jinx

PR #4680, opened 2024-03-01
Add custom styles for labs pages

tomodwyer:
Whilst this can be tested by running OpenPrescribing locally, the fastest way is to use DevTools.

## Steps to test

1. Go to https://openprescribing.net/labs/sicbl-improvement-radar/ in Chrome (or Chrome based browser)
2. Open Chrome DevTools
3. Go to the "**Sources**" tab
4. Right-mouse click on the "**custom.css**" file and select "**Override content**"
5. A toolbar will pop up above the DevTools, click on "**Select folder**" and select somewhere on your drive you can access
6. You may now need to "**Allow**" Chrome access to this location (a toolbar will appear at the top of the window)
7. Go to the folder in your code editor of choice, and replace the contents of the "**custom.css**" file with the file in this PR
8. The page should update in your browser when you save the file in your editor

### Before

Page header before styling changes

---

Example page content before styling changes

### After

Page header after styling changes

---

Example page content after styling changes

### DIFF
--- a/openprescribing/templates/labs/custom.css
+++ b/openprescribing/templates/labs/custom.css
@@ -1,3 +1,15 @@
+:root {
+  --oxford-50: #f1f7ff;
+  --oxford-600: #0058be;
+  --oxford-900: #001936;
+
+  --gray-200: #e5e7eb;
+  --gray-300: #d1d5db;
+
+  --sun-300: #ffedb0;
+  --sun-500: #ffd13a;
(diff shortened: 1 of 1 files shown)


StevenMaude (approved):
I tested this using the steps in the PR comment and I was able to review the changes on `openprescribing.net`, overriding with local CSS changes.

PR #4757, opened 2024-04-22
Amend measure  icsdose

chrisjwood16:
Update measure definition.
Update links.

### DIFF
--- a/openprescribing/measures/definitions/icsdose.json
+++ b/openprescribing/measures/definitions/icsdose.json
@@ -1,24 +1,18 @@
 {
   "name": "High dose inhaled corticosteroids",
-  "title": [
-    "high dose inhaled corticosteroids",
-    "compared with prescribing of all inhaled corticosteroids"
-  ],
-  "description": [
-    "Prescribing of high dose inhaled corticosteroids",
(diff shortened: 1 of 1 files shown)


chrisjwood16:
Pending agreement regarding the following (include or exclude from denominator):
Budesonide inhaler dose cut off - DuoResp 320/4.5 and Symbicort 400/12 licensed for use at both none high dose (1 puff BD) and high dose (2 puffs BD doses)
Ciclesonide inhaler dose cut off - Ciclesonide 160 licensed for use at both none high dose (2 puff OD) and high dose (2 puffs BD doses)

a-d-brown (approved):

chrisjwood16:
Discussed at ClinInfo meeting 15/07/24. Agree with inclusive approach (include inhalers where reasonable likelihood a high dose could be being given e.g. Symbicort 400/12. 
- This is consistent with previous approach. 
- To exclude could reveal more about formulary choices than actual clinical practice - as an extreme example an ICB prescribing only Symbicort for LABA/LAMA would appear to have 0 high dose ICS use, whereas a an identical ICB prescribing equivalent doses of Fostair would appear much higher.

chrisjwood16:
@richiecroker ready to merge

PR #4758, opened 2024-04-23
Amend measure  semagglutidequantity

chrisjwood16:
Add tirzepatide multidose pen to numerator
Add tirzepatide single dose and multidose pen to denominator
Reword "Why it matters" to include tirzepatide
Rename measure to account for addition of tirzepatide


a-d-brown:
Suggested some very minor rewording to enhance clarity.

Since we're now referencing that '**single-use pens**' exist, I think there's potential to conflate that with the term '**single pens**' also used in the text. So I've suggested removing the term 'single pens' and have reworded to be more explicit.

a-d-brown (approved):

chrisjwood16:
Ready to merge but waiting for tirzepatide data to appear before doing so

chrisjwood16:
Reviewed with Feb 24 data - appears to be correct e.g. https://openprescribing.net/measure/preview_excess_pens/practice/M81040/
Ready to merge @richiecroker

PR #4769, opened 2024-04-29
Tell PhantomJS to ignore certificate errors

evansd:
PhantomJS is now sufficiently old that it can't validate whatever certificate Cloudflare is now using. Obviously this is unsustainable and we'll have to rewrite the image grabbing code eventually, but this keeps us going for now.

These are requests made from our own server back to our own server so I'm not too concerned by the MitM threat here.


inglesp (approved):
:sake:

PR #4790, opened 2024-05-16
Consolidate logic for `org_type` aliasing

evansd:
This allows us to make all our APIs accept the "public facing" org type names e.g `icb` instead of `stp`.


inglesp (approved):
baby steps!

PR #4833, opened 2024-06-06
Handle price strings with previous prices

evansd:
This fixes the NCSO importer to handle price strings like:

    £7.02 (previously £6.17)


inglesp (approved):
Fine -- but do we handle the price of a concession changing?

evansd:
Yes, we do. We use `update_or_create` to update the price on existing concession objects:
https://github.com/ebmdatalab/openprescribing/blob/edc38ccf2e522e45d437d3a93fe41206e5a6549e/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py#L284-L297

PR #4869, opened 2024-07-04
Update price concession calculation

### LINKED ISSUE #4860: Price Concessions calculations need updating
There have been a couple of changes to the price concessions calculations, which means that we are possibly underestimating the effect on NHS organisations:

1. From April 2023 price concessions are not subject to a discount deduction.  We have not taken this into account when calculating the price difference between the tariff price (which does have a deduction) and the concession price - but are still using the average National Average Discount Percentage (NADP) from when we implemented the tool.
2. From April 2024 there is a new discount model (with caveats, see $ below), which (for tariff price) depends on the category of the Drug Tariff the item is in.  See [here](https://www.drugtariff.nhsbsa.nhs.uk/#/00857004-DC/DC00856269/Part%20V%20-%20Deduction%20Scale%20(Pharmacy%20Contractors)) for details.

Consequently we need to make some changes to the price concessions calculation.  Having discussed with @evansd we agreed that the SQL would be the best place to make amendments.

I've created a draft SQL query below in BigQuery, will need some amendments, but does show that the concept works.

```SELECT
  ncso.date AS date,
  vmpp.bnf_code AS bnf_code,
  tariff.price_pence AS tariff_price,
  CASE
    WHEN ncso.date >='2024-04-01' THEN #for data after April 2024
      (CASE 
        WHEN tariff.tariff_category IN (1,11) THEN #tariff categories 1 (XIII Cat A) and 11 (XIII Cat M)
          COALESCE(ncso.price_pence - (tariff.price_pence *0.8), 0) #have 20% discount on tariff, not concessions
        WHEN tariff.tariff_category IN (5,6,7,8,10) THEN #Part IX products - very unlikely to be a price concession
          COALESCE(ncso.price_pence - (tariff.price_pence *0.9015), 0)  #have 9.85% discount on tariff, not concessions
          ELSE
          COALESCE(ncso.price_pence - (tariff.price_pence *0.95), 0)  #other drug tariff categories, 5% discount on tariff)
        END)
    WHEN ncso.date BETWEEN '2023-04-01' AND '2024-03-01' THEN #for data between April 2023 and March 2024
      COALESCE(ncso.price_pence - (tariff.price_pence *0.928), 0) #7.2% NADP for tariff, not concession
      ELSE
      COALESCE(ncso.price_pence - tariff.price_pence, 0)*.928 #Before April 2023, 7.2% on both tariff and concession
      END
      AS price_increase,
#CASE WHEN presentation.quantity_means_pack THEN
# 1
# ELSE
  vmpp.qtyval
# END
  AS quantity_per_pack
FROM
  dmd.ncsoconcession AS ncso
JOIN
  dmd.tariffprice AS tariff
  ON
  ncso.vmpp = tariff.vmpp AND ncso.date = tariff.date
JOIN
  dmd.vmpp AS vmpp
  ON
  vmpp.id=ncso.vmpp
JOIN
  hscic.presentation AS presentation
  ON
  presentation.bnf_code = vmpp.bnf_code
  #WHERE
  #ncso.date >= %(min_date)s AND ncso.date <= %(max_date)s
```

One further question - now that the NADP has been changed, shall we take the opportunity to add a table with the values from previous years, and use the actual value rather than the value (7.2%) we have used for all years?

$ Technically this is for pharmacy contractors only - dispensing doctors get a different discount rate, but this is probably too complicated to work out, and the differences are likely to be small for estimation purposes.

Also technically there has been a transition period for the new discount method since October 2022, but there has still been a [NADP value available](https://www.nhsbsa.nhs.uk/prescription-data/understanding-our-data/financial-forecasting).

evansd:
This involves two changes:

1. Applying a zero discount to concession prices from April 2023:
   https://cpe.org.uk/our-news/changes-to-discount-deduction-arrangements-from-april-2023/

2. Applying per-category discounts to drug tariff prices from April 2024:
   https://www.drugtariff.nhsbsa.nhs.uk/#/00857004-DC/DC00856269/Part%20V%20-%20Deduction%20Scale%20(Pharmacy%20Contractors)

Closes #4860

### DIFF
--- a/openprescribing/frontend/tests/data_factory.py
+++ b/openprescribing/frontend/tests/data_factory.py
@@ -126,9 +126,6 @@ class DataFactory(object):
     def create_tariff_and_ncso_costings_for_presentations(
         self, presentations, months=None
     ):
-        tariff_category, _ = dmd_models.DtPaymentCategory.objects.get_or_create(
-            cd=1, descr="TariffCategory VIIIA Category A"
-        )
         for presentation in presentations:
             for date in months:
                 price_pence = self.random.randint(10, 100)
@@ -138,7 +135,7 @@ class DataFactory(object):
                     tariff_price = TariffPrice.objects.create(
                         date=date,
                         vmpp=vmpp,
-                        tariff_category=tariff_category,
+                        tariff_category=self.create_tariff_category(),
                         price_pence=price_pence,
                     )
                     if self.random.choice([True, False]):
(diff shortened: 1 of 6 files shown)


inglesp (approved):
How tedious, thanks

PR #4900, opened 2024-07-22
Update opioidper1000.json

richiecroker:
add coalesce for tapentadol

### DIFF
--- a/openprescribing/measures/definitions/opioidper1000.json
+++ b/openprescribing/measures/definitions/opioidper1000.json
@@ -51,7 +51,7 @@
   ],
   "denominator_type": "list_size",
   "authored_by": "richard.croker@phc.ox.ac.uk",
-  "checked_by": "",
+  "checked_by": "christopher.wood@phc.ox.ac.uk",
   "date_reviewed": "2024-08-05",
   "next_review": "2025-08-05",
   "measure_complexity": "medium",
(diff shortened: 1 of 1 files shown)


chrisjwood16:
This looks good to me but a couple of thoughts - oxycodone - we use specifically "oxycodone hydrochloride" VPI. 
There is an ["Oxycodone" VPI](https://dmd-browser.nhsbsa.nhs.uk/ingredient/view/8147?ref=bmFtZT1veHljb2RvbmUmc2hvd0ludmFsaWRJdGVtcz1mYWxzZQ%3D%3D) and ["morphine" VPI](https://dmd-browser.nhsbsa.nhs.uk/ingredient/view/8131?ref=bmFtZT1tb3JwaGluZSZzaG93SW52YWxpZEl0ZW1zPWZhbHNl) created relatively recently - presumably there is a chance that like with tapentadol some products might move to these base substances in future?
For example for Oxycodone - go with COALESCE(bs_subid,ing) IN (55452001, 387024006)

For buprenorphine and below you use (COALESCE(vpi.bs_subid,vpi.ing)IN (387173000) instead of (COALESCE(vpi.bs_subid,vpi.ing) = (387173000). I don't think makes any difference in function but for consistency would use = where one value. (Possibly it's more efficient code?).

You've removed a few comments about the 24 hour preps. But I think these are helpful context to understand the code.

richiecroker:
@chrisjwood16 I think we need link to our OME table, so don't think we should add oxycodone and morphine VPIs as base substance at the mo.

`IN` vs `=` was a copy/paste thing.  Fixed now.  have also reinstated 24 hour comment - no idea where that went!

chrisjwood16 (approved):

PR #4933, opened 2024-08-06
Amend seven_day_prescribing.json

richiecroker:
Amend code and links.  Add metadata.


richiecroker:
The preview is now showing the same data as the current live version.  I have replaced looking for "caps" with dmd tags for caps.oral etc.  A useful lesson in how to write measure definitions - it didn't work first off, as there was a many-to-one relationship in the form route table.

a-d-brown (approved):

PR #5023, opened 2024-09-17
Amend title page

richiecroker:


chrisjwood16:
Looks good.
Only comment
> Every month, the NHS in England publishes anonymised data about the drugs prescribed by GPs.

It's not just drugs in the prescribing data. Could we just say "about prescriptions issued by GPs." (This is what EPD page says)

chrisjwood16 (approved):

PR #5062, opened 2024-10-25
Fix links to measure in email alerts

### LINKED ISSUE #4588: Direct linking to measures from email alerts is broken
A link like this (https://openprescribing.net/practice/J81010/measures/?tags=cost#lpherbal) (note the anchor fragment) is configured to scroll to the measure and highlight it, like this:

image

However, this doesn't usually work when linked from the email alerts. Here's an example link from an alert:

https://openprescribing.net/practice/J81010/measures/?utm_medium=email&utm_campaign=monthly+alert+2024-01-23&utm_source=dashboard-alerts&utm_content=%2Femail%2Fmonthly+alert+2024-01-23%2Fdashboard-alerts%2F1502#lpherbal

(in passing, I think we don't use google analytics any more? So those tracking tags are redundant?)

evansd:
We link to measures by linking to the appropriate organisation dashboard and including a fragment identifier which causes the JS to jump to and highlight the target measure. However this only works if the measure is actually on the page. By default the dashboard only displays measures tagged `core`. And so all links to non-core measures were effectively broken.

We don't include many non-core measures in the alerts which is why this issue wasn't more obvious. But we certainly do include the Low Priority Prescribing measures, and possibly others, so this needed fixing.

Closes #4588

### DIFF
--- a/openprescribing/frontend/templatetags/template_extras.py
+++ b/openprescribing/frontend/templatetags/template_extras.py
@@ -4,6 +4,7 @@ from django import template
 from django.contrib.humanize.templatetags.humanize import intcomma
 from django.contrib.staticfiles.storage import staticfiles_storage
 from django.utils import timezone
+from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 
 register = template.Library()
@@ -98,3 +99,12 @@ def fancy_join(lst, sep=", ", final_sep=" and "):
 @register.filter
 def username_from_email(email):
     return email.split("@")[0]
+
+
+@register.simple_tag(takes_context=True)
+def dashboard_measure_uri(context, measure):
+    return format_html(
+        "{}#{}",
+        context["dashboard_uri"],
+        measure.id,
+    )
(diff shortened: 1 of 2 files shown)


openprescribing/frontend/templatetags/template_extras.py
Jongmassey, line 1: I'm not 100% au fait with custom template tags, but could the whole `<a>` tag not be generated rather than just the url, to save on a bit more repetition? They all seem to be of the form `<a href="{% dashboard_measure_uri measure %}">{{ measure.name }}</a>`

Jongmassey (approved):
With the proviso that the unit tests pass this looks good. I only had one question which may be borne of a lack of understanding of custom template tags.

openprescribing/frontend/templatetags/template_extras.py
evansd: Yeah it could, but then you're out of the realm of `register.simple_tag` and into something more complex. As for the repetition, in general these templates aren't particularly well-factored anyway and I just want to make the minimal changes needed to fix the bug.

PR #5077, opened 2024-11-27
Fix logic for selecting practices by SICBL

evansd:
The API allows you to retrieve practice-level data by supplying a 3-character code for a SICBL/CGG which then selects all the practices belonging to it. However, previously, if you supplied the code of a closed SICBL with no practices it would end up returning five years' worth of data for every practice in England. Not only is this an incorrect respond but it takes a lot of server resources to generate and has the side effect of segfaulting PhantomJS when it tries to render the chart image.

This PR applies the smallest fix to prevent it doing this.

### DIFF
--- a/openprescribing/api/views_spending.py
+++ b/openprescribing/api/views_spending.py
@@ -367,9 +367,10 @@ def spending_by_org(request, format=None, org_type=None):
 
 def _get_org_type_and_orgs(org_type, org_ids):
     org_type = utils.translate_org_type(org_type)
+    select_all_orgs = not org_ids
 
     # If no org parameters are supplied then we sum over absolutely everything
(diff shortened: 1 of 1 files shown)


evansd:
Sorry, yes. I should have explained better.

The problem comes here:
https://github.com/ebmdatalab/openprescribing/blob/f03576a3b0d0fb84f6e3a128eb255c304a95984e/openprescribing/api/views_spending.py#L376-L379

`org_ids` may start off containing a 3-char CCG code, but if that refers to a closed CCG then it gets turned into an empty list which then resulted in an inadvertent select-all.

inglesp:
Aha, thanks.  Revolting!

PR #5089, opened 2024-12-10
Update raw_prescribing_normalised.sql

richiecroker:
Replace `PARSE_DATETIME` with `PARSE_TIMESTAMP` to get timestamp, in line with other views and tables


evansd:
@inglesp This seems reasonable to me but I don't know if anything else depends on the specific type here?

inglesp:
Am not convinced this will work but we can merge and run the pipeline's end-to-end tests to check.

inglesp (approved):

PR #5102, opened 2025-01-06
Change to SQL-based formulation swaps csv

richiecroker:


evansd (approved):
Approved. Feel free to merge whenever you're happy.

Just to warn you that the very first request after deploying this will be a bit slow because it will have to re-do the calculations and cache the results. It can even be so slow you get a timeout error. But subsequent requests should work fine.

I tend to use this page as my default "load all the data" page:
https://openprescribing.net/national/england/

evansd:
This was implemented in:
 * https://github.com/bennettoxford/openprescribing/pull/5126

PR #4530, opened 2023-12-20
Run tests under Python 3.12

evansd:
This involves building and pushing new Docker images.

We do this from a clean checkout to avoid including any local cruft in the image:
```sh
git clone . ../op-tmp
cd ../op-tmp
```

Build and push a new base image:
```sh
docker build -t ghcr.io/ebmdatalab/openprescribing-py312-base .
docker push ghcr.io/ebmdatalab/openprescribing-py312-base:latest
```

Build and push a new test image:
```
docker build -t ghcr.io/ebmdatalab/openprescribing-py312-test -f Dockerfile-test .
docker push ghcr.io/ebmdatalab/openprescribing-py312-test:latest
```

Make each of the newly pushed images public:
https://github.com/orgs/ebmdatalab/packages/container/openprescribing-py312-base/settings
https://github.com/orgs/ebmdatalab/packages/container/openprescribing-py312-test/settings

PR #5106, opened 2025-01-10
Update the docker docs for ghrc.io

madwort:


README.md
```diff
@@ -128,17 +128,17 @@ database configuration, you'll need to blow away the volume with:
...
     # Built from base image, with extras for testing
-    docker build -t ebmdatalab/openprescribing-py3-test -f Dockerfile-test .
-    docker login  # details in `pass`; only have to do this once on your machin
-    # push the images to hub.docker.io
-    docker push ebmdatalab/openprescribing-py3-base
-    docker push ebmdatalab/openprescribing-py3-test
+    docker build -t ebmdatalab/openprescribing-py312-test -f Dockerfile-test .
+    docker login  # login with a PAT from Github. only have to do this once on your machine
+    # push the images to ghrc.io
+    docker push ebmdatalab/openprescribing-py312-base
```
madwort, line 140: hmm not sure this is correct

evansd (commented):
I think these instructions are pretty out of date. It's up to you if you just want to delete them or put the time in to make them correct. 

If the latter, then one thing to note is that the images only need rebuilding when the npm dependencies change and not when the Python requirements change.

There are some docs on what I did last time I had to rebuild and image here:
https://github.com/ebmdatalab/openprescribing/pull/4530

PR #5108, opened 2025-01-14
Update SICBL names based on ICB name

### LINKED ISSUE #5104: Handle sub-ICB name changes
A user has emailed to advise that sub-ICB D2P2L is now just NHS Black Country and not [NHS Black Country and West Birmingham](https://openprescribing.net/sicbl/D2P2L/) as displayed.

This issue is due to sub-ICBs not having an official user friendly name. For example where an ICB contains multiple sub-ICBs the sub-ICB names will be ICB name plus sub-ICB code:
Image

This isn't very user friendly to understand the data being displayed. We have therefore made the decision to continue to use the legacy CCG names. 

With the implementation of ICBs there were a small number of changes which effects a small number of sub-ICB names:
https://digital.nhs.uk/services/organisation-data-service/upcoming-code-changes#summary-of-boundary-changes

Of these:

- [x] Bassetlaw CCG - **correctly showing under Nottingham and Nottinghamshire Integrated Care Board**

- [x] Part of the Glossop area, previously part of Tameside and Glossop CCG, moved into Derby and Derbyshire CCG. This meant that the boundary of the Joined Up Care Derbyshire system expanded, and the boundary of the Greater Manchester Health and Social Care system reduced.
**Only part of Glossop area has moved - so presumably Tameside and Glossop CCG is still valid even if not all of Glossop. Possibly not perfect but good enough in the absence of an official sub-ICB name**

- [ ] Part of the West Birmingham area, previously part of Black Country and West Birmingham CCG, moved into Birmingham and Solihull CCG. This meant that the boundary of the Birmingham and Solihull system expanded, and the boundary of the previous Black Country and West Birmingham system reduced. The new ICB is be NHS Black Country Integrated Care Board.
**Looking to develop code so that where only 1 sub-ICB in an ICB then the sub-ICB name mirrors the ICB name - see below**

- [x] Part of the Oundle area, previously part of Cambridgeshire and Peterborough CCG, moved into Northamptonshire CCG. This meant that the boundary of the Northamptonshire system expanded, and the boundary of Cambridgeshire and Peterborough system reduced. Note that, as an exception to reflect specific local considerations, a single GP practice (Wansford and Kings Cliffe) was retained within Cambridgeshire and Peterborough ICS.
**No obvious change needed**

evansd:
Where a SICBL is the only active one in its ICB it should inherit the name of that ICB (minus the "INTERGRATED CARE BOARD" suffix).

I've confirmed that this has the same behaviour as Chris's SQL and only renames two SICBLs:

    11N: NHS KERNOW -> NHS CORNWALL AND THE ISLES OF SCILLY
    D2P2L: NHS BLACK COUNTRY AND WEST BIRMINGHAM -> NHS BLACK COUNTRY

These changes will happen automatically during the next data import, but we could push them through manually earlier if we need to.

Closes #5104


richiecroker (commented):
Thanks @evansd - happy with the concept, although I will leave others to review the code!

PR #5112, opened 2025-01-17
Update bdzper1000.json

richiecroker:
add new calculation based on dm+d values


richiecroker:
@a-d-brown can you check this one please.

I was doing a bit of work at the ICB and noticed a couple of (small) items missing, as well as daridorexant appearing in the table.

I've altered in the following way:

- Moved to a chemical substance-based table (in measures/tables)
- Created a view (in measures/views) which calculates ADQ values from the dm+d strength tables, correcting for micrograms where appropriate.
- Removed *other hypnotics and anxiolytics* from the code

I'll do the same for the other measure.

a-d-brown:
Structure looks good! @richiecroker 

- I don't think the injectable formulations of Diazepam (e.g. Diazepam 10mg/2ml emulsion for injection ampoules) have ADQ values, so should we exclude these too?

- Also, I think some brand names like Diazepam 10mg RecTubes aren't excluded in the current code but probably should be?

These items wouldn't be prescribed in primary care very often (if ever), but if we're being fully watertight then might make sense to add them? If that makes sense, would you like me to run through and add all relevant items directly into the code?

a-d-brown (approved):
I've added in a few additional drugs / formulations that don't have an ADQ. Check you're happy with the amendments then good to go @richiecroker

richiecroker:
@a-d-brown updated view and measure code for oral meds via dm+d and removed other codes

a-d-brown (approved):

a-d-brown (approved):

PR #5114, opened 2025-01-20
Update bdzadq.json

richiecroker:


richiecroker:
See other measure for details.  Have added an `items` column in the view, so we can calculate off one table.

a-d-brown (approved):

PR #5134, opened 2025-02-13
Update PPU costs to account for discounts

evansd:
This updates the PPU calculation so that when we aggregate the prices paid for a given presentation we apply the appropriate discount, as determined by the tariff category of that presentation. We also take into account whether there was a price concession for the presentation in that month because, if so, no discount is applicable.

This borrows from logic already implemented in:
 * https://github.com/bennettoxford/openprescribing/pull/4869


richiecroker:
@evansd happy with changes made.

PR #5137, opened 2025-02-17
Update ncso import command to handle new rolled over concessions table

rebkwok:
In previous months, rolled over concessions have been in the monthly concession table, either with an * or in a different colour. This month, they are still in the Feb concession table, but there is an additional section and table for "Rolled over concessions" which shows just the rolled over ones. This has caused the [import to error](https://bennettoxford.slack.com/archives/CGF9TKZLG/p1739791386365429).

I considered just skipping this table, but I'm concerned that future reports might separate out the rolled-over concessions so we'd miss them by parsing only the current month's table. So I've made it handle both tables, and filter out duplicate VMPP matches.


evansd (approved):
Looks like a sensible robust approach to this kind of ongoing nonsense. Thanks so much for picking this up!

PR #5148, opened 2025-03-04
Make tweaks surfaced by running tests locally on a Mac

alarthast:
Running the django test suite on a Mac has surfaced:
- A log file that is generated when the tests are run and has the absolute path on a dev's local machine written to it: Propose to add it to the .gitignore
- Warnings / TypeErrors: these are fixable with minuscule changes; proposed here
- Some tests that fail in the Mac local dev setup: the tests have been tweaked slightly to make them pass


evansd (approved):
Thanks for doing this Alice, and sorry it's all such a mess!

Issue #5109, opened 2025-01-15
Update the README instructions regarding the `environment` file


alarthast:
The README has a line asking the user to create the environment file from the provided sample.
https://github.com/ebmdatalab/openprescribing/blob/9e141d85dc378b26b083b408704fd9d48184bc94/README.md?plain=1#L83

At least at first glance, this contradicts with the first line of the makefile for tests, which overwrites the file
https://github.com/ebmdatalab/openprescribing/blob/9e141d85dc378b26b083b408704fd9d48184bc94/openprescribing/Makefile#L2

We should clarify the instructions in the README.

PR #5150, opened 2025-03-05
Add instructions for local setup on MacOS

alarthast:
Recently, OpenPrescribing was set up on my M2 MacBook. We thought it would be worth documenting the steps taken, as some parts of the README are outdated or linux-specific. 

I have taken the approach of creating DEVELOPERS.md and putting a MacOS section in it, and have tried to make it such that someone can use this doc only to do the local setup and not have to cross-reference with the readme. Hence, a handful of sentences in the README are repeated here. 

#### Notes
- I've not mentioned the `make test` recipe
  - As noted in #5109, `make test` overwrites the existing environment file, which is a bit annoying.
  - On a Mac, I don't seem to have needed the environment variables in the `environment-test` template - Django did its thing with setting up and destroying a test database and the tests ran fine.
- I've also not mentioned the `TEST_SUITE` environment variable.
  - It seems that `TEST_SUITE="nonfunctional"` doesn't actually exclude the functional tests? I might be wrong but with and without it, the same number of tests seem to run.
  - `TEST_SUITE="functional"` looks to be only a shorthand for `(manage.py test) frontend.tests.functional`, given that I point already towards how to run a subset of tests, it doesn't seem important enough to mention, especially since `TEST_SUITE="nonfunctional"` is confusing.

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -1,0 +1,152 @@
+This document describes the local setup process for developers.
+
+It is a work in progress and only has instructions for macOS for now.
+
+## macOS local setup
+
+### Prerequisites
+These instructions assumes:
+ - Usage of [Homebrew](https://brew.sh/) as a package manager
+ - [Postgres.app](https://postgresapp.com/) has been installed
+
+### Install gdal
+This can be done [via Homebrew](https://formulae.brew.sh/formula/gdal).
+
+### Set up a virtualenv
+Set up a virtual environment with the Python version specified in `.python-version`.
+The architecture of your Python executable must match that of gdal's (which should
+be x86_64 if installed via Homebrew).
+If your Python is of another architecture (e.g. aarch64), you can get the x86_64
+Python [via Homebrew](https://formulae.brew.sh/formula/python@3.12).
+
+### Install Python dependencies
+Install the Python dependencies into your virtual environment:
+```sh
+    pip install -r requirements.dev.txt
+    pip install -r requirements.txt
+```
+
+### Create environment file
+Copy `environment-sample` to `environment`, leaving the fields unchanged for now.
+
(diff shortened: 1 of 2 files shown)


DEVELOPERS.md
```diff
@@ -0,0 +1,150 @@
...
+`openprescribing/pipeline/data/matrixstore_build/matrixstore_live.sqlite`.
+
+At this point, you should be able to run the local server and verify that the
+"Analyse" and "Find a Practice" pages work.
+
+### Set up Google Cloud and BigQuery permissions for tests
+To run the tests, you would need to create datasets in the `ebmdatalabtest`
+project on BigQuery. 
+
+Contact the Bennett admins to add you to the project.
```
alarthast, line 101: @inglesp: Does this sound about right? Do I need to mention what role or would you basically be the bennett admin  doing it?

DEVELOPERS.md
```diff
@@ -0,0 +1,150 @@
...
+To generate monthly alert emails (and run the tests for those) you'll
+need a `phantomjs` binary located at `/usr/local/bin/phantomjs`. Get
+it from [here](http://phantomjs.org/download.html).
+
+### Run the tests locally
+In CI, the frontend assets are rebuilt prior to testing. This is
+usually unnecessary when testing locally, and has the downside of
+modifying the assets. You can set the `SKIP_NPM_BUILD` environment
+variable (to any value) to skip the rebuild, and might want to
+consider adding it to your ~/.bashrc file (or equivalent).
```
alarthast, line 125: @inglesp I think I've got the rationale right, would you be able to double check?

inglesp (approved):
Only suggestion is to link to this from the README.  Otherwise, it's great!

PR #5174, opened 2025-04-23
Support ingesting a CSV file of manually added price concessions

### LINKED ISSUE #5170: Support ingesting a CSV file of manually added price concessions
We scrape price concession data from the website of Community Pharmacy England. Mostly this works OK, but occasionally there are concessions announced "out-of-band" i.e. elsewhere on the CPE website which aren't included in the tables we scrape. We'd like a mechanism for adding these concessions to our data.

For now, the simplest way to do this is to have a CSV file of additional concessions committed to the repository. When we run the scraper we can also read the contents of this CSV file and add the results to the list of scraped concessions. 

I imagine adding a new line after this one:
https://github.com/bennettoxford/openprescribing/blob/91153e1364a2ab0ab5e74628255bca4b12dcfd8b/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py#L62

Which would be something like:
```py
items.extend(read_concessions_csv(MANUALLY_ADDED_CONCESSIONS_PATH))
```

The CSV file itself would look something like:

Date | Name | Pack Size | Price Pence | Notes | URL (if applicable)
-- | -- | -- | -- | -- | --
2025-03-01 | Trimethoprim 200mg tablets | 6 | 179 |   |  
2025-03-01 | Trimethoprim 200mg tablets | 14 | 419 |   |  

This should be parsed as a list of dictionaries with the following structure:
```py
[
    {
        "date": datetime.date(2025, 3, 1),
        "drug": "Trimethoprim 200mg tablets",
        "pack_size": "6",
        "price_pence": 179,
    },
    ...
]
```

Note that `pack_size` is a string but `price_pence` is an int. The "Notes" and "URL" columns are just for our own reference and can be ignored.

Dates should always be the first of the month and you should add an assertion to this effect.

I'm not sure exactly where's best for this CSV file to live. But I'd probably default to putting it next to the management command.

The tests (which are [here](https://github.com/bennettoxford/openprescribing/blob/91153e1364a2ab0ab5e74628255bca4b12dcfd8b/openprescribing/pipeline/tests/test_fetch_and_import_ncso_concessions.py)) should ensure that the current version of the CSV file can be parsed without error. But we shouldn't assert anything specific about its contents as these will change over time.

alarthast:
Closes #5170. 

See the issue and the [Slack thread](https://bennettoxford.slack.com/archives/C051A4P27GE/p1744025726762869) for context.

I've made the commits with the intention that they would be reviewed one by one. 
See the commit messages for what has been done.

### DIFF
--- a/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
+++ b/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
@@ -251,6 +251,7 @@ def convert_concessions_csv_row(row):
         "drug": row["Name"],
         "pack_size": row["Pack Size"],
         "price_pence": int(row["Price Pence"]),
+        "manually_added": True,
     }
 
 
@@ -369,15 +370,19 @@ def insert_or_update(items):
 
 
 def format_message(inserted):
-    created = [i for i in inserted if i["created"]]
+    num_manually_added = sum(1 for i in inserted if i.get("manually_added"))
+    num_created = sum(1 for i in inserted if i["created"])
     unmatched = [i for i in inserted if i["vmpp_id"] is None]
 
-    msg = f"Fetched {len(inserted)} concessions. "
+    msg = (
+        f"Fetched {len(inserted)} concessions, "
+        f"including {num_manually_added} manually added concessions. "
+    )
 
-    if not created:
+    if num_created == 0:
         msg += "Found no new concessions to import."
     else:
-        msg += f"Imported {len(created)} new concessions."
+        msg += f"Imported {num_created} new concessions."
 
     # Warn about cases where we couldn't match the drug name and pack size to a VMPP
     if unmatched:
--- a/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
+++ b/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
@@ -352,7 +352,7 @@ def format_message(inserted):
 
     # Warn about cases where we couldn't match the drug name and pack size to a VMPP
     if unmatched:
-        msg += "\n\n" "The following concessions will need to be manually matched:\n"
+        msg += "\n\nThe following concessions will need to be manually matched:\n"
         for item in unmatched:
             msg += f"Name: {item['drug']} {item['pack_size']}\n"
 
--- a/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
+++ b/openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
@@ -1,5 +1,6 @@
 # coding=utf8
 
+import csv
 import datetime
 import logging
 import re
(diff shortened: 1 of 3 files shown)


openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
```diff
@@ -352,7 +352,7 @@ def format_message(inserted):
 
     # Warn about cases where we couldn't match the drug name and pack size to a VMPP
     if unmatched:
-        msg += "\n\n" "The following concessions will need to be manually matched:\n"
+        msg += "\n\nThe following concessions will need to be manually matched:\n"
```
evansd, line 390: This change is fine, but worth noting that OpenPrescribing still uses Black so that's what the lint checks will run with.

openprescribing/frontend/migrations/0084_alter_pcn_options.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 4.2.20 on 2025-04-22 08:16
```
evansd, line 1: Thanks for spotting this. This is one of those migrations which don't actually involve any change to the database so I'm guessing that's why it never got created. I don't feel particularly strongly either way, but I'd lean towards _not_ having this in the PR just because it's completely unrelated to the other changes. Having it as a tiny separate PR would make for a slightly more comprehensible history I think.

openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
```diff
@@ -225,6 +226,25 @@ def match_concession_vmpp_ids(items, vmpp_id_to_name):
     return matched
 
 
+def read_concessions_csv(path):
+    def convert(row):
```
evansd, line 242: I'd be minded not to make this an inner function but just a regular top level function instead (`convert_csv_row` or something). One reason is that having is an inner function makes it awkward to test without having to construct an entire CSV file to do it.

evansd (approved):
This looks really good, thanks Alice. I've made a few comments but I think the only think that really needs to change before deployment is switching to `DictReader` and using header names rather than column position when parsing.

Great work though, thanks for taking this on.

openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
alarthast: Noted, I've kept this anyway to keep my IDE from highlighting it.

openprescribing/frontend/migrations/0084_alter_pcn_options.py
alarthast: I see - I've dropped that commit from this PR and moved it to https://github.com/bennettoxford/openprescribing/pull/5176 .

openprescribing/pipeline/management/commands/supplementary_files/fetch_and_import_ncso_concessions/manually_added_concessions.csv
```diff
@@ -0,0 +1,3 @@
+ID,Date,Name,Pack Size,Price Pence,Notes,URL (if applicable)
+0,2025-03-01,Trimethoprim 200mg tablets,6,179,,
```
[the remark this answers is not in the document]
alarthast: Ahh, that should have been obvious, apologies. I've added it back in!

openprescribing/pipeline/management/commands/fetch_and_import_ncso_concessions.py
```diff
@@ -84,7 +85,7 @@ def handle(self, *args, **kwargs):
         Client("dmd").upload_model(NCSOConcession)
 
         # Report results
-        msg = format_message(inserted)
+        msg = format_message(inserted, len(manually_added_items))
```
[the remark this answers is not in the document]
alarthast: Thank you!! I have made the change except I used `sum` instead of `len(list)`. 
I hope it's still acceptable - if not, I can change things back!!

alarthast:
Many thanks @evansd for reviewing! I've converted this from a draft PR to a real one. 
Would you like to take a second look or should I go ahead and merge?

evansd (approved):
All looks great, Alice! Merge whenever you like. Deployment should happen automatically and then the new concessions will be imported the next time the scraper runs.

PR #5188, opened 2025-05-22
add data submission to button text

richiecroker:


evansd:
The changes to the tariff text make it take 4 lines (on my browser) so they still don't line up!
image

evansd (approved):
Bingo!
image

PR #5224, opened 2025-08-18
Turn DOAC into having a cost component

richiecroker:


richiecroker:
Now > 6months since price reduction, so OK for us to switch to having a cost calculation.

PR #5235, opened 2025-09-05
Update the header logo and nav bar styles

tomodwyer:
And remove the text shadow from headings.

## Before

CleanShot 2025-09-05 at 08 39 06@2x

## After

CleanShot 2025-09-04 at 17 33 39@2x


openprescribing/templates/base.html
```diff
@@ -90,12 +90,68 @@
             <span class="icon-bar"></span>
           </button>
           <a class="navbar-brand" href="{% url 'home' %}">
-          <img class="logo" src="{% static 'img/logo.svg' %}" alt="OpenPrescribing logo" onerror="this.src='{% static "img/logo.png" %}';this.onerror=null;" />
-          OpenPrescribing
+            <span class="sr-only">OpenPrescribing</span>
+            <svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" fill-rule="evenodd" stroke-linejoin="round" stroke-miterlimit="2" clip-rule="evenodd" viewBox="0 0 383 110">
```
evansd, line 94: Do we have to inline this so we can style it? Is there any alternative?

evansd (approved):

openprescribing/templates/base.html
tomodwyer: I would prefer it to be code rather than an `img` tag, because that is easier to style. Would it be best if I switched it to an include tag and made a HTML partial?

PR #5265, opened 2025-10-22
Make the 500 page slightly nicer

evansd:
Note that this has to be entirely self-contained and so can't rely on external CSS or assets.

It is served directly by nginx and so is visible even if the Python web process isn't running.


evansd:
It looks like this:

image

inglesp (approved):

PR #5285, opened 2025-11-25
Import organisations from ODS

inglesp:
The NHS's [Organisation Data Service](https://www.odsdatasearchandexport.nhs.uk/) is now the only place where we can get information about practices and the organisations they belong to.  This PR updates our import pipeline to pull in data from the ODS.

### DIFF
--- a/openprescribing/pipeline/runner.py
+++ b/openprescribing/pipeline/runner.py
@@ -84,6 +84,8 @@ class Task(object):
     def filename_pattern(self):
         """Return pattern that matches the part of the task's command that
         should be substituted for the task's input filename."""
+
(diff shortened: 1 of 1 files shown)


openprescribing/pipeline/tests/test_import_org_details.py
```diff
@@ -0,0 +1,209 @@
+import datetime
+from django.forms.models import model_to_dict
+from django.test import TestCase
+from frontend.models import PCN, PCT, Practice, RegionalTeam, STP
+from pipeline.management.commands.import_org_codes import import_all, import_sicbls
```
evansd:
```suggestion
from pipeline.management.commands.import_org_details import import_all, import_sicbls
```

PR #5556, opened 2026-07-22
Remove SeleniumTestCase.host

iaindillingham:
SeleniumTestCase.host set the server's bind address to 0.0.0.0, or "listen on every network interface". However, SeleniumTestCase.host is used by SeleniumTestCase.live_server_url to construct URLs. Whilst most OSs reject attempts to connect to 0.0.0.0, some treat 0.0.0.0 as equivalent to 127.0.0.1 (localhost). Consequently, tests that called SeleniumTestCase.live_server_url would pass or fail, depending on the OS.

Rather than reasoning about why these tests pass or fail, it's easier to remove SeleniumTestCase.host, and default to LiveServerTestCase.host, which is is set to localhost.

### DIFF
--- a/openprescribing/frontend/tests/functional/selenium_base.py
+++ b/openprescribing/frontend/tests/functional/selenium_base.py
@@ -33,7 +33,6 @@ def use_browserstack():
     "nonfunctional tests specified in TEST_SUITE environment variable",
 )
 class SeleniumTestCase(StaticLiveServerTestCase):
-    host = "0.0.0.0"
     display = None
 
     @classmethod


iaindillingham:
Alas, this isn't going to work. To understand why, it's important to know that:

* on some systems, requests to connect to `0.0.0.0` are treated as requests to connect to the local host, if a listener is bound on the system.
* BrowserStack's local agent (`BrowserStackLocal`) seems to do the same.
* Django's live server (a listener) must bind to `0.0.0.0` in a container, such that it can be reached from the host.

Functional tests can be run either locally (the host) or in a container, and either with or without BrowserStack. Let's consider the four cases:

* Locally with BrowserStack. Either the system or BrowserStack's local agent treats a request to connect to `0.0.0.0` as a request to connect to the local host. A connection is established.
* Container with BrowserStack. As above. A connection is established.
* Locally without BrowserStack. The system may or may not establish a connection. macOS doesn't establish a connection.
* Container without BrowserStack. As above, remembering that "local" means "local to the container", rather than "local to the host". The system may or may not establish a connection. The container runs Debian GNU/Linux 12 (bookworm), which establishes a connection.

Changing the bind address of Django's live server to `localhost` means that it can't be reached from the host when run in a container, so the two container cases break. Doing so means that it can be reached from the host when run locally, fixing the "Locally without BrowserStack" case. However, this is the least important case, as the tests are unlikely to be targeting the local browser. There is no change to the "Locally with BrowserStack" case.

If you've followed me this far, then congratulations. If you haven't, then this isn't going to work: it fixes one less important thing but breaks two more important things.

PR #5567, opened 2026-07-29
Update geckodriver, and add recipes for building and pushing images

iaindillingham:
The older version of geckodriver isn't compatible with the version of Firefox on the base and test images, which causes the functional tests to fail when run on a container without BrowserStack.


iaindillingham:
I've built and pushed [base](https://github.com/orgs/bennettoxford/packages/container/package/openprescribing-py312-base) and [test](https://github.com/orgs/bennettoxford/packages/container/package/openprescribing-py312-test) images.

PR #5574, opened 2026-08-04
Update lpaliskiren.json

richiecroker:


chrisjwood16:
There was quite a sudden drop off in 2024 across primary care. Possibly the UK product discontinued - only available products now are parallel imports. If the trend continues primary care use might stop entirely in next few years.

chrisjwood16 (approved):

PR #5606, opened 2026-08-19
Remove `environment-test`

iaindillingham:
I expect the checks to fail, because of missing environment variables.


openprescribing/Makefile
```diff
@@ -1,5 +1,4 @@
 test :
-	cp ../environment-test ../environment
```
iaindillingham: This is harmless when run in CI, annoying when run in development ("Why can I no longer connect to the database?"), and a significant problem if run in production. Removing it motivated this PR.

PR #5610, opened 2026-08-20
Don't clobber `environment`

iaindillingham:
Previously, calling `make test` would replace `environment` with `environment-test`. This behaviour was harmless when run in CI; annoying when run locally ("Why can't I connect to the database any more?"); and a significant problem if run in production. Who would do such a thing? Well, with flaky tests run on containers that aren't themselves run in production, and with Django's ability to switch between default and test databases, it's not _completely_ unreasonable that someone would. Ouch.

Still can't see someone calling `make test`? Well, `test.command` (i.e. the command that runs when the test service container starts) calls it, so calling `docker compose run test` was equally troublesome.

After realising how enmeshed `environment` is, thanks to [django-dotenv](https://pypi.org/project/django-dotenv/) (latest release, December 11th 2017!), I decided to dispense with `environment` for development and testing. There's some redundancy, as I didn't want to embark on adding `just` to our GitHub Actions jobs, but it's contained within 1b1190fc.

### DIFF
--- a/openprescribing/Makefile
+++ b/openprescribing/Makefile
@@ -1,5 +1,4 @@
 test :
-	cp ../environment-test ../environment
 	coverage run manage.py test -v 2
 	coverage report
 	cd media/js; npm run test
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,6 +3,9 @@ services:
   postgis:
     image: postgis/postgis:16-3.4
     environment:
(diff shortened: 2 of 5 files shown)


openprescribing/Makefile
```diff
@@ -1,5 +1,4 @@
 test :
-	cp ../environment-test ../environment
```
iaindillingham, line 2: You didn't want to keep that `environment` file, did you?

PR #5635, opened 2026-09-08
Flag BNF codes with no prescribing as `is_current=False`

evansd:
Previously, we would automatically flag codes as `is_current=True` when prescribing appeared for them but we wouldn't do the reverse.

I've looked at the PR which introduced the original behaviour and the associated issues, but it's not clear why it worked this way:
 * https://github.com/bennettoxford/openprescribing/pull/380

This change was prompted by the discussion below, in which @richiecroker agrees that the new behaviour is what we want:
https://bennettoxford.slack.com/archives/C051A4P27GE/p1787666287464959

This command will run automatically as part of the data import process, but if we want the change to take effect without waiting for that then we can run it manually after deployment:
```
./manage.py refresh_bnf_class_currency
```


iaindillingham (approved):
Thank you for a clear description of the problem (with links to the archaeological record) and a logical sequence of commits 🙂

